PR #156, opened 2015-01-27
:arrow_down: nan@^1.5.1

jonathanong:
1.6 breaks iojs. this lowers the minimum nan version so that developers can still do `nan@~1.5.1`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   "dependencies": {
     "bluebird": "^2.8.2",
     "color": "^0.7.3",
-    "nan": "^1.6.1",
+    "nan": "^1.5.1",
     "semver": "^4.2.0"
   },
   "devDependencies": {


lovell:
Hi Jonathan, it looks like https://github.com/rvagg/nan/pull/265 should fix this problem. Perhaps we can help get it merged?

I believe Travis CI adds support for iojs next month so we can catch this sort of problem earlier.

The Snap CI failure is unrelated - it seems libexif is no longer included by default in their CI environment - I'll sort that.

jonathanong:
up to you if you wanna wait. was there a new feature in 1.6 that you need? otherwise, there really isn't a point in bumping the version.

lovell:
I'd started to investigate some of the new Notification methods, but we're not using them yet so happy to downgrade for now. Thanks for the patch!

lovell:
This patch is included in v0.9.1 - happy iojs-ing!

PR #161, opened 2015-02-06
Change nan dependency back to ^1.6.2

ide:
The issue with nan on io.js was fixed in https://github.com/rvagg/nan/pull/273. `npm install` on io.js 1.1.0 works now.


lovell:
Fantastic, thanks James.

Issue #162, opened 2015-02-10
GIF resize throws Error


joelabair:
sharp@0.9.1 
vips-7.40.11-1.fc20.x86_64
vips-devel-7.40.11-1.fc20.x86_64
node v0.10.33

Error: Input buffer contains unsupported image format

calling code
```javascript

var transformer = sharp().resize(maxWidth, maxHeight).max().withoutEnlargement().toFile(downSampleCacheFile, function(err) {
	if (err) {
		debug("Sharp Error", err);
		return next(err);
	}
	fs.stat(downSampleCacheFile, function(err, stats) {
		if (err) {
			debug("Stat Error", err);
			return next(err);
		}
		res.set('Content-Length', stats.size);
		fs.createReadStream(downSampleCacheFile).pipe(res);
	});
});

readStream.pipe(transformer);
```

lovell:
Hi Joel, whilst GIF files are supported, GIF data is unsupported for Stream-based input.

https://github.com/lovell/sharp#sharpinput

> JPEG, PNG, WebP or TIFF format image data can be streamed into the object when input is not provided.

This is because libvips decompresses GIF data using *magick, which doesn't provide any stream-like support.

joelabair:
Thanks!  Sorry, I overlooked that.  I'll buffer the GIF stream first.

lovell:
No probs. For GIF you'll need to "buffer" to the filesystem (tmpfs will suffice) then provide the filename to sharp's constructor.

joelabair:
@lovell how would I output the gif, after transform?

lovell:
No, sorry.

joelabair:
NP.  I'll just switch to gm on unsupported formats.  Thanks for the quick followup!

lovell:
Thanks to @mcuelenaere, sharp v0.9.3 adds support for GIF *input* via Stream and Buffer. You'll need to compile libvips from the latest source to take advantage.

joelabair:
ooooh!  Sweet!  Thanks for the heads up @lovell

lovell:
Closing this inactive question - feel free to re-open if you have a related question.

PR #165, opened 2015-02-15
Add support for loading images through ImageMagick as a buffer

mcuelenaere:
This requires https://github.com/jcupitt/libvips/pull/241

### DIFF
--- a/index.js
+++ b/index.js
@@ -83,11 +83,13 @@ var Sharp = function(input) {
       (input[0] === 0x52 && input[1] === 0x49) ||
       // TIFF
       (input[0] === 0x4D && input[1] === 0x4D && input[2] === 0x00 && (input[3] === 0x2A || input[3] === 0x2B)) ||
-      (input[0] === 0x49 && input[1] === 0x49 && (input[2] === 0x2A || input[2] === 0x2B) && input[3] === 0x00)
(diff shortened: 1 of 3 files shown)


mcuelenaere:
This fixes https://github.com/lovell/sharp/issues/162.

mcuelenaere:
Not sure how to fix the errors found by the CIs: moving the header checks into a separate function is a possible solution here I suppose?

Incidentally, https://github.com/lovell/sharp/pull/166 should also fix this I presume.

lovell:
Fantastic, thanks for both this and the upstream libvips changes to allow it. I'll let John merge in your PR first.

The test failure is due to a high water mark I've added for cyclomatic complexity. As you say, moving this logic to C++ land will reduce the complexity here.

mcuelenaere:
I changed the minimum version requirements to >= 8.0, as that's what John says in https://github.com/jcupitt/libvips/pull/241.

lovell:
I updated a few `#if` preprocessor statements to accommodate the forthcoming libvips major version increment - see https://github.com/lovell/sharp/commit/f89e9d726d9c6ee99e8e7cf0220b8da6b90cd83d

Thanks again for this - I'll update the docs and provide a credit for your good work.

PR #166, opened 2015-02-15
Let libvips check whether we received a valid image or not

mcuelenaere:
This removes the custom image fingerprinting code and uses the libvips
is_a_buffer() infrastructure instead.

This requires https://github.com/jcupitt/libvips/pull/240.

### DIFF
--- a/src/common.cc
+++ b/src/common.cc
@@ -29,47 +29,40 @@ namespace sharp {
     return EndsWith(str, ".tif") || EndsWith(str, ".tiff") || EndsWith(str, ".TIF") || EndsWith(str, ".TIFF");
   }
 
-  // Buffer content checkers
-  unsigned char const MARKER_JPEG[] = {0xff, 0xd8};
-  unsigned char const MARKER_PNG[] = {0x89, 0x50};
-  unsigned char const MARKER_WEBP[] = {0x52, 0x49};
-
-  static bool buffer_is_tiff(char *buffer, size_t len) {
-    return (
-      len >= 4 && (
-        (buffer[0] == 'M' && buffer[1] == 'M' && buffer[2] == '\0' && (buffer[3] == '*' || buffer[3] == '+')) ||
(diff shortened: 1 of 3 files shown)


lovell:
Thanks again; loving these changes.

This PR slightly changes the API in that the constructor no longer fails fast with an unsupported/invalid Buffer. Perhaps we should verify the Buffer synchronously?

mcuelenaere:
Yes, that's the only controversial change. The only downside to doing it synchronously is that it might cause a possible performance hit. But I suppose that doing a couple of memory comparisons won't be that slow.

lovell:
The more I think about about, the more I'd rather "break" the API and keep your change as it is.

I'd describe it as very useful rather than "controversial" :)

I'll wait for John to merge in the related libvips change before merging this PR. Thanks again!

src/common.cc
```diff
@@ -29,36 +29,40 @@ namespace sharp {
...
-  }
-
   /*
     Determine image format of a buffer.
   */
   ImageType DetermineImageType(void *buffer, size_t const length) {
     ImageType imageType = ImageType::UNKNOWN;
-    if (length >= 4) {
-      if (memcmp(MARKER_JPEG, buffer, 2) == 0) {
+#if (VIPS_MAJOR_VERSION >= 7 && VIPS_MINOR_VERSION >= 42 && VIPS_MICRO_VERSION > 3)
```
lovell: I think this condition now needs to use `VIPS_MAJOR_VERSION >= 8`.

lovell:
Hi @mcuelenaere, John's merged the dependent PR. Are you able to rebase your `mcuelenaere:no-custom-image-header-checking` branch against upstream/master and squash to a single commit?

mcuelenaere:
Changed, squashed and rebased.

lovell:
Marvellous, thank you Maurus.

PR #146, opened 2015-01-07
Added DeepZoom support

mvictoras:
Added DZI as supported output format. Openslide is a prerequisite for this to work.

Issue #19, opened 2014-04-10
Support for Windows


DullReferenceException:
I'd like to be able to use sharp on the Windows platform. Because `libvips` is available for Windows, I assume that lack of support is mostly in figuring out the build script.

Has this been investigated already? If so, what approaches have already been tried and roadblocks encountered? Knowing what's been tried so far, if anything, will help me or other potential developers come to a working solution quicker.

lovell:
@michaeldrotar I think you may be experiencing a different problem to @blowsie as you're not seeing the `'true' is not recognized as an internal or external command...` error. It might be that the `which` executable is not found/available on your system.

The `which brew >/dev/null 2>&1 && eval $(brew --env) && echo $PKG_CONFIG_LIBDIR` part of PKG_CONFIG_PATH is applicable for Macs only. We need to work out a portable way of making Windows ignore any error it generates, like the `|| true` command does for Linux.

I now have in my possession a new Windows 8.1 laptop so hope to make a bit more progress with this. Please do keep adding comments to this issue about your own attempts of getting it working as these all help narrow down the problems.

(On a multi-platform note, I managed to get libvips and sharp working on an ARMv6 Raspberry Pi over Christmas. WebP support was the trickiest bit to get working.)

Karnith:
+1
Would be great to get this working!

connor4312:
:+1: would love to move from gm to Sharp for actual good streaming support :D

mrmuh:
+1 :)

blowsie:
>@blowsie Please can you try changing || true to || exit 0 on line 12 of binding.gyp.

How do I go about doing this when I'm running `npm install`? The files don't exist.

lovell:
@blowsie Thanks for the reply. You'll need to fork your own copy of the sharp repo, clone it locally then make the change. You should then be able to run `npm install` inside the cloned repo to prove it works before committing the change and pushing back to Github.

blowsie:
@lovell thanks!

Unfortunately I get a similar error

```
Microsoft Windows [Version 6.3.9600]
(c) 2013 Microsoft Corporation. All rights reserved.

D:\Projects\github\sharp>npm install
\
> sharp@0.9.3 install D:\Projects\github\sharp
> node-gyp rebuild


D:\Projects\github\sharp>if not defined npm_config_node_gyp (node "C:\Program Fi
[...]

PR #171, opened 2015-02-25
Intake

mvictoras:
New pull request based on the changes that we discussed.
There is no intake branch anymore?


lovell:
Hi Victor, the `intake` branch was the work-in-progress for the 0.9 release. As you've guessed, the `master` branch is probably the best home for this at the moment as things seem pretty stable with the current release.

Are you able to rebase against lovell:master first then squash your changes into a single commit?

binding.gyp
```diff
@@ -9,14 +9,25 @@
...
+      'PKG_CONFIG_PATH': '<!(which brew >/dev/null 2>&1 && eval $(brew --env) && echo $PKG_CONFIG_LIBDIR || true):$PKG_CONFIG_PATH:/usr/local/lib/pkgconfig:/usr/lib/pkgconfig',
+      'HAS_OPENSLIDE': '<!(./has_openslide.sh)'
     },
-    'libraries': [
-      '<!(PKG_CONFIG_PATH="<(PKG_CONFIG_PATH)" pkg-config --libs vips)'
-    ],
-    'include_dirs': [
-      '<!(PKG_CONFIG_PATH="<(PKG_CONFIG_PATH)" pkg-config --cflags vips glib-2.0)',
-      '<!(node -e "require(\'nan\')")'
+    'conditions': [
```
lovell: Is this conditionality required? `pkg-config --cflags vips` already includes the openslide dependency (when present).

src/common.cc
```diff
@@ -28,7 +28,11 @@ namespace sharp {
   bool IsTiff(std::string const &str) {
     return EndsWith(str, ".tif") || EndsWith(str, ".tiff") || EndsWith(str, ".TIF") || EndsWith(str, ".TIFF");
   }
-
+#ifdef HAS_OPENSLIDE
```
lovell: We probably only need to use the `#ifdef HAS_OPENSLIDE` wrapper around code that won't otherwise compile, e.g. when calling `vips_openslideload`.

lovell:
I'd like to take John's [advice](https://github.com/lovell/sharp/pull/146#issuecomment-74052077) and use `vips_type_find` at runtime to build a map of available features, which could include Openslide support. This should provide a more generic equivalent of the `hasOpenslide` this PR adds.

The compile-time `#ifdef HAS_OPENSLIDE` around libvips methods that only exist when it is compiled with Openslide support would still be required with the current minimum libvips version of 7.40.0.

Would you like me to add the generic `hasFeature` method before you have to handle any rebase conflicts? I'll try to implement it in such as way as to avoid conflicts with your incoming changes.

mvictoras:
No I did not rebase and squash my changes into a single commit, I just created a pull request. I will do it in my next pull request.

Yes please, add a hasFeature method, so I can handle the rebase conflicts.

binding.gyp
mvictoras: I need to test that, I think you are right.

src/common.cc
mvictoras: Sounds good.

mvictoras:
I am also working on an install script for openslide, (like [this](https://github.com/lovell/sharp/blob/master/preinstall.sh) one), do you want this to be incorporated into the preinstall script? In that case, every time you run preinstall you install openslide and then vips. Another option would be being optional as an argument. For example, you run
./preinstall.sh --with-openslide.

Or it can be a completely new script, ex. preinstall-openslide.sh

lovell:
I've started on the "hasFeature" API this morning, which for file-formats will hopefully re-use the existing `sharp.format.*` enum-like Object employed by `toFormat`.

I will also add openslide for loading and libgsf for saving deepzoom as dependencies to the existing `preinstall.sh` script for flavours of Linux where a suitable, pre-compiled version is available via package manager (e.g. >=3.4.0 for openslide).

Once these are done your very useful additions should fit more neatly - sorry this has been a bit messy.

mvictoras:
@lovell, I actually already added openslide support in the preinstall.sh script. You can see it here:
https://github.com/mvictoras/sharp/blob/intake/preinstall.sh

They way it works is as follows:
To install only vips (same behavior as before)
./preinstall.sh 

To install vips with openslide support:
./preinstall.sh --with-openslide
Checks for openslide and vips. If openslide not installed, it installs it. If vips is already installed but lacks openslide support, then vips gets recompiled with openslide support.

I have tested the script on all platforms EXCEPT Macports and Amazon cloud, since I don't have any of those around. I am currently working on adding openSuse, since this is the distro we use in our servers.

One last thing: regarding your comment for the pre-compiled version of openslide, remember that the pre-compiled version of vips on all platforms has not been compiled with openslide support, so you will need to compile it anyways.

Hope my comment got there in time, so you don't have to re-write the script!

mvictoras:
Of course, we can make openslide mandatory and get rid of the "--with-openslide" option. That is up to you to decide.

lovell:
Pre-compiled libgsf-1-dev(el) seems to be universally available and I was going to add (lib)openslide-dev(el) only where the package was readily available for a given version of Linux.

I like your idea of an optional `--with-openslide` to compile openslide from source for the others, so please keep that in the PR, thank you.

Getting openSUSE support in there too would be fantastic, thank you!

lovell:
Commit https://github.com/lovell/sharp/commit/c7ccf6801d5b74215e15b32975d8a1a61e03b400 should hopefully make things a little easier, albeit with a few conflicts when you rebase.

The new [format](https://github.com/lovell/sharp#format) attribute allows things like:
```javascript
if (sharp.format.magick.input.buffer) {
  // I can load GIF images from a Buffer
  sharp(gifBuffer)...
}
```
... which you'll see the tests are using too - much simpler than the semver logic before. You might even be able to remove most if not all of the `#ifdef` logic from your changes.

You'll [see](https://github.com/lovell/sharp/blob/master/src/utilities.cc#L99) the `openslide` and `dz` formats are checked, waiting for your new API to expose them.

Commit https://github.com/lovell/sharp/commit/1565522ecca1b476e28f0498c9392673e92e0c84 makes libgsf a dependency, no big deal as it's pre-compiled for all the Linux varieties and [homebrew](https://github.com/Homebrew/homebrew-science/blob/master/vips.rb#L25) already includes it.

Finally, I need to configure Snap CI to use a modern compiler as it's still using g++ 4.4, which can't handle C++11 properly, hence the failures we're seeing there.

mvictoras:
Great, thank you!
Will be working on this over the weekend, want to have this done by Monday.

mvictoras:
Also have you had the chance to test on Windows too?

lovell:
(Lack of) Windows support will be covered by #19 - it's proving the trickiest OS to get everything working on.

mvictoras:
Great, thanks!

lovell:
Superseded by #146.

Issue #100, opened 2014-10-06
Debian Wheezy support


a2nt:
libvips-dev is installed, but something goes wrong:

```
npm install sharp
|
> sharp@0.6.2 install /srv/node_modules/sharp
> node-gyp rebuild

Package vips was not found in the pkg-config search path.
Perhaps you should add the directory containing `vips.pc'
to the PKG_CONFIG_PATH environment variable
No package 'vips' found
gyp: Call to 'PKG_CONFIG_PATH=":$PKG_CONFIG_PATH:/usr/local/lib/pkgconfig:/usr/lib/pkgconfig" pkg-config --libs vips' returned exit status 1. while trying to load binding.gyp
gyp ERR! configure error
gyp ERR! stack Error: `gyp` failed with exit code: 1
gyp ERR! stack     at ChildProcess.onCpExit (/usr/lib/node_modules/npm/node_modules/node-gyp/lib/configure.js:343:16)
gyp ERR! stack     at ChildProcess.emit (events.js:98:17)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (child_process.js:810:12)
gyp ERR! System Linux 3.2.0-4-amd64
gyp ERR! command "node" "/usr/lib/node_modules/npm/node_modules/node-gyp/bin/node-gyp.js" "rebuild"
gyp ERR! cwd /srv/node_modules/sharp
gyp ERR! node -v v0.10.32
gyp ERR! node-gyp -v v1.0.2
gyp ERR! not ok

npm ERR! Linux 3.2.0-4-amd64
npm ERR! argv "node" "/usr/bin/npm" "install" "sharp"
npm ERR! node v0.10.32
npm ERR! npm  v2.0.0
npm ERR! code ELIFECYCLE
npm ERR! sharp@0.6.2 install: `node-gyp rebuild`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the sharp@0.6.2 install script.
npm ERR! This is most likely a problem with the sharp package,
npm ERR! not with npm itself.
npm ERR! Tell the author that this fails on your system:
npm ERR!     node-gyp rebuild
npm ERR! You can get their info via:
npm ERR!     npm owner ls sharp
npm ERR! There is likely additional logging output above.
```

nordisk:
I have the same problem here: https://github.com/alex88/heroku-buildpack-vips/issues/20

PR #173, opened 2015-02-26
Install libgsf-1-dev on Debian 8

LinusU:
This was needed on my docker container running Debian Jessie. This is the error without it:

```text
Package libgsf-1 was not found in the pkg-config search path.
Perhaps you should add the directory containing `libgsf-1.pc'
to the PKG_CONFIG_PATH environment variable
Package 'libgsf-1', required by 'vips', not found
gyp: Call to 'PKG_CONFIG_PATH=":$PKG_CONFIG_PATH:/usr/local/lib/pkgconfig:/usr/lib/pkgconfig" pkg-config --libs vips' returned exit status 1. while trying to load binding.gyp
```

### DIFF
--- a/preinstall.sh
+++ b/preinstall.sh
@@ -90,7 +90,7 @@ case $(uname -s) in
         jessie|vivid)
           # Debian 8, Ubuntu 15
           echo "Installing libvips via apt-get"
-          apt-get install -y libvips-dev
+          apt-get install -y libvips-dev libgsf-1-dev
           ;;
         trusty|utopic|qiana|rebecca)
           # Ubuntu 14, Mint 17


lovell:
Thanks @LinusU, this problem has come up before, e.g. https://github.com/lovell/sharp/issues/100#issuecomment-64142901 and https://github.com/jcupitt/libvips/issues/173

I believe @jberkenbilt maintains the [vips source package in Debian](https://packages.debian.org/source/vips).

Jay, if you're able to help us, I think `libgsf-1-dev` should be added as a mandatory dependency of `lipvips-dev` given `libgsf-1-114` is a mandatory dependency of `libvips38`.

jberkenbilt:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=780750

Thanks!

lovell:
Great, thank you Jay.

LinusU:
Nice work @jberkenbilt!

PR #175, opened 2015-02-27
feature: min

### LINKED ISSUE #174: Feature request: `.min()`
Thanks for the awesome library!

**request** I think it would be awesome if we could have a new function called `min()` which would be `max()`'s weird brother. It would take the shortest edge and scale that one inside the provided geometry.

**example** Given a image `test.jpg` with the dimensions `10800x2332` and the code `sharp('test.jpg').resize(1080, 1080).min()` it would resize the image to be `5002x1080`.

**use case** This is great when producing an image that is supposed to be viewed on a screen of a specific size, especially when dealing with panoramas.

Using something like `sharp('test.jpg').resize(1440, 1440).max()` works great for getting images that are `1400x1080` or `1080x1440`. But this only works when the images are `4:3`, processing a panorama will make a image that is only 233 pixels tall, this isn't very good when displaying on a screen.

LinusU:
*fixes #174*

This add a new feature `.min()` which preserves the aspect ration and resize the image to the minimum `width` or `height` specified.

On another note, we might want to make the documentation for `max()` and `min()` a bit clearer. This is how Mozilla describes the `background-size` values `contain` and `cover`:

> **cover**
> This keyword specifies that the background image should be scaled to be as small as possible while ensuring both its dimensions are greater than or equal to the corresponding dimensions of the background positioning area.

> **contain**
> This keyword specifies that the background image should be scaled to be as large as possible while ensuring both its dimensions are less than or equal to the corresponding dimensions of the background positioning area.

I think that if we draw some inspiration from here we can make it a bit more clear what they do. @lovell?

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,6 +332,12 @@ Preserving aspect ratio, resize the image to the maximum `width` or `height` spe
 
 Both `width` and `height` must be provided via `resize` otherwise the behaviour will default to `crop`.
 
+#### min()
+
+Preserving aspect ratio, resize the image to the minimum `width` or `height` specified.
+
+Both `width` and `height` must be provided via `resize` otherwise the behaviour will default to `crop`.
+
 #### background(rgba)
 
 Set the background for the `embed` and `flatten` operations.
--- a/index.js
+++ b/index.js
@@ -41,7 +41,7 @@ var Sharp = function(input) {
     heightPost: -1,
     width: -1,
     height: -1,
-    canvas: 'c',
+    canvas: 'crop',
     gravity: 0,
     angle: 0,
     rotateBeforePreExtract: false,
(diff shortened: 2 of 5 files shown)


LinusU:
The Snap CI build fails because of c7ccf6801d5b74215e15b32975d8a1a61e03b400, I don't think that I can be blamed for that :)

lovell:
This is a perfect patch, favour fortunes the brave, thank you!

Good call on simplifying the docs - leave it with me.

Snap CI is failing as it only provides gcc 4.4 by default and therefore isn't fully C++11 compatible. I've updated the environment to manually install gcc 4.7 each time but it doesn't seem to apply to Pull Requests.

LinusU:
Thanks for quick pull! :beers:

Would appreciate it if you could publish this as `0.9.4` to npm :)

lovell:
v0.9.3 will be published within the next hour - currently fixing CI environment problems.

LinusU:
Understandable, thanks! I appreciate it

lovell:
All done, enjoy!

Issue #191, opened 2015-04-14
Progress on #118 (ignore aspect ratio)


skedastik:
Happy to say I'm making good progress on feature request #118. Everything is working as intended and all tests are passing. There is one complication, however.

In ResizeWorker::Execute, sharp applies a Gaussian blur prior to large affine reductions. To do so, it generates a square Gaussian image based on the standard deviation of the residual. This is fine if the aspect ratio is preserved, but I suspect a square kernel is not suitable for "non-aspect" reductions.

I'd love to hear ideas on how to proceed. Thinking out loud:

* Generate a Gaussian image based on the smaller residual (x or y), and box-shrink it to match the new aspect?
* The kernel should probably have __odd__ dimensions after the shrink.
* If I'm not mistaken, libvips' separable convolution won't work for a non-square kernel, meaning non-aspect reductions won't be as optimized.

Also, I have to confess ignorance as to why a blur is applied prior to large affine reductions in the first place. Feel free to clue this image-processing newbie in.

lovell:
Hi Alaric, good to hear you're making progress.

The Gaussian blur is to reduce the effects of aliasing when performing a large affine reduction. I suspect you'll still need this if either x or y residuals are < 1.

If you've not already tried it, a quick approach might be to use the average of x and y residual factors as the single residual input to the [sigma approximation](https://github.com/lovell/sharp/blob/judgement/src/resize.cc#L454). You'd need to manually inspect a few images to see how this looks, especially as the difference between x and y factors increases.

skedastik:
Good call on averaging the residuals. Here's the test image (512x512):

![original](http://i.imgur.com/WmFL1Bi.png)

And results of down-scaling to the following non-integral dimensions (cropped and magnified):

![results](http://i.imgur.com/T17tLbo.png)

The hard vertical edge is preserved in the first case which is perfect since no horizontal scaling occurs. The blur doesn't become perceptible until the x and y residuals reach more of a parity. These results were computed using the "nohalo" interpolator.

I'll go ahead with this approach in the provisional PR.

Issue #118, opened 2014-11-14
Feature request: ignore aspect ratio (was "Question: Un-uniform scale possible?")


jsongHBO:
Hi folks,

How do I resize an image disrespect of aspect ratio and without cropping and embedded padding to the exact size I want?
In gm's term, it'd be resize(width, height, '!');

Thank you

bkw:
that would be exactly what I need as well (See dupe #153)

jcupitt:
I haven't supported this feature in vipsthumbnail because it's hard to do correctly. 

Upscaling and downscaling are rather different operations and need separate code paths. If you allow hscale and vscale to differ, you at least double the number of code paths you need (both down, both up, hscale down vscale up, vscale down, hscale up), and perhaps double again, since (in some cases) downscale will already have two paths internally. 

Supporting not-very-different hscale and vscale would be pretty easy, which would probably cover most people. I suppose it could print a warning if the scales are either side of unity or if the scales differ by more than 20%.

bkw:
Thanks for the explaination, @jcupitt !

Unfortunately, for my (admittedly rather odd) usecase this would not be helpful. Maybe other users would still approve, dunno.

jcupitt:
Oh, that's a shame. How large a difference do you need?

bkw:
The thing is - I don't know! I'm mass-processing images from third party feeds and need to scale them down to a square for further processing. Since i have no control over the source images, the ratio could be anything.

If it helps anything: the target square has a size n*n with n being a power of 2, but never greater than 256:
```javascript
smallestEdge = Math.min(orig.width, orig.height, 256);
squareSize = Math.pow(2, Math.floor(Math.log(smallestEdge) / Math.log(2)));
```

jcupitt:
Wow that is a bizarre use case :) I guess this for some sort of query-by-image-content application? If that's the case I suppose you probably wouldn't care about small errors and a simple solution would be fine.

bkw:
You got that right :-)

I cannot really say how important "small errors" will be before I tried them out, but since the algorithm is resilient against compression artifacts and such things, it will probably be ok.

lovell:
Very happy to accept a PR implementing this feature - there's a fairly basic [guide for contributors](https://github.com/lovell/sharp/blob/master/CONTRIBUTING.md#submit-a-pull-request-with-a-new-feature).

blowsie:
[...]

PR #192, opened 2015-04-14
Support resize without preserving aspect ratio

skedastik:
# Goal

Add an `ignoreAspectRatio` method to support resizing of images to arbitrary aspect ratios, as requested in [#118](https://github.com/lovell/sharp/issues/118).

# Behavior

Fixed width and height:

```javascript
sharp('foo.jpg').ignoreAspectRatio().resize(300, 300);
```

Fixed width (leaving height unchanged):

```javascript
sharp('foo.jpg').ignoreAspectRatio().resize(300);
```

Fixed height (leaving width unchanged):

```javascript
sharp('foo.jpg').ignoreAspectRatio().resize(null, 300);
```

sharp's behavior is otherwise unchanged.

# Code paths affected

* src/resize.cc
    - Canvas
        - Additional member `IGNORE_ASPECT`
    - ResizeWorker::Execute
        - Logic is mostly unchanged. Add separate x and y variables for `factor`, `shrink`, and `residual`.
        - Move small amount of logic into private class methods:
            - `CalculateShrink`
            - `CalculateResidual`
        - Use average of `xresidual` and `yresidual` to compute `sigma` for pre-affine-reduction Gaussian blur (see #191).
* index.js
    - New method `Sharp.prototype.ignoreAspectRatio`
        - Sets `this.options.canvas` to "ignore_aspect"
* test/unit/resize.js

### DIFF
--- a/test/unit/resize.js
+++ b/test/unit/resize.js
@@ -259,5 +259,104 @@ describe('Resize dimensions', function() {
         done();
       });
   });
+  
+  it('Downscale width and height, ignoring aspect ratio', function(done) {
+    sharp(fixtures.inputJpg).resize(320, 320).ignoreAspectRatio().toBuffer(function(err, data, info) {
+      if (err) throw err;
+      assert.strictEqual(true, data.length > 0);
+      assert.strictEqual('jpeg', info.format);
+      assert.strictEqual(320, info.width);
+      assert.strictEqual(320, info.height);
+      done();
+    });
+  });
+  
+  it('Downscale width, ignoring aspect ratio', function(done) {
+    sharp(fixtures.inputJpg).resize(320).ignoreAspectRatio().toBuffer(function(err, data, info) {
+      if (err) throw err;
+      assert.strictEqual(true, data.length > 0);
+      assert.strictEqual('jpeg', info.format);
+      assert.strictEqual(320, info.width);
+      assert.strictEqual(2225, info.height);
+      done();
+    });
+  });
+  
+  it('Downscale height, ignoring aspect ratio', function(done) {
+    sharp(fixtures.inputJpg).resize(null, 320).ignoreAspectRatio().toBuffer(function(err, data, info) {
+      if (err) throw err;
+      assert.strictEqual(true, data.length > 0);
+      assert.strictEqual('jpeg', info.format);
+      assert.strictEqual(2725, info.width);
+      assert.strictEqual(320, info.height);
+      done();
+    });
+  });
+  
+  it('Upscale width and height, ignoring aspect ratio', function(done) {
+    sharp(fixtures.inputJpg).resize(3000, 3000).ignoreAspectRatio().toBuffer(function(err, data, info) {
+      if (err) throw err;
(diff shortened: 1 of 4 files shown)


lovell:
Hi @jsongHBO @bkw @blowsie, as requestors of #118 are you able to help test @skedastik's great work on the feature you need?

bkw:
Great work, @skedastik !
I noticed something odd with the size calculations, though:
Resizing a 610x600 image to 256x256 with ignoreAspectRatio() consistently results in a 256x260 image for me.

skedastik:
Thanks, @bkw.
I'm having a hard time reproducing the bug on my end. Running vips-7.42.1 under OSX. I'm using a 610x600 input image with combinations of JPEG/PNG/TIFF inputs and outputs and getting a 256x256 result consistently. I'll keep you posted...

bkw:
I just noticed I was a bit behind on my checkout of libvips. rebuilding with 7.42 now, will report back if it makes a difference.

bkw:
I can still reproduce it with this:

```javascript
var sharp = require('sharp'),
    request = require('request'),
    resizer = sharp().ignoreAspectRatio().resize(256,256);

request('http://i.imgur.com/dLKxTCk.jpg')
  .pipe(resizer)
  .pipe(sharp().metadata(console.log));
```

This gives the following output for me:
```
null { format: 'jpeg',
  width: 256,
  height: 260,
  space: 'srgb',
  channels: 3,
  hasProfile: false,
  hasAlpha: false }
```

Strange but true: Adding .gamma() produces the correct height of 256...

skedastik:
Curiouser and curioser. Reproduced on my system. Looking into it...

lovell:
Quick thought: possibly related to the use of gamma correction preventing JPEG shrink-on-load.

skedastik:
Right again, @lovell. There was a typo in the JPEG shrink-on-load conditional. In fixing it I realized we could retain the shrink-on-load optimization even if the aspect ratio changes. An improvement all around!

Nice work exposing the bug @bkw.

Fix has been squashed and pushed.

bkw:
I can confirm that for me it works as expected now. Thanks for the quick response, @skedastik !

test/unit/resize.js
```diff
@@ -259,5 +259,38 @@ describe('Resize dimensions', function() {
...
+  it('Fixed height ignoring aspect ratio', function(done) {
+    sharp(fixtures.inputJpg).resize(null, 320).ignoreAspectRatio().toBuffer(function(err, data, info) {
+      if (err) throw err;
+      assert.strictEqual(true, data.length > 0);
+      assert.strictEqual('jpeg', info.format);
+      assert.strictEqual(2725, info.width);
+      assert.strictEqual(320, info.height);
+      done();
+    });
+  });
```
lovell: Are you able to add a couple of tests for upscaling width but downscaling height and vice versa?

skedastik:
More unit tests: done!

lovell:
Fantastic, thank you Alaric. This will be included in the forthcoming v0.10.0 release.

PR #198, opened 2015-04-19
avoid segfault with unknown interpolator

bkw:
By accident I called `interpolateWith()` with an unknown interpolator name, which resulted in a segfault.
The attached patch avoids this by checking the return values of `vips_interpolate_new` and raising an exception if this returns NULL.


lovell:
Thank you, anything to remove the chance of a segfault is always much appreciated.

I'm going to make it fail fast in the JavaScript [interpolateWith](https://github.com/lovell/sharp/blob/master/index.js#L295) method too. (It's possible to compile some versions of libvips without support for some interpolators so the runtime guard you've added is still required.)

Issue #130, opened 2014-11-30
feature request: .extract() with %s instead of pixels


jonathanong:
perhaps allow %s as strings instead of numbers

lovell:
The `before` method hook under discussion in #236 should provide enough API to support this feature. It would allow the extract parameters to be set based on input image metadata.

lovell:
I'm going to close this request as #236 will provide the ability to do this (and a lot more).

PR #205, opened 2015-04-29
Allow arbitrary gravity shifts when cropping

ndziorlg:
In an CMS type application I develop I need to determine the gravity for the image for many different crop sizes.
For example, the person's head in the photography should always be visible and the thumbnail on the main page is different then the thumbnail in the list of articles. If it is placed in the upper left corner then I would have to choose gravity "north" or "west" dependent on the aspect ratio. This is the reason why set of "centre", "north", "east", etc. is not enough.

Additionally, I am trying to add visual tool to the editor to choose the most important part of the photo (always visible), with a single click. For this purpose I need to set arbitrary value for the gravity, e.g. in units of 0-100% in horizontal and vertical dimension.

### DIFF
--- a/index.js
+++ b/index.js
@@ -43,7 +43,7 @@ var Sharp = function(input) {
     width: -1,
     height: -1,
     canvas: 'crop',
-    gravity: 0,
+    gravity: { left: 0.5, top: 0.5 },
     angle: 0,
     rotateBeforePreExtract: false,
     flip: false,
@@ -132,11 +132,20 @@ Sharp.prototype._write = function(chunk, encoding, callback) {
 };
 
 // Crop this part of the resized image (Center/Centre, North, East, South, West)
-module.exports.gravity = {'center': 0, 'centre': 0, 'north': 1, 'east': 2, 'south': 3, 'west': 4};
+module.exports.gravity = {
+  'center': { left: 0.5, top: 0.5 },
+  'centre': { left: 0.5, top: 0.5 },
+  'north':  { left: 0.5, top: 0.0 },
+  'east':   { left: 1.0, top: 0.5 },
+  'south':  { left: 0.5, top: 1.0 },
+  'west':   { left: 0.0, top: 0.5 }
+};
 
 Sharp.prototype.crop = function(gravity) {
   this.options.canvas = 'crop';
-  if (typeof gravity === 'number' && !Number.isNaN(gravity) && gravity >= 0 && gravity <= 4) {
+  if (typeof gravity === 'object' &&
+      typeof gravity.left === 'number' && !Number.isNaN(gravity.left) && gravity.left >= 0 && gravity.left <= 1 &&
+      typeof gravity.top === 'number' && !Number.isNaN(gravity.top) && gravity.top >= 0 && gravity.top <= 1) {
     this.options.gravity = gravity;
   } else {
     throw new Error('Unsupported crop gravity ' + gravity);
--- a/src/resize.cc
+++ b/src/resize.cc
@@ -73,7 +73,8 @@ struct ResizeBaton {
   int width;
   int height;
   Canvas canvas;
-  int gravity;
+  double gravityLeft;
+  double gravityTop;
   std::string interpolator;
   double background[4];
   bool flatten;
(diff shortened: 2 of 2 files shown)


lovell:
Hello Marcin, thank you very much for your work on this enhancement.

I understand the need for additional north-east, south-east, south-west and north-west gravities. This would bring things line with *magick's `-gravity` flag. Perhaps these additional compass points should be exposed in the `sharp.gravity` object?

Are you able to provide some functional tests to help me understand what the slightly more complex variable (0.0 - 1.0) crop gravity might provide beyond the existing [extract](https://github.com/lovell/sharp#extracttop-left-width-height) feature?

ndziorlg:
I think the gravity feature does not provide new feature beyond extract in general. You can always first extract area with desired aspect ratio, then scale. With gravity you do not have to check the size of the photo first. For this reason adding noth-east, etc. would be valuable, because it does not require apriori knowledge of aspect ration.
In case of slightly more complex gravity with horizontal and vertical shift between 0-100% the additional benefit is smaller - but in my case I need it :) My use case:
Choose the position of the face in the portrait aspect ratio photo [http://en.wikipedia.org/wiki/Portrait#/media/File:VanGogh_1887_Selbstbildnis.jpg] - lets say 50% from left (horizontal center) 30% from top (shifted). For the crop into landscape banner (e.g. aspect ratio 4:3) gravity center could cut upper part of the head, and gravity top would leave to much white space above and cut the lower part of the face. I want to give full control to the editor of the post, so the photo will look good anywhere in the CMS system will show it.
Just like in the case of the simple gravity - you can always do it extract, but its is harder. You have to check the size of the photo while uploading and store it.

Gravity centre:
vangogh_1887_selbstbildnis-center
Gravity top:
vangogh_1887_selbstbildnis-top

lovell:
Thanks for providing more info about "variable gravity" @ndziorlg. Your scenario is similar to #130 and your addition would seem to make more sense as an improvement to the `extract` method. Any thoughts @jonathanong as originator of that request?

lovell:
@ndziorlg Sorry for the delay getting back to this PR and thanks again for your contributions.

Are you able to separate the features in your excellent work and, at least initially, include only the addition of new `northEast`, `southEast`, `southWest` and `northWest` attributes to the existing [sharp.gravity](https://github.com/lovell/sharp#cropgravity) Object, preferably with associated tests?

The next release is taking shape on the `knife` branch, so please use that as the PR target.

This should keep separate and therefore allow any deliberation on adding the % to crop vs extract (vs crop+extract) to continue unabated.

martinheidegger:
:+1:

lovell:
Hi @martinheidegger, thanks for the +1. Are you able to provide more details of which feature(s) you're referring to as there are a couple of improvements under discussion here?

martinheidegger:
@lovell Just read the code and really liked what I saw. Having the same properties as GraphicsMagick is a good step ahead. Since `sharp`'s api is already quite departed from `gm` in other regards it doesn't hurt to me to see the percentage option open. I found the solution clean and superior, so: :+1:

lovell:
@martinheidegger Thank you!

lovell:
I'm still very happy to accept a PR that adds support for north-west etc. gravities part of this PR.

Our golang sibling has just added this via https://github.com/DAddYE/vips/commit/6e5942207107df6a405dd72b5302d78814a40139

The addition of #236 should provide enough API for the variable cropping/extracting feature.

brandonaaron:
Might take a look at pulling in the additional gravities soon if there still isn't movement here.

brandonaaron:
I briefly looked at using the approach laid out here but it wasn't passing unit tests. In favor of time, I simply extended the existing approach to support these extra gravity settings in pull request #291.

lovell:
#291 has been merged, adding support for the additional gravity values, thanks @brandonaaron.

lovell:
v0.11.4 is now available via npm and includes the additional corner gravities part of this PR.

The variable gravity idea should be made possible via #236 - please feel free to continue the discussion there.

Thanks again for your time and suggestions @ndziorlg!

Issue #97, opened 2014-10-03
Feature request: overlay images with transparency (was: Resize, combine multiple images)


kembs:
Hello,

I was wondering if I could get help if this is possible with this library. I've been looking at the documentation and the code but can't find any info for combining multiple images on top of each other.

I assume I'd to do something like resize them beforehand and save them to buffers, then actually joining them?

Thanks

oaleynik:
@lovell I would be interested in blending two images of the *different* dimensions, perhaps taking advantage of an alpha transparency channel in the top image. I saw that vips has im_insert operation which could place one image over another at specified position. However, number of bands should be equal for both images or one of them should have one band.

This feature will let me fully switch to sharp and remove graphicsmagick fallback for this operation

lovell:
Hi @oaleynik, libvips provides an [ifthenelse](http://www.vips.ecs.soton.ac.uk/supported/current/doc/html/libvips/libvips-Hi conversion.html#vips-ifthenelse) operator that could be used for this - the transparency layer could be extracted and used as the `cond` image.

The API might look something like:
```javascript
var image1 = sharp('image1.jpg');
var image2 = sharp('image2.png');
image1
  .overlayWith(image2, top, left)
  .toBuffer(function(err, data) {
    ...
  });
```
In this example `data` would be in JPEG format as you're operating on `image1`. The `top` and `left` parameters could be made optional, where not providing them results in the images being centred.

Does this make sense?

(I'm planning to migrate to the new vips8 C++ API in 7.42.0+ later this year. This feature will be much easier to implement after doing so.)

oaleynik:
@lovell looks awesome! Exactly what I need and I think it will be useful for others.

nebkam:
If I understand correctly, this would make watermarking images a breeze, and fits my use case perfectly
:+1:

aseemk:
+1 :smile:

gasi:
FYI, I started exploring this and will try to report back on my progress later this week.

lovell:
Great news, thanks @gasi!

venox:
@gasi Looking forward to it. Thanks!

gasi:
@lovell: I have a proof-of concept and while it’s rough, I’d appreciate your feedback: https://github.com/gasi/sharp/blob/89e6dfbd3ce0fe776b85a768d0e9c809693a0af6/test/unit/composite.c

[...]

Issue #122, opened 2014-11-19
Investigate use of image hashing with automated tests


lovell:
http://phash.org/

Allow a % difference in the Hamming distance between fingerprints of actual vs expected?

lovell:
Also worth trying https://github.com/mgmtio/dhash-image and a hand-rolled frequency-sorting version.

lovell:
A dHash-based method produced the best results - see https://github.com/lovell/sharp/commit/c15994812756c206987d29102471e69c6bbd3fc1#diff-900514821d804f6815b2dcd954387ac4

Clean test run on [Linux](https://travis-ci.org/lovell/sharp/builds/60719491) and [Windows](https://ci.appveyor.com/project/lovell/sharp/build/34).

PR #207, opened 2015-04-30
API: `sharp.overlayWith` (alpha compositing)

gasi:
Adds `sharp.overlaysWith` for compositing an image on top of another image while handling alpha channels correctly.

I left all commits intact for better understanding but feel free to merge as one squash commit, @lovell.

This is now ready for review.

**Tests and usage examples:** `test/unit/overlay.js`

**TODO**
- [x] ~~Test with non-PNGs.~~ Non-4-channel images are not supported.
- [x] ~~Add support for specifying `top` and `left` position as in `sharp.overlaysWith(filename, top, left)`. This might be useful for watermarking.~~ Punted.
- [ ] Benchmark.
- [ ] Investigate (slight) discrepancy to ImageMagick `convert -composite` output. Not visible to human eye (at least not mine :wink:) but detectable by `compare …`.

**References**
- http://en.wikipedia.org/wiki/Alpha_compositing#Alpha_blending
- https://github.com/jcupitt/ruby-vips/issues/28#issuecomment-9014826
- https://github.com/gasi/sharp/pull/2

See: #97.


CONTRIBUTING.md
```diff
@@ -30,6 +30,8 @@ Please select the `master` branch as the destination for your Pull Request so yo
 
 Please squash your changes into a single commit using a command like `git rebase -i upstream/master`.
 
+To test C++ changes, you can compile the module using `npm run compile`.
```
gasi: This was useful to me and hope it will be to others :smile:

index.js
```diff
@@ -60,6 +60,9 @@ var Sharp = function(input) {
     gamma: 0,
     greyscale: false,
     normalize: 0,
+    // overlay
+    hasOverlay: false,
```
gasi: @lovell: If you show me how to test check if a string is empty in C++, we might be able to get rid of `hasOverlay`. On the other hand, having this flag might be more understandable instead of just relying on empty string `overlayPath`.

package.json
```diff
@@ -48,13 +49,14 @@
...
+    "gm": "^1.17.0",
+    "istanbul": "^0.3.13",
     "mocha": "^2.2.4",
     "mocha-jshint": "^2.2.0",
-    "istanbul": "^0.3.13",
-    "coveralls": "^2.11.2",
     "node-cpplint": "^0.4.0",
-    "rimraf": "^2.3.2",
-    "async": "^0.9.0"
+    "rimraf": "^2.3.2"
```
gasi: Alphabetized `devDependencies` for easier scanability and added `gm` to use `gm.compare` in automated tests.

src/composite.c
```diff
@@ -0,0 +1,119 @@
+#include <stdio.h>
+#include <vips/vips.h>
+
+/* Composite images `src` and `dst`.
+ */
+int Composite(VipsObject *context, VipsImage *src, VipsImage *dst,
```
gasi: Used `PascalCase` for `Composite` (used to be `composite`) to match the rest of the codebase.

gasi: I also tightened up whitespace to match this codebase’s style which is different from `libvips`.

src/composite.c
```diff
@@ -0,0 +1,119 @@
...
+
+  // Extract RGB bands:
+  VipsImage *srcRGB;
+  VipsImage *dstRGB;
+  if (vips_extract_band(src, &srcRGB, 0, "n", NUM_COLOR_BANDS, NULL) ||
+      vips_extract_band(dst, &dstRGB, 0, "n", NUM_COLOR_BANDS, NULL))
+    return -1;
+
+  vips_object_local(context, srcRGB);
+  vips_object_local(context, dstRGB);
```
gasi: Thanks for the advice with `vips_object_local`. Please check that I made that change correctly.

test/unit/overlay.js
```diff
@@ -0,0 +1,113 @@
...
+var gm = require('gm');
+var sharp = require('../../index');
+
+sharp.cache(0);
+
+
+// Helpers
+var tryDelete = function (file) {
+  try {
+    fs.unlinkSync(file);
```
gasi: I hope it’s ok I used a sync delete to save some typing.

test/unit/overlay.js
```diff
@@ -0,0 +1,113 @@
...
+    // Ignore missing files:
+    if (error.code !== 'ENOENT') {
+      throw error;
+    }
+  }
+};
+
+var compare = function (actual, expected, callback) {
+  // TODO: Allow for `tolerance` setting
+  gm.compare(actual, expected, function (error, result) {
```
gasi: I know this isn’t quite perceptual hashing #122 but it was helpful to me.

CONTRIBUTING.md
lovell: This will compile but not test the changes. I usually use `npm i` to do both.

index.js
lovell: std::string provides an [empty](http://www.cplusplus.com/reference/string/string/empty/) method.

src/composite.c
lovell: I try to keep the style JavaScript-like to help contributors from that arena :)

src/composite.c
lovell: Thank you. You should also be able to replace the use of `vips_object_local_array` and `g_object_ref` with matching pairs of `VipsImage *example` and `vips_object_local(context, example)`.

lovell:
Thanks again Daniel for all your work on this very useful and much-requested feature.

I'm not sure I want to go down the route of using *magick for image comparisons. My experience of managing CI environments warns me otherwise (I'm looking at you, [AppVeyor](https://ci.appveyor.com/project/lovell/sharp)) and, more importantly, I believe libvips can do this with minimum effort.

I'll spend some time this afternoon to see what I can do using libvips to get #122 off the ground.

test/unit/overlay.js
lovell: Perfect, tests are the right place to use simpler sync i/o.

lovell:
Commit c159948 on the `knife` branch adds a [fixtures.assertSimilar](https://github.com/lovell/sharp/blob/knife/test/fixtures/index.js#L78) method you can use instead of *magick's "compare".

By way of example, all of the tests for the [extract](https://github.com/lovell/sharp/blob/knife/test/unit/extract.js) method now employ it.

CONTRIBUTING.md
gasi: Oh, you’re right, I didn’t know `npm install` runs `node-gyp` on the package itself. I always tried `npm rebuild` but that only recompiles dependencies. However, you said it ‘will […] test the changes’. This doesn’t happen for me (see below). Am I missing something?

```
npm i

> sharp@0.10.0 install /Users/dani/workspace/sharp
> node-gyp rebuild

  CXX(target) Release/obj.target/sharp/src/common.o
  CC(target) Release/obj.target/sharp/src/composite.o
  CXX(target) Release/obj.target/sharp/src/utilities.o
  CXX(target) Release/obj.target/sharp/src/metadata.o
  CXX(target) Release/obj.target/sharp/src/resize.o
  CXX(target) Release/obj.target/sharp/src/sharp.o
  SOLINK_MODULE(target) Release/sharp.node
  SOLINK_MODULE(target) Release/sharp.node: Finished
```

lovell: Sorry, my mistake. Judging by the command line history `npm i && npm t` will do both.

gasi: No worries, I documented `npm install` and `npm test` :+1:

index.js
gasi: :+1: That worked, thanks.

src/composite.c
gasi: :+1: to making it more familiar for JavaScript developers. Is `PascalCase` more common for C++ than `camelCase`?

src/composite.c
gasi: Do you mean using explicit (named) references for `t[0]`, `t[1]`, and `t[2]`? This suggestion originally came from John Cupitt in https://github.com/jcupitt/ruby-vips/issues/28#issuecomment-97049450. He used `t[…]` for all intermediate result and I thought it might be valuable to use named references where possible. What do you think? Sorry, if I misunderstood your suggestion.

test/unit/overlay.js
gasi: Thanks for `fixtures.assertSimilar` and your example use of `toBuffer`. This enabled me to kill the intermediate files altogether :smile:

gasi:
> Commit c159948 on the knife branch adds a fixtures.assertSimilar method you can use instead of *magick's "compare".

Awesome! Thanks for providing this and I agree it’s more elegant without the GraphicsMagick dependency. However, I have two observations:

1. I found a case where the similarity test should fail: c9301c1fb5790c34adca1eb17fb1c1f953fdd76f
2. Isn’t it valuable for some operations like alpha compositing to do a comparison using no tolerance? If you agree, I could add that as parameter to `assertSimilar`.

.jshintrc
```diff
@@ -4,6 +4,7 @@
   "maxparams": 4,
   "maxcomplexity": 13,
   "globals": {
+    "before": true,
```
gasi: Although I don’t use this anymore, I figure it might be valuable for the future :smile:

src/composite.c
```diff
@@ -0,0 +1,119 @@
...
+  const int NUM_COLOR_BANDS = 3;
+
+  // These images will all be unreffed when context is unreffed:
+  VipsImage **t = (VipsImage **) vips_object_local_array(context, 3);
+
+  // Extract RGB bands:
+  VipsImage *srcRGB;
+  VipsImage *dstRGB;
+  if (vips_extract_band(src, &srcRGB, 0, "n", NUM_COLOR_BANDS, NULL) ||
+      vips_extract_band(dst, &dstRGB, 0, "n", NUM_COLOR_BANDS, NULL))
```
gasi: This should probably use `sharp::HasAlpha` and it might be useful to add `sharp::GetAlpha` similar to `get_alpha` in https://github.com/jcupitt/ruby-vips/issues/28#issuecomment-9014826. 

Currently, we only have the need to composite RGBA PNGs for our purposes. Do you think we could land this PR like this or should we keep it open until any n-band images are correctly handled?

src/composite.h
```diff
@@ -0,0 +1,17 @@
...
+#define SRC_COMPOSITE_H_
+
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+/*
+  Composite images `src` and `dst`.
+*/
+int Composite(VipsObject *context, VipsImage *src, VipsImage *dst, VipsImage **out);
```
gasi: Getting used to your 140 line length limit. I usually operate at 80 out of habit :wink:

src/resize.cc
```diff
@@ -790,6 +792,33 @@ class ResizeWorker : public NanAsyncWorker {
     }
 #endif
 
+    // Composite with overlay, if present
+    if (!baton->overlayPath.empty()) {
+      VipsImage *overlayImage = NULL;
+      ImageType overlayImageType = ImageType::UNKNOWN;
+      overlayImageType = DetermineImageType(baton->overlayPath.c_str());
+      if (overlayImageType != ImageType::UNKNOWN) {
+        overlayImage = InitImage(baton->overlayPath.c_str(), baton->accessMethod);
```
gasi: TODO: Handle non-transparent images more efficiently, e.g. using `vips_insert`.

test/unit/overlay.js
```diff
@@ -0,0 +1,62 @@
...
+      .toBuffer(function (error, data, info) {
+        if (error) return done(error);
+
+        sharp(data)
+          .overlayWith(fixtures.inputPngOverlayLayer2)
+          .toBuffer(function (error, data, info) {
+            if (error) return done(error);
+
+            // TODO: Remove once this fails
+            fixtures.assertSimilar(fixtures.expected('alpha-layer-012-low-alpha.png'), data, function (error, result) {
```
gasi: This `assertSimilar` should fail, but it doesn’t. Please see reason why it should fail below.

src/composite.h
lovell: People who use an 80 limit do so to leave room at the side of their monitor for Twitter.

gasi: Haha, good one :stuck_out_tongue_winking_eye: On a more serious note: I do it because I like to do three-way merges on my laptop :+1:

gasi: Then there is science :wink: http://baymard.com/blog/line-length-readability

test/unit/overlay.js
```diff
@@ -0,0 +1,62 @@
...
+
+sharp.cache(0);
+
+// Main
+describe('Overlays', function() {
+  it('Overlay transparent PNG on solid background', function(done) {
+    sharp(fixtures.inputPngOverlayLayer0)
+      .overlayWith(fixtures.inputPngOverlayLayer1)
+      .toBuffer(function (error, data, info) {
+        if (error) return done(error);
```
gasi: Minor: I noticed you use `if (error) throw error;` in your unit tests but doesn’t mixing synchronous `throw` inside an async callback cause issues? I am fine doing whatever you prefer in terms of style. Just let me know.

test/unit/overlay.js
lovell: The image fingerprinting method normalises image luminance so won't be able to detect all differences. It's worth trying your suggestion of exposing assertSimilar's threshold.

src/composite.c
lovell: To start with we can generate an explicit error if the image to be overlaid does not have 4 bands and an alpha channel.

test/unit/overlay.js
lovell: Yeah, good spot. I suspect `throw` still works as the `done` method will never get called but I definitely prefer your approach, thank you.

src/composite.c
lovell: Yes, using named references is preferable to an array. This avoids the chance of being out-of-bounds, which C does not guard against.

src/composite.c
lovell: PascalCase for classes and camelCase for variable names seems to be near-standard across C++, Java and JavaScript. The Google C++ guidelines use PascalCase for (non-accessor) method names, but I suspect that may have been influenced by the prevalence of C++ for Windows development in the 90s. The Sun Java guidelines use camelCase for method names, which JavaScript seems to have inherited. 

The only conclusion I've been able to reach is that consistency within a code base is more important than consistency across code bases.

test/fixtures/index.js
```diff
@@ -11,7 +11,7 @@ var getPath = function(filename) {
 
 // Generates a 64-bit-as-binary-string image fingerprint
 // Based on the dHash gradient method - see http://www.hackerfactor.com/blog/index.php?/archives/529-Kind-of-Like-That.html
-var fingerprint = function(image, done) {
+var fingerprint = function(image, callback) {
```
gasi: `callback` is a more common name than `done` in Node.js modules :+1:

test/unit/overlay.js
gasi: :+1:

src/composite.c
gasi: Done: 4c02c1d418fb4e7ca4208ef76c63019c930866f2 &  5edc24e77737b910aa81993d92f8cbe5f4ff9520

test/unit/overlay.js
gasi: Thanks, I added `threshold` option 23a151f20219a42d0615dfe0c851aee24607afae and it seems to have worked detecting that case using `threshold: 0`. However, I still wonder if it’d be valuable to have a identity check at an image level (ignoring metadata, etc) for testing subtle changes such as sharpening, opacity, etc.

src/composite.c
gasi: > The Google C++ guidelines use PascalCase for (non-accessor) method names, but I suspect that may have been influenced by the prevalence of C++ for Windows development in the 90s.

Thanks for sharing, I didn’t know that. C# uses the same convention.



> The only conclusion I've been able to reach is that consistency within a code base is more important than consistency across code bases.

I wholeheartedly agree and tried to match my new code as closely as possible :+1:

src/composite.c
gasi: Thanks for your advice, I changed it to use individual references.

gasi:
@lovell I made another pass on this today and it should be ready to merge given the restrictions we talked about (only RGBA images can be composited). I might be able to add more functionality at some point but for now, I think this might be already valuable for some people :smile:

src/composite.c
```diff
@@ -0,0 +1,128 @@
...
+
+  // Combine RGB and alpha channel into output image:
+  VipsImage *joined;
+  if (vips_bandjoin2(outRGB, outAlpha, &joined, NULL))
+    return -1;
+
+  vips_object_local(context, joined);
+
+  // Return a reference to the output image:
+  g_object_ref(joined);
```
lovell: I think this `g_object_ref` can be removed as `vips_object_local` will reference `joined`.

lovell:
Thanks again for all your work on this Daniel. Other than the possible reference leak, this is definitely ready to merge into the `knife` WIP branch. Are you able to squash to a single commit?

test/unit/overlay.js
aseemk: (Just FYI, Mocha catches synchronous errors during async test execution. That's necessary since every assertion library, including Node's own, throws errors synchronously.)

aseemk: (Via `process.on('uncaughtException')`)

gasi: Thanks for pointing this out, @aseemk. It makes sense in hindsight as the tests previously did work :smile:

gasi: It’s probably still nicer to _throw_ async errors via callback, right?

gasi:
> Other than the possible reference leak, this is definitely ready to merge into the knife WIP branch. Are you able to squash to a single commit?

Thanks, I simplified the `out` reference as you suggested and squashed everything into a single commit :smile: It’s been a pleasure collaborating with you and I hope I’ll be able to contribute more in the future.

test/unit/overlay.js
aseemk: I don't feel strongly, but think `throw`ing is maybe a bit better in test code — simpler, direct, and less room for bugs, all of which you particularly want in test code. But in this case, it's really up to @lovell. =)

lovell: The mocha [async examples](http://mochajs.org/#asynchronous-code) contain both `throw err` and `done(err)` variants. Reducing the number of conditional branches in test code decreases the need to test-the-tests. In this case, both approaches introduce one `if/else` split. Promises (and generators) will take over the world soon so we can, ahem, *throw* the error handling part of this test code away at some point.

gasi: :+1:

gasi:
I forgot to add `overlayWith` documentation to `README`. Will do that later today.

lovell:
No worries on the docs. I intend to add the ability to overlay onto non-alpha images too and will update the README after that. Thanks again @gasi for all your work on this new feature!

gasi:
Thanks, @lovell, I appreciate it :+1:

Issue #152, opened 2015-01-25
Take advantage of libvips v8.1.0+ features


lovell:
Features of v8.1.0+ we can take advantage of:

* [x] Remove all the version-specific pre-processor `#if` statements
* [x] Switch to new `vips_(un)premultiply` API for transparencies
* [x] Allow `normalize` operation on win32 - see https://github.com/jcupitt/libvips/commit/7495690cd6cf387c76a6043897bbbf765a0dc846
* [x] Use NanAssignPersistent/NanDisposePersistent to prevent GC of input buffer instead of taking a memcpy - ses https://github.com/jcupitt/libvips/commit/6ff1d5320b111a526224f4ade070b8e65313f4cf

Supplying pre-built binaries as part of #42 should happen first.

lovell:
This move can take advantage of the new `vips_resize` API - see also #126

gasi:
:+1:

lovell:
I plan to tackle this task in two stages:

1. Use libvips 8.1.x features via the C bindings, then (in a later release)
2. Switch to the C++ bindings

oaleynik:
@lovell :+1:

lovell:
Looks like Windows can't handle memory allocated by one shared library being freed in another - https://ci.appveyor.com/project/lovell/sharp/build/240 - will probably have to use [FreeCallback](https://github.com/nodejs/nan/blob/master/doc/buffers.md#api_nan_free_callback) to solve this.

lovell:
C++ migration moved to #299

lovell:
The `look` branch has now been merged to the `master` branch pending release.

PR #213, opened 2015-05-07
Premultiply alpha for transparent images to prevent dark edges

gasi:
This change premultiplies the alpha channel into the RGB channels for RGBA images to prevent dark edges when transforming (scaling, blurring, etc.) images. I opened this against `knife` as it had significant changes to `Composite` as we want to avoid double premultiplication.

On a less serious note: I am happy to take superior naming suggestions for `Unpremultiply`, but it might be actually pretty standard: https://msdn.microsoft.com/en-us/library/windows/desktop/hh780397%28v=vs.85%29.aspx Oh, ’muhrica :us: 

Supercedes https://github.com/lovell/sharp/pull/211.

I’ve done testing of this against ImageMagick with our own images output from *Paper*, but haven’t had a chance to do general regression testing with other kinds of inputs. This is a starting point and I’d be happy to polish it for inclusion in `knife` in the coming days. I placed the premultiplication logic before any image transformations and the unpremultiplication (sic!) before output. Please let me know if the placement can be optimized. The reference articles talk about having to watch out for zero alpha values (division by zero) and clamping, but I haven’t had a chance to explore that yet. I wonder if `vips_divided`, etc. take care of this internally, but maybe you know.

~~I did add the `file-compare` module as I needed really strict regression testing, although I realize it might fail due to different versions of `vips` creating the output. Happy to take suggestions for a better approach, although I think it needs to be closer to ImageMagick `compare` than the more loose perceptual hash. At some point, even with `tolerance: 0` it couldn’t distinguish two reference images with similar shape but different colors.~~ Added `sharp.compare` to compute mean squared error (though it doesn’t match the same numbers I get from `compare -metric mse` :cry:). At least it lets us detect whether images are identical at a pixel level.

References:
- https://github.com/jcupitt/libvips/issues/291
- http://entropymine.com/imageworsener/resizealpha/

/cc @julianwa


gasi:
Argh, I see the file hashing based comparison doesn’t pass on CI. Will look into it on Friday.

lovell:
Thanks Daniel, as always, for all your work on improving sharp.

To quote George Bernard Shaw, "England and America are two nations unpremultiplied by a common language."

I don't think `Premultiply` will have any effect on `vips_shrink` so it can occur after that, perhaps only when we know either an affine transform or composite will occur.

`vips_divide` [guards against](https://github.com/jcupitt/libvips/blob/master/libvips/arithmetic/divide.c#L165) divide-by-zero.

As for the functional tests, perhaps you could use `...extract( ... ).raw().toBuffer( ... )` to inspect specific pixels, comparing the Euclidean distance between two RGB values (technically sRGB, but probably close enough) against a given threshold.

lovell:
Thanks to John's sterling work, this improvement should now be able to take advantage of the new `vips_premultiply` and `vips_unpremultiply` operations available in libvips 8.1.0, wrapped in something like:
```cc
#if (VIPS_MAJOR_VERSION >= 9 || (VIPS_MAJOR_VERSION >= 8 && VIPS_MINOR_VERSION >= 1))
...
#endif
```
Are you able to make this update, which I guess should also apply to the Composite logic?

gasi:
Now that the cat is out of the bag—we launched [Think Kit](http://news.fiftythree.com/post/118777260503/introducing-think-kit)—I will dedicate today to polishing up this addition to `sharp`. Thanks again for all your help, @lovell!

> To quote George Bernard Shaw, "England and America are two nations unpremultiplied by a common language."

The team and I had a good laugh about that one :gb: :us: :stuck_out_tongue_winking_eye: 

> I don't think Premultiply will have any effect on vips_shrink so it can occur after that, perhaps only when we know either an affine transform or composite will occur.

Is that because it doesn’t do any interpolation? Either way, I am happy to give it a shot and see what happens. Is your main motivation for avoiding premultiplication where possible for performance reasons? If so, what’s the best way to capture a condition that says we will require it, i.e. `shouldPremultiplyImageAlpha`? Is there a better way than the following?

```c
bool shouldApplyAffineTransform = xresidual != 0.0 || yresidual != 0.0;
bool shouldComposite = !baton->overlayPath.empty();
// …
bool shouldPremultiplyImageAlpha = HasAlpha(image) && image->Bands == 4 &&
  (shouldApplyAffineTransform || shouldComposite /* || … */);`
```

> vips_divide guards against divide-by-zero.

Thanks for confirming.

> As for the functional tests, perhaps you could use ...extract( ... ).raw().toBuffer( ... ) to inspect specific pixels, comparing the Euclidean distance between two RGB values (technically sRGB, but probably close enough) against a given threshold.

Thanks, I’ll give this a shot. This sounds like *Magick’s `compare` with _mean square error_, right?

> Are you able to make this update, which I guess should also apply to the Composite logic?

Yes, I can do that too. Thanks for the snippet with the preprocessing statement :+1:

gasi:
Status update: I made the changes you suggested in `resize.cc`. Please take a look. I haven’t had a chance yet to actually compile & test using VIPS 8.1.0 for the built-in `vips_premultiply` / `vips_unpremultiply` changes. Additionally, I had this hope I could build my own *Magick `compare -metric mse` command using VIPS and expose it as `Sharp.compare`, but I got stuck on the lack of examples for `vips_stats` and `vips_getpoint`. In my last commit, I was trying to compute the stats for the difference of two input images but I am only getting `0` for `sum2`, etc.: https://github.com/lovell/sharp/commit/ca42ec70d7dd6b4fe108ddcabd8c4fa6563ea1fb

1. Do you agree it’d be valuable to have a VIPS based `Sharp.compare` for `fixtures.assertIdentical`?
2. Do you know of any examples on how to use `vips_stats` and `vips_getpoint`?

src/compare.c
```diff
@@ -0,0 +1,123 @@
...
+  printf("min: %f\n", (double) stats->data[0]);
+  printf("max: %f\n", (double) stats->data[1]);
+  printf("sum: %f\n", (double) stats->data[2]);
+  printf("sum2: %f\n", (double) stats->data[3]);
+  printf("avg: %f\n", (double) stats->data[4]);
+  printf("sd: %f\n", (double) stats->data[5]);
+  printf("xmin: %f\n", (double) stats->data[6]);
+  printf("ymin: %f\n", (double) stats->data[7]);
+  printf("xmax: %f\n", (double) stats->data[8]);
+  printf("ymax: %f\n", (double) stats->data[9]);
```
gasi: What’s the right way to read the values from `vips_stats`? If this is correct, why I am only getting `0` for `sum2` when running it with two different images?

src/composite.c
```diff
@@ -7,27 +7,28 @@ const int NUM_COLOR_BANDS = 3;
...
 
 /*
-  Composite images `src` and `dst`
+  Composite images `src` and `dst` with premultiplied alpha channel and output
+  image with premultiplied alpha.
  */
-int Composite(VipsObject *context, VipsImage *src, VipsImage *dst, VipsImage **out) {
-  if (src->Bands != 4 || dst->Bands != 4)
+int Composite(VipsObject *context, VipsImage *srcPremultiplied, VipsImage *dstPremultiplied, VipsImage **outPremultiplied) {
+  if (srcPremultiplied->Bands != 4 || dstPremultiplied->Bands != 4)
```
gasi: FYI, this didn’t require any changes in terms of `vips_premultiply` & Co., since it assumes that the input is premultiplied. I am wondering: Is there a way to store the fact that an image has a premultiplied alpha channel in `VipsImage`?

src/resize.cc
```diff
@@ -480,8 +480,28 @@ class ResizeWorker : public NanAsyncWorker {
...
 
+    // Premultiply image alpha channel before all transformations to avoid
+    // dark fringing around bright pixels
+    // See: http://entropymine.com/imageworsener/resizealpha/
+    bool shouldAffineTransform = xresidual != 0.0 || yresidual != 0.0;
+    bool shouldBlur = baton->blurSigma != 0.0;
+    bool shouldSharpen = baton->sharpenRadius != 0;
+    bool shouldTransform = shouldAffineTransform || shouldBlur || shouldSharpen;
+    bool hasOverlay = !baton->overlayPath.empty();
+    bool shouldPremultiplyAlpha = HasAlpha(image) && image->Bands == 4 && (shouldTransform || hasOverlay);
```
gasi: Do you like this approach or did you have something else in mind for efficiently deciding when to premultiply the alpha channel?

README.md
```diff
@@ -416,6 +417,10 @@ If the background contains an alpha value then WebP and PNG format output images
 
 Merge alpha transparency channel, if any, with `background`.
 
+#### Experimental: overlayWith(filename)
+
+**Experimental:** Composite image with transparent overlay. Both input and overlay image must be RGBA.
```
gasi: Added docs for this.

README.md
```diff
@@ -771,7 +787,7 @@ This module would never have been possible without the help and code contributio
 * [Brandon Aaron](https://github.com/brandonaaron)
 * [Andreas Lind](https://github.com/papandreou)
 * [Maurus Cuelenaere](https://github.com/mcuelenaere)
-* [Linus Unneb�ck](https://github.com/LinusU)
+* [Linus Unnebäck](https://github.com/LinusU)
```
gasi: There was some issue with the Unicode encoding of this file on Mac OS X. Hope that didn’t break anything for you.

src/compare-internal.c
```diff
@@ -0,0 +1,38 @@
...
+  if (vips_subtract(actual, expected, &difference, NULL) ||
+      vips_abs(difference, &absolute, NULL) ||
+      vips_stats(absolute, &stats, NULL))
+    return -1;
+
+  vips_object_local(context, difference);
+  vips_object_local(context, absolute);
+  vips_object_local(context, stats);
+
+  double *statsData = (double*) stats->data;
```
gasi: /ht @yoojin53 :smile:

src/compare-internal.c
```diff
@@ -0,0 +1,38 @@
...
+  VipsImage *absolute;
+  VipsImage *stats;
+  if (vips_subtract(actual, expected, &difference, NULL) ||
+      vips_abs(difference, &absolute, NULL) ||
+      vips_stats(absolute, &stats, NULL))
+    return -1;
+
+  vips_object_local(context, difference);
+  vips_object_local(context, absolute);
+  vips_object_local(context, stats);
```
gasi: Is it safe to free up `stats` here? I am asking because an oversight on my part caused a huge memory leak on our app: https://github.com/gasi/sharp/commit/f620e0bc2c5523481d6cfcf6f682412439dddcfa

src/compare-internal.c
```diff
@@ -0,0 +1,38 @@
...
+
+  double *statsData = (double*) stats->data;
+  int numValues = actual->Xsize * actual->Ysize * actual->Bands;
+  double sumOfSquares = statsData[STATS_SUM2_COLUMN];
+  double meanSquaredError = sumOfSquares / numValues;
```
gasi: Based on VIPS source code for `vips_stats` and [Wikipedia](http://en.wikipedia.org/wiki/Mean_squared_error), but I am getting different values than ImageMagick’s `compare -metric mse`. Either way, it at least is 0 if the input images match.

src/composite.c
```diff
@@ -115,11 +98,93 @@ int Composite(VipsObject *context, VipsImage *src, VipsImage *dst, VipsImage **o
...
+  return 0;
+}
+
+/*
+ * Premultiply alpha channel of `image`.
+ */
+int Premultiply(VipsObject *context, VipsImage *image, VipsImage **out) {
+  VipsImage *imagePremultiplied;
+
+#if (VIPS_MAJOR_VERSION >= 9 || (VIPS_MAJOR_VERSION >= 8 && VIPS_MINOR_VERSION >= 1))
```
gasi: Would you mind testing this on your end? Homebrew is still on `8.0.2.` :(

gasi:
@lovell Alright, this is ready for another review. Will squash the commits once you give me a :+1:

src/compare-internal.c
gasi: Again, turns out this was because ImageMagick premultiplies the alpha, and I didn’t in this case. Fixed: c8be9be738ff352dbe236a45787cbcbd8bffec12 :smile: 

Getting similar numbers to `compare -metric mse`, within 1%.

lovell: Thanks, good news, this is a handy addition to the test suite.

src/compare-internal.c
```diff
@@ -0,0 +1,45 @@
...
+int Compare(VipsObject *context, VipsImage *actual, VipsImage *expected, double *out) {
+  if (actual->Bands != expected->Bands)
+    return -1;
+
+  if (actual->Type != expected->Type)
+    return -1;
+
+  VipsImage *actualPremultiplied;
+  VipsImage *expectedPremultiplied;
+  if (Premultiply(context, actual, &actualPremultiplied) ||
```
lovell: Does the use of `Premultiply` here need wrapping in the `shouldPremultiplyAlpha` logic?

src/compare.cc
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    if (image1 != NULL && imageType1 != ImageType::UNKNOWN && image2 != NULL && imageType2 != ImageType::UNKNOWN) {
+      double meanSquaredError;
+      if (Compare(hook, image1, image2, &meanSquaredError)) {
+        (baton->err).append("Failed to compare two images");
+        return Error();
+      }
+      baton->meanSquaredError = meanSquaredError;
+      g_object_unref(image1);
```
lovell: I think you'll need to use `vips_object_local` to "hang" `image1` and `image2` on the `hook`, then use a final `g_object_unref(hook);` in the clean-up section to correctly unref everything.

src/composite.c
lovell: Will do. After 8.1.0 is out, and you've updated the Heroku buildpack, we can update sharp to rely exclusively on libvips' `vips_*premultiply` methods.

src/resize.cc
lovell: Your approach makes the logic clear, thanks.

test/fixtures/compare-im-references.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
lovell: Does the new `sharp.compare` method remove the need for these?

README.md
lovell: Great, thanks. Probably worth mentioning about dimensions having to match too.

test/unit/alpha.js
```diff
@@ -76,4 +75,30 @@ describe('Alpha transparency', function() {
...
+  it('Reduction with non-nearest neighbor interpolation shouldn’t cause dark edges', function(done) {
+    var BASE_NAME = 'alpha-premultiply-reduction-1024x768-paper.png';
+    var actual = fixtures.path('output.' + BASE_NAME);
+    var expected = fixtures.expected(BASE_NAME);
+    sharp(fixtures.inputPngAlphaPremultiplicationLarge)
+      .resize(1024, 768)
+      .interpolateWith('bicubic')
+      .toFile(actual, function(err) {
+        if (err) throw err;
+        fixtures.assertEqual(expected, actual, done);
```
lovell: The Windows CI is reporting `Expected images be equal. Mean squared error: 0.0003159345069434494.` so you'll probably need to allow a small MSE value via a threshold or similar.

lovell:
Thanks again for all your work on this Daniel. I've added a few comments inline. Sorry I wasn't able to make time to review this yesterday.

test/fixtures/compare-im-references.sh
gasi: Yes, will remove! :+1: These are still interesting to visually see where the differences are but I am ok to remove it to reduce the size of the project.

test/unit/alpha.js
gasi: Good catch, will add that :smile:

README.md
gasi: :+1:

src/resize.cc
gasi: :+1:

src/compare.c
gasi: I didn’t cast `stats` to `double *` correctly.

gasi:
> Thanks again for all your work on this Daniel. I've added a few comments inline. Sorry I wasn't able to make time to review this yesterday.

Thanks for your feedback and no need to apologise. I pushed some more improvements, but I haven’t checked if that was all since we had a little party for our _Think Kit_ launch.

A main thing I’d love your :eyes: on is, if there are any obvious or not so obvious memory leaks. We are seeing some in production and it’s likely from my `sharp` changes but not 100% confirmed. I tried running your leak tests but frankly haven’t fully understood how to interpret the results. I’d appreciate your help with that :gb: :heart: :us: Thanks :wink:

README.md
```diff
@@ -661,6 +666,17 @@ Provides access to internal task counters.
...
 
+#### Experimental: sharp.compare(filename1, filename2, callback)
+
+`filename1` and `filename2` are strings containing the filenames of the images to compare.
+
+`callback` is called with two arguments `(err, info)` where:
+
+* `err` contains an error message, if any.
+* `info` contains the info about the difference between the two images such as `isEqual` (Boolean), `meanSquaredError` (Number; present iff `status='success'`, otherwise `undefined`), and `status` (String; one of `success`, `mismatchedDimensions`, `mismatchedBands`, `mismatchedType`).
+
```
gasi: Added `status` to not report certain errors as fatal, e.g. mismatched dimensions, bands, etc.

src/compare-internal.c
gasi: I made `Premultiply` a noop (strictly: a copy) if the image has no transparency. I found it an example in John’s code.

src/compare.cc
```diff
@@ -0,0 +1,186 @@
...
+  }
+
+  void CleanUp() {
+    // Clean up any dangling image references
+    g_object_unref(hook);
+
+    // Clean up
+    vips_error_clear();
+    vips_thread_shutdown();
+  }
```
gasi: Added `CleanUp` function to centralise the logic.

src/composite.c
```diff
@@ -2,32 +2,42 @@
...
 const int NUM_COLOR_BANDS = 3;
 
+// TODO: Copied from `common.cc`. Deduplicate once this becomes a C++ module.
+int HasAlpha(VipsImage *image) {
+  return (
+    (image->Bands == 2 && image->Type == VIPS_INTERPRETATION_B_W) ||
+    (image->Bands == 4 && image->Type != VIPS_INTERPRETATION_CMYK) ||
+    (image->Bands == 5 && image->Type == VIPS_INTERPRETATION_CMYK)
+  );
+}
```
gasi: I realize it’s less convenient to have my code be straight C as I can’t reuse your function (as far as I know). Using C wasn’t a strong choice, but simply what I was slightly more familiar with, due to John’s VIPS examples.

test/unit/compare.js
```diff
@@ -0,0 +1,77 @@
...
+      assert.strictEqual(info.status, 'success');
+      // ImageMagick reports: 42242.5
+      // `compare -metric mse 'actual' 'expected' comparison.png`
+      assert(41900 <= meanSquaredError && meanSquaredError <= 41950,
+        'Expected: 41900 <= meanSquaredError <= 41950. Actual: ' + meanSquaredError);
+      done();
+    });
+  });
+
+});
```
gasi: Added unit tests for `sharp.compare`.

test/unit/overlay.js
```diff
@@ -6,19 +6,78 @@ var sharp = require('../../index');
...
+
+// Helpers
+var getPaths = function(baseName, extension) {
+  if (typeof extension === 'undefined') {
+    extension = 'png';
+  }
+
+  var actual = fixtures.path('output.' + baseName + '.' + extension);
+  var expected = fixtures.expected(baseName + '.' + extension);
+  var expectedMagick = fixtures.expected(baseName + '-imagemagick.' + extension);
```
gasi: Since I already generated and checked in ImageMagick reference images, I use them in tests here.

test/unit/overlay.js
```diff
@@ -6,19 +6,78 @@ var sharp = require('../../index');
 
 sharp.cache(0);
 
-// Main
+
+// Constants
+var MAX_ALLOWED_IMAGE_MAGICK_MEAN_SQUARED_ERROR = 0.3;
```
gasi: :arrow_down: Turns out `sharp`’s results are really close to ImageMagick, as one would expect :wink:

gasi: Let me know if this fails again on other platforms or versions. We can adjust it upwards then.

src/resize.cc
```diff
@@ -840,6 +869,17 @@ class ResizeWorker : public NanAsyncWorker {
...
     }
 
+    // Reverse premultiplication after all transformations:
+    if (shouldPremultiplyAlpha) {
+      VipsImage *imageUnpremultiplied;
+      if (Unpremultiply(hook, image, &imageUnpremultiplied)) {
+        (baton->err).append("Failed to unpremultiply alpha channel.");
+        return Error();
+      }
+
```
lovell: Adding `vips_object_local(hook, imageUnpremultiplied);` here should remove the reference/memory leak.

src/compare.cc
lovell: +1

src/composite.c
lovell: No worries - it'll all be moved to C++ as part of #152.

lovell:
Running `npm test` with this PR on Ubuntu 14.04 using libvips master and its new (un)premultiply methods generates test failures beyond the `0.0005` MSE threshold:
```
  1) Alpha transparency Enlargement with non-nearest neighbor interpolation shouldn't cause dark edges:
     Error: Expected images be equal. Mean squared error: 0.028674960136413574.
      at test/fixtures/index.js:151:25

  2) Alpha transparency Reduction with non-nearest neighbor interpolation shouldn't cause dark edges:
     Error: Expected images be equal. Mean squared error: 0.008033929392695427.
      at test/fixtures/index.js:151:25

  3) Overlays Composite two transparent PNGs into one:
     Error: Expected images be equal. Mean squared error: 0.08248437941074371.
      at test/fixtures/index.js:151:25

  4) Overlays Composite two low-alpha transparent PNGs into one:
     Error: Expected images be equal. Mean squared error: 0.011143956333398819.
      at test/fixtures/index.js:151:25
```
Here's `output.alpha-premultiply-enlargement-2048x1536-paper.png` generated by the first test, which doesn't appear (to my eyes) to suffer from the dark edges seen previously:

output alpha-premultiply-enlargement-2048x1536-paper

Rather than increase the threshold, perhaps either extracting a region known to suffer regression/problems or using smaller test images will help.

src/composite.c
gasi: :+1:

src/resize.cc
gasi: Thanks! @yoojin53 taught me how to use Valgrind and I found this and another leak: da24175a0cd3ae258163db0db6c26b94a4ba953e

We still found quite a few reports about `ResizeWorkerExecute` but none of them matched a missing `vips_object_local`.

gasi:
Thanks for running the test with the latest VIPS version. Agreed, it doesn’t show black fringing.

> Rather than increase the threshold, perhaps either extracting a region known to suffer regression/problems or using smaller test images will help.

I thought about that, but wouldn’t we want to ensure output images have a low error against a visually verified reference image as a whole? We might not want to only catch black fringes but any kind of errors. In this case, we could allow an MSE of `0.1` which based on my understanding would be a `0.1 / 255 * 100 ~= 0.04%` error for an RGB image. What do you think?

More importantly, it might be interesting to investigate what causes the discrepancy. I am sure John was much more finessed in his code for the native `vips_*multiply` and maybe there is something we can backport to `sharp`’s `*multiply` functions to minimize the error.

src/composite.c
```diff
@@ -72,40 +82,22 @@ int Composite(VipsObject *context, VipsImage *src, VipsImage *dst, VipsImage **o
...
+  VipsImage *t2;
   VipsImage *outRGBPremultiplied;
-  if (vips_ifthenelse(srcAlpha, srcRGB, dstRGBPremultiplied,
-      &outRGBPremultiplied, "blend", TRUE, NULL))
+  if (vips_multiply(dstRGBPremultiplied, t0, &t2, NULL) ||
+      vips_add(srcRGBPremultiplied, t2, &outRGBPremultiplied, NULL))
     return -1;
 
+  vips_object_local(context, t2);
   vips_object_local(context, outRGBPremultiplied);
```
gasi: @lovell: Isn’t this another memory leak: When `vips_multiply` succeeds `t2` is allocated, but then `vips_add` fails and the function returns without ever calling `vips_object_local(context, t2)`. I originally combined multiple operations into one—`if (vips_* || vips_* || vips_*)`—based on @jcupitt’s pattern using `vips_object_local_array`: https://gist.github.com/jcupitt/abacc012e2991f332e8b#file-composite2-c-L16-L22. However in his case, all `t[…]` seemed to be pre-hooked onto `context` right away. No more extra `vips_object_local` calls afterwards. Then we switched back to using individual variables for better readability but I kept the `if (… || … || …)` pattern. Your code uses individual variables but only single operations per `if (…)`. Am I right in identifying this as a potential leak (in error cases) or did I miss something?

lovell: Yes, in the event of an error occurring within `vips_add`, `t2` would never be unreferenced.

I still believe using explicit references is safer even when it means having to spread multiple operations over multiple conditionals, although I may be suffering from a bout of "[once bitten](https://github.com/jcupitt/libvips/pull/224), twice shy".

gasi: Thanks for confirming. I agree, I’d rather play it safe and will try to rewrite it with multiple conditionals. Can’t wait for #152 :heart:

gasi: BTW, in ’Murica they say ‘Fool me once…’: https://www.youtube.com/watch?v=eKgPY1adc0A

lovell:
> "I thought about that, but wouldn’t we want to ensure output images have a low error against a visually verified reference image as a whole? We might not want to only catch black fringes but any kind of errors."

Yes, how about we test the image as a whole using a "general" threshold *and* test the region known to cause problems at a specific, lower threshold?

lovell:
> "I am sure John was much more finessed in his code for the native vips_*multiply and maybe there is something we can backport to sharp’s *multiply functions to minimize the error."

I suspect we're probably looking at difference due to cumulative rounding errors. John's approach avoids the band splitting/joining and seems to be a bit faster as a result, although I've yet to benchmark this.

src/composite.c
jcupitt: libvips uses this style, as a possible alternative:

```C
VipsImage **t = vips_object_local_array(..);
VipsImage *in;

in = ..->input;

if( vips_thing(in, &t[0], NULL ) || 
    vips_thing2(t[0], &t[1], NULL )
    return -1;
in = t[2];

if( vips_thing3(in, &t[2], NULL ) || 
    vips_thing4(t[2], &t[3], NULL )
    return -1;
in = t[3];

if (vips_image_write(in, out))
    return -1;

return 0;
```

This has some nice properties:

* you can use `.. || ..` to chain operations together, so it's concise
* guaranteed leak-free
* you only need to keep track of the annoying `t[]` variables with a `.. || ..` block -- the blocks are independent and joined with properly named variables
* because blocks are independent, you can reorder them safely: you could swap these two over and it would all work

jcupitt:
The vips test suite looks for `(a - b).abs().max() < threshold`, ie. maximum absolute difference, I don't know if that would work here. Testing average difference can miss things, and picking the right threshold is difficult.

src/composite.c
lovell: Thanks John, I'm hoping this is relatively short-lived as sharp's move to the new vips8 C++ bindings will remove all the reference counting, taking advantage of internal calls to `vips_object_local_array` :)

gasi: Converted all code to use individual error conditionals :+1:

gasi:
> @lovell: Yes, how about we test the image as a whole using a "general" threshold and test the region known to cause problems at a specific, lower threshold?
> @jcupitt: The vips test suite looks for (a - b).abs().max() < threshold, ie. maximum absolute difference, I don't know if that would work here. Testing average difference can miss things, and picking the right threshold is difficult.

Both might be worth pursuing as well and the second shouldn’t be hard to add to `Compare`. Unfortunately, I won’t be able to spend more time on this PR as I have to work on other areas. I changed the control flow to use the `sharp` default of single `if` per `vips` operation to certainly avoid memory leaks and squashed all commits into one for merging into `knife` :+1: 

Thanks again to both of you for your help :smile:

gasi:
```
  1) Normalization keeps an existing alpha channel:
     Error: timeout of 20000ms exceeded. Ensure the done() callback is being called in this test.
```

Not sure why this is failing. I didn’t make any changes to it.

lovell:
Great job Daniel, I'm happy to take it from here and will use John's `max` recommendation to sort out the tests.

Thanks again for you for all your work on this improvement.

gasi:
> Great job Daniel, I'm happy to take it from here…

Thanks for merging this and my pleasure :heart:

PR #255, opened 2015-08-19
crop: Permit specifying the gravity as a string

papandreou:
I have a [use case](https://github.com/papandreou/express-processimage/) where the gravity is taken from a query string and it's cumbersome to use the numeric values or do a lookup in `require('sharp').gravity`. Seemed like it would be easy (and desirable) to support the strings directly.

Will conflict slightly with #205, but it should be easy to resolve.


lovell:
Great, thank you Andreas.

PR #263, opened 2015-09-06
Update pipeline.cc

chrisriley:
In the upgrade to nan v2 the baton values for blurSigma, sharpenFlat, and sharpenJagged are being cast to int32_t causing blur(sigma) and sharpen(radius, flat, jagged) to fail with "parameters out of range" (at least with nodejs 0.12.7). This patch fixes the issue by casting these baton values to doubles.


lovell:
Thank you for finding and fixing this Chris!

It looks like I broke this in b7e0a6f. The functional tests failed to detect it as the [sharpen tests](https://github.com/lovell/sharp/blob/master/test/unit/sharpen.js) have yet to take advantage of the [fingerprint-based assertions](https://github.com/lovell/sharp/blob/master/test/fixtures/index.js#L15) used in most other tests - I'll fix this.

PR #266, opened 2015-09-11
(preinstall) add freya to 'supported' distros

roryrjb:
Simply to add `freya` to the recognised Linux distros in the __preinstall__ script. Freya (i.e. Elementary OS 0.3) is based on Ubuntu 14.04 and the script completes successfully.


lovell:
Thanks Rory.

(Ignoring the Appveyor CI network failures as this change does not affect Windows.)

roryrjb:
@lovell :+1:

Issue #239, opened 2015-07-11
Enhancements to overlayWith method


lovell:
The main aim here is to add support for Buffer input and allow for differing dimensions with an optional `gravity`. The API should look something like:

`withOverlay(image, options)`

where:

* `image` is a String filepath or Buffer image data
* optional `options` is an Object containing `gravity` e.g. `sharp.gravity.southeast`, defaults to `sharp.gravity.centre`

```javascript
sharp('under.jpg')
  .overlayWith('over.png', {
    gravity: sharp.gravity.northwest
  })
  .toBuffer()
  .then( function(data) {
      ...
  });
```

Resizing of the image to overlay can be carried out beforehand using the existing resize method.

Applying a padding to the image to overlay will be possible with #128. This too can be carried out beforehand.

(See #97 for the original work on the `overlayWith` feature.)

fentas:
I am wondering if there is any progress on this?
Especially of using an Buffer (text as watermark).

lovell:
@fentas I expect these enhancements to be much easier/faster to implement after #152.

lovell:
Commit d92ea31 on the needle/v0.14.0 branch enhances the existing `overlayWith` method by adding support for Buffer-based input, allowing differing formats (8-bit, 16-bit), allowing differing image sizes and therefore accepting a `gravity` option.

This can be tested via:
```sh
npm install lovell/sharp#needle
```

Many thanks to @chrisriley for his original PR #335 on which I based this work.

lovell:
v0.14.0 now available via npm, thanks again for all the help with this.

lemnisk8:
@lovell Please expose the replicate feature from Vips 
replicate( int across , int down )
It would be really useful to tile watermark images for overlay

lovell:
@lemnisk8 Please can you create a new issue for this suggestion along with some examples of how you think the API might look, so we can invite others to discuss and/or submit a PR.

PR #272, opened 2015-09-19
adding a transparent color mask to the image

leiyangyou:
some refactoring can be done on this by merging the hasOverlay and hasMask sections together


leiyangyou:
looks like the result of image checks relies on version of libvips?

lovell:
Hello Lei, thank you for your PR and thank you adding functional tests - always appreciated! The behaviour of (un)premultiply will be slightly different (better, faster) with libvips 8.1.0+, hence the CI failures.

Your suggestion to combine the proposed feature (the ability to overlay an image with a single, optionally-transparent colour) into the existing `withOverlay` feature makes sense to me.

If the existing `withOverlay` method could accept an input image as a (raw pixel) Buffer *and* scale it to the same dimensions as the underlying image, would this also meet your requirements?

leiyangyou:
Hi @lovell, thanks for creating this solid library :+1: 
If overlayWith can expand the 1 pixel image, that would work for my current scenario, but an additional overlayWithColor method may be handy if someone needs to be able to change the masking color quickly.

lovell:
Thanks for confirming my suggestion might work. It is however only a suggestion and therefore a long way from being available, whereas your code works!

How about we fold your proposed new behaviour into the existing `withOverlay` method, which currently accepts a filename String, so it can also accept an `rgba` Object? That should keep the API clean, allow for the possible enhancements I mentioned and might even have the benefit of reducing duplicate code in `pipeline.cc`.

leiyangyou:
That'd be great, only problem there is that we will need to somehow indicate
that it's a Color rather than filename. Do we parse for Color first and upon
failure treat it as a file?

Sent using CloudMagic [https://cloudmagic.com/k/d/mailapp?ct=pi&cv=6.0.64&pv=8.2]
On Mon, Sep 21, 2015 at 02:51, Lovell Fuller <notifications@github.com> wrote:
Thanks for confirming my suggestion might work. It is however only a suggestion
and therefore a long way from being available, whereas your code works!

How about we fold your proposed new behaviour into the existing withOverlay method, which currently accepts a filename String, so it can also accept an rgba Object? That should keep the API clean, allow for the possible enhancements I
mentioned and might even have the benefit of reducing duplicate code in pipeline.cc .

—
Reply to this email directly or view it on GitHub
[https://github.com/lovell/sharp/pull/272#issuecomment-141820811] .[https://github.com/notifications/beacon/AAGXzbBlpBoDUcMxMPP1sP9HJifYjZUQks5ozvetgaJpZM4GAZoR.gif]

lovell:
Ah, I forgot that the `color` module can accept a String as well as an Object. It looks like it throws an Error if it can't parse it so your approach should work.

I wonder how many people are going to pass a valid filename that is also a valid colour? I'm looking forward to that bug report ;)

leiyangyou:
I merged the new call with overlayWith, let's see if the test passes on CI

lovell:
This looks great, thank you. Are you able to generate the expected outputs using libvips 8.0.2 to keep the CI environments happy?

Once these are passing, please perform the well-deserved task of adding yourself to the contributors section of `package.json` then rebase/squash your changes into a single commit. Cheers!

docs/api.md
```diff
@@ -372,6 +372,14 @@ sharp('input.png')
   });
 ```
 
+#### maskWithColor(rgba)
```
lovell: These details should move to the `withOverlay` method. I'm happy to fix-up the docs if you're unsure about this.

src/pipeline.cc
```diff
@@ -712,47 +719,86 @@ class PipelineWorker : public AsyncWorker {
...
+
+        VipsImage *overlayColorUCHAR = NULL;
+        if (vips_cast(overlayColor, &overlayColorUCHAR, VIPS_FORMAT_UCHAR, NULL)) {
+          (baton->err).append("Unable to apply overlay with color");
+          return Error();
+        }
+        vips_object_local(hook, overlayColorUCHAR);
+
+        VipsImage *overlayColorUCHARsRGB = NULL;
+        if (vips_copy(overlayColorUCHAR, &overlayColorUCHARsRGB, "interpretation", VIPS_INTERPRETATION_sRGB, NULL)) {
```
lovell: Using `vips_colourspace` might make what's happening here be a bit more self-descriptive, but I'm unsure it'll be any faster/better.

index.js
```diff
@@ -211,13 +213,21 @@ Sharp.prototype.flatten = function(flatten) {
...
 Sharp.prototype.overlayWith = function(overlayPath) {
-  if (typeof overlayPath !== 'string') {
-    throw new Error('The overlay path must be a string');
-  }
-  if (overlayPath === '') {
-    throw new Error('The overlay path cannot be empty');
+  try {
+    var colour = color(overlayPath);
+    this.options.overlayColor = colour.rgbArray();
+    this.options.overlayColor.push(colour.alpha() * 255);
```
lovell: We might want to Math.round the alpha value as it gets clipped to an integer in C++ land. (I've spotted that the existing `background` method should do this also.)

leiyangyou: Thanks Lovell, I will get these fixed up in the next commit :)

Sent using CloudMagic [https://cloudmagic.com/k/d/mailapp?ct=pi&cv=6.0.64&pv=8.2]
On Fri, Sep 25, 2015 at 23:05, Lovell Fuller <notifications@github.com> wrote:
In index.js [https://github.com/lovell/sharp/pull/272#discussion_r40439276] :

> @@ -211,13 +213,21 @@ Sharp.prototype.flatten = function(flatten) {
>  };
>  
>  Sharp.prototype.overlayWith = function(overlayPath) {
> -  if (typeof overlayPath !== 'string') {
> -    throw new Error('The overlay path must be a string');
> -  }
> -  if (overlayPath === '') {
> -    throw new Error('The overlay path cannot be empty');
> +  try {
> +    var colour = color(overlayPath);
> +    this.options.overlayColor = colour.rgbArray();
> +    this.options.overlayColor.push(colour.alpha() * 255);


We might want to Math.round the alpha value as it gets clipped to an integer in
C++ land. (I've spotted that the existing background method should do this also.)

lovell:
@leiyangyou How are things going? Can I provide any help?

leiyangyou:
@lovell haven't had a chance to relook at this yet, but let me give this a go next weekend :)

lovell:
@leiyangyou Great to hear you're still keen. There are a few relatively big changes occurring the `look` branch that I hope to merge into `master` within the next week. You may want to wait until after that to only have to deal with conflicts, if any, once.

leiyangyou:
ok, thanks @lovell. I will wait till then then.

lovell:
@leiyangyou  The API for `overlayWith` has been extended via #239 and the internals have altered, mainly due to the switch to the vips8 C++ bindings. I'm going to close this PR as it won't be "mergeable" as is, but please do feel free to open a new PR if/when you're able to work on this feature again.

Thank you for all your work on this (so far...)!

Issue #42, opened 2014-05-27
Investigate reducing the dependencies to install vips


ide:
One of the hardest parts about using Sharp is setting it up. Being able to just say "npm install sharp" would make it easier for a lot of people to try it out.

Currently libvips isn't very easy to set up on an Amazon Linux box, and on Mac OS X it requires brew (there are some instances in which one may wish not to install brew). So if there were an easier way to build vips and it were integrated with Sharp's installation that could help a lot.

lovell:
#186 provides the a rather compelling solution to this.

lovell:
You'll be pleased to hear that the next priority for sharp is improving the situation with dependencies.

My current thinking is to try to build and publish a static library comprising libvips and its dependencies required by sharp. This should target at least 64-bit Linux and Mac OS. Windows can probably use the [pre-built static libraries](http://www.vips.ecs.soton.ac.uk/supported/current/win32/) without too much effort.

These can then be retrieved via HTTPS, decompressed and placed *locally* at `npm install` time via the use of a custom JS or Python (as node-gyp requires Python) script.

We should avoid static-linking licensing problems as libvips is LGPL2.1+, which includes LGPL3, which is compatible with Apache 2.0.

papandreou:
Very pleased to hear that! Currently I'm very challenged because our production platform doesn't have a sufficiently recent libvips, so this could be a life saver.

Please keep us posted about how it's going :)

lovell:
I'm currently building a Dockerfile to help isolate the compilation of libvips and its dependencies, the output of which should hopefully be a compressed bundle of static libraries.

Initially this will target 64-bit Linux and I'm aiming for compatibility with at least Centos 6 and Ubuntu 12.04, which probably covers >80% of the Linux user base.

The complete (and I mean complete, it even includes zlib) set of static library dependencies will involve a ~15MB download during `npm install`.

lovell:
Update: the static library approach was far too complex. I've now got a pre-compiled libvips and all of its dependencies as shared libraries installed locally (without root) on Ubuntu 12.04, which sharp is able to compile and link against. The compressed tarball is down to ~12MB.

Next step: test across a number of Linux flavours. It's probably going to be kernel v3.2+ only, which means skipping Centos 6. This also means we can drop the somewhat esoteric Snap CI and instead move all of the Linux-based CI into scripted Docker containers.

lovell:
[...]

PR #280, opened 2015-10-03
(preinstall) add wily to 'supported' distros

roryrjb:
Hey @lovell, similar to my previous _pull request_ here I'm adding `wily` (Ubuntu 15.10) to the list of distros, perhaps as it is still in beta (I'm using Beta 2) this is premature but I can confirm that the preinstall script works perfectly.


lovell:
Thanks again Rory. Subscribe to #42 if you're interested in what will probably become the replacement for the preinstall.sh script.

roryrjb:
Cool, will have a look.

PR #286, opened 2015-10-13
Clarify fileName argument by changing to path

rayshan:
I noticed we can do something like `.toFile("tmp/temp.jpg")`


lovell:
Great, thank you Ray.

I prefer your naming as it uses the `path` parameter consistent with Node's built-in [fs](https://nodejs.org/api/fs.html) module.

rayshan:
You're welcome!

Issue #169, opened 2015-02-24
aws lambda installation


salzhrani:
I am trying to use sharp within aws lambda ... I got it running on an EC2 instance and then copied the the `node_modules` with the script as explained [here](https://aws.amazon.com/blogs/compute/nodejs-packages-in-lambda/).. 
when invoking the function I get the following error:
```
module initialization error: Error: libvips.so.40: cannot open shared object file: No such file or directory
 at Module.load (module.js:356:32)
 at Function.Module._load (module.js:312:12)
 at Module.require (module.js:364:17)
 at require (module.js:380:17)
 at Object.<anonymous> (/var/task/node_modules/sharp/index.js:11:13)
 at Module._compile (module.js:456:26)
 at Object.Module._extensions..js (module.js:474:10)
 at Module.load (module.js:356:32)
 at Function.Module._load (module.js:312:12)
 at Module.require (module.js:364:17)
 at require (module.js:380:17)
 at Object.<anonymous> (/var/task/index2.js:4:13)
 at Module._compile (module.js:456:26)
 at Object.Module._extensions..js (module.js:474:10)
 at Module.load (module.js:356:32)
 at Function.Module._load (module.js:312:12)
END RequestId: 4446cc91-bc30-11e4-823c-439adf27cb88
REPORT RequestId: 4446cc91-bc30-11e4-823c-439adf27cb88	Duration: 79.25 ms	Billed Duration: 100 ms 	Memory Size: 128 MB	Max Memory Used: 12 MB	

Message
-------
module initialization error:Error: libvips.so.40: cannot open shared object file: No such file or directory: at Module.load (module.js:356:32)
 at Function.Module._load (module.js:312:12)
 at Module.require (module.js:364:17)
 at require (module.js:380:17)
 at Object.<anonymous> (/var/task/node_modules/sharp/index.js:11:13)
 at Module._compile (module.js:456:26)
 at Object.Module._extensions..js (module.js:474:10)
 at Module.load (module.js:356:32)
 at Function.Module._load (module.js:312:12)
 at Module.require (module.js:364:17)
 at require (module.js:380:17)
 at Object.<anonymous> (/var/task/index2.js:4:13)
 at Module._compile (module.js:456:26)
 at Object.Module._extensions..js (module.js:474:10)
 at Module.load (module.js:356:32)
 at Function.Module._load (module.js:312:12)
```

 Not sure how to get it running properly

lovell:
[...]

PR #287, opened 2015-10-14
Add --runtime_link=static

vlapo:
libvips (https://github.com/jcupitt/libvips/issues/196) use Libs.private, Requires.private options in pkg-config vips.pc files. In case static compiling libvips it is necessary use ```--static``` option with ```pkg-config``` command during compiling sharp. https://github.com/lovell/sharp/issues/169

npm install static example:
```PKG_CONFIG_PATH=/path_to_static_compiled_vips/lib/pkgconfig/ npm install --prefix=/sharp sharp --runtime_link=static```

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -1,6 +1,9 @@
 {
   'targets': [{
     'target_name': 'sharp',
+    'variables': {
+      'runtime_link%':'shared', 
+    },
     'sources': [
       'src/common.cc',
       'src/metadata.cc',
@@ -51,6 +54,11 @@
             'include_dirs': [
                 '<!(PKG_CONFIG_PATH="<(PKG_CONFIG_PATH)" pkg-config --cflags vips glib-2.0)',
                 '<!(node -e "require(\'nan\')")'
+            ],
+            'conditions': [
+                ['runtime_link == "static"', {
+                  'libraries': ['<!(PKG_CONFIG_PATH="<(PKG_CONFIG_PATH)" pkg-config --libs vips --static)']
+                }]
             ]
         }]
     ],


lovell:
Great, thanks for this. I ran into similar problems trying to statically link libvips' dependencies as part of https://github.com/lovell/sharp/issues/42#issuecomment-146356996

Does this change prevent the need to modify `vips.pc` as mentioned in https://github.com/lovell/sharp/issues/169#issuecomment-147994604 ?

Was the `runtime_link` name inspired by https://github.com/naturalatlas/node-gdal/pull/2 ?

For Linux users, the next sharp release (v0.12.x) should include install-time detection of libvips and, if not available globally via `pkg-config`, will download and locally use pre-built shared libraries. This might alleviate the need for bundling static libraries when deploying to AWS Lambda (but we'll need to test/verify this).

vlapo:
yeah, I tested it on my ec2 enviroment without modify ```vips.pc``` and all run.

and yes I use ```runtime_link``` from pull request you mentioned :)

I think install-time detection dont resolve aws lambda problem because you have to upload node_modules with all required modules in zip file to lambda enviroment. There is no ```npm install``` run in lambda functions.
Maybe I am wrong, I know nothing about your install-time solution. Do you mean local as local to sharp folder? Still have to create lambda zip package on ec2 enviroment. 
I think there is some space to "Using spark on AWS Lambda" guide (with ```runtime_link``` all install-time solution).

lovell:
> "Do you mean local as local to sharp folder?"

Yes, the libvips+deps shared libraries will live within `node_modules/sharp/lib/...`, which should hopefully work with Lambda's approach of requiring a .zip file containing `node_modules`.

> "I think there is some space to "Using sharp on AWS Lambda" guide"

If you're able to write a short guide to using sharp with AWS Lambda for inclusion within https://github.com/lovell/sharp/blob/master/docs/install.md that would be fantastic!

escholtz:
Trying to piece together the details across this thread and the others that are referenced. @vlapo Setup instructions would be very much appreciated (for both sharp and libvips. since it sounds like you aren't using the preinstall.sh script?). Would love to get sharp running on Lambda.

vlapo:
@escholtz I am working on pull request to install.md with full steps. But I have to test it again on fresh enviroment ec2 with all tests green.

@lovell Do you know release date of changes in this pull request? I mean, do you want install guide with this changes or with workaround from https://github.com/lovell/sharp/issues/169#issuecomment-147994604.

lovell:
@vlapo A guide to Lambda installation guide using static libraries would be most welcome. I'm happy to wait for this for inclusion with the next patch release, v0.11.4.

(The work-in-progress `look` branch is where I'm experimenting with bundling pre-built shared libraries, which may also work with Lambda, although I've just discovered it only provides Node v0.10).

PR #288, opened 2015-10-16
EXIF Orientation tags 2 and 4 were flipping instead of flopping

brandonaaron:
This pull request corrects the handling for the EXIF orientation tags 2 and 4. Prior to this fix 2 would be upside down and mirrored and 4 would just be upside down.


lovell:
Amazing work Brandon, thank you for improving the quality of this feature with the complete range of test fixtures too!

(Looks like the Appveyor CI environment is currently suffering from network problems... will merge when it's up and passing again.)

brandonaaron:
Thanks! :)

PR #291, opened 2015-10-27
Add corner gavity support

brandonaaron:
Adds new gravity options: `northeast`, `southeast`, `southwest`, `northwest`.


lovell:
Brilliant, thank you Brandon.

PR #303, opened 2015-11-17
Implemented greyscale thresholding

dacarley:
This PR implements greyscale thresholding -- convert all pixels in an image to either white, or black, depending on whether they are above or below a given threshold.

This operation is useful in a number of image-processing scenarios. In my case, I'll be utilizing it as part of a handwriting recognition system, which requires that the images are normalized in this sort of way.


dacarley:
The appveyor test run aborted with ELIFECYCLE on Node 5.0, with no detail.  Doesn't look like there's a way to force a re-run.  How would you like me to proceed?

dacarley:
Thanks for re-running.  Still failing however.  I looked up the error code:

```
npm ERR! code ELIFECYCLE
npm ERR! sharp@0.12.0 test-win: `node ./node_modules/mocha/bin/mocha --slow=5000 --timeout=30000 ./test/unit/*.js`
npm ERR! Exit status 3221226356
```

This appears to be "heap corruption".  However, it appears to be failing before it ever gets to a test that would execute my changes.

Any thoughts?

lovell:
Hi David, thank you for this PR exposing a very useful feature of libvips.

The `look` branch on the Windows CI environment has been segfaulting in this manner for Node v4 and v5 since I switched to wrapping an input Buffer with a v8::Persistent instead of memcpy-ing it. I'm still trying to reproduce this problem on a local Windows machine. I've run valgrind tests on Linux but not found anything obvious yet; work continues.

I'll add a few comments inline...

.editorconfig
```diff
@@ -0,0 +1,12 @@
...
+
+[*]
+indent_style = space
+indent_size = 2
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false
```
lovell: Thanks for adding this. Are there markdown files with trailing space? We should probably fix that.

index.js
```diff
@@ -328,6 +340,19 @@ Sharp.prototype.sharpen = function(radius, flat, jagged) {
   return this;
 };
 
+Sharp.prototype.threshold = function(threshold) {
+  if (typeof threshold === 'undefined') {
+    this.options.threshold = 128;
+  } else if (typeof threshold === 'boolean') {
```
lovell: Should `false` set `threshold` to zero?

src/pipeline.cc
```diff
@@ -686,6 +690,15 @@ class PipelineWorker : public AsyncWorker {
       image = extractedPost;
     }
 
+    // Threshold - must happen before blurring, due to the utility of blurring after thresholding
+    if (shouldThreshold) {
+        VipsImage *thresholded;
+        if (Threshold(hook, image, &thresholded, baton->threshold)) {
+            return Error();
+        }
+        image = thresholded;
```
lovell: `vips_object_local(hook, thresholded);` is required here to prevent a memory leak.

test/unit/threshold.js
```diff
@@ -0,0 +1,70 @@
...
+      sharp(fixtures.inputJpg).threshold(-1);
+    });
+  });
+
+  it('invalid threshold 256', function() {
+    assert.throws(function() {
+      sharp(fixtures.inputJpg).threshold(256);
+    });
+  });
+});
```
lovell: Thanks for adding tests. Please can you try some of the other test fixtures as input (as well as JPEG), especially those with an alpha channel.

.editorconfig
dacarley: I'm not aware of any, but trailing whitespace can be significant in Markdown.

(Honestly, I copied the file from another of my own projects, and tweaked the indent settings to match your style)

src/pipeline.cc
dacarley: lol -- I did the same thing, twice, when reviewing the code myself after the heap corruption thing.

index.js
dacarley: Absolutely.  Fixed.

test/unit/threshold.js
dacarley: Sure thing.  I'll have an update pushed to the PR shortly.

lovell:
Thanks for the updates David, looking good. Are you able to rebase/squash to a single commit?

dacarley:
Squashed.  Should be good to go, once CI is complete.

lovell:
Great, thank you for this new feature.

PR #307, opened 2015-11-17
Linux Dockerfile: Use Debian Wheezy as the base image

papandreou:
Tested with Debian Wheezy and Ubuntu Vivid.


lovell:
Great, thanks Andreas. Inspired by this I might also test with Debian 6 as that should allow support for RHEL/Centos 6.

papandreou:
Thanks a lot!

lovell:
@papandreou I've gone a quick test with Debian 6 inside a Docker container (on Ubuntu 14.04) but it keeps freezing the CPU, so probably a no-go without a lot of extra effort.

papandreou:
Bummer, but cool that you tried it out. I'm still happy that Wheezy fell on the right side of the sweet spot :)

Issue #276, opened 2015-09-28
Parameters for extract are ordered unintuitively


papandreou:
The other day I spent a long time wondering why I couldn't get `extract` to do the right thing, only to find out that it wants `top`, `left`, `width`, `height`, whereas I intuitively put `left`, `top`, `width`, `height`. It works according to the documentation and all, but somehow I just assumed it would be X-Y-X-Y, and certainly not Y-X-X-Y.

I realize it'd be a hard thing to change, but I just wanted to check whether it's intentional? And if it's not, I move that `extract` is deprecated in favor of a differently named method that takes the parameters in the "correct" order ;). Mostly to save new users from the gotcha.

lovell:
Hej Andreas,

The underlying [vips_extract_area](http://www.vips.ecs.soton.ac.uk/supported/current/doc/html/libvips/libvips-conversion.html#vips-extract-area) method accepts `(left, top, width, height)` so you are definitely not alone in thinking that, intuitively, the parameters should be in that order.

Perhaps the `extract` method could be modified to (also) accept a single Object parameter with `left`, `top`, `width`, `height` as keys?

@brandonaaron originally implemented the much-requested `extract` feature in 83b72a1 so also may have relevant comments or suggestions.

papandreou:
> Perhaps the extract method could be modified to (also) accept a single Object parameter with left, top, width, height as keys?

That could be a fair compromise, except that none of the methods with similar purposes take options objects (`resize`, `rotate`, `crop` etc.). That could be fixed, of course.

I'm mostly concerned with the long time effects of having the weirdly ordered `extract` around in any shape or form. As long as it's there and documented, I predict that users to waste time figuring it out. I still think deprecating it, undocumenting it, and introducing a new method (`cut`, `shave`, `zoom`, ...) would be the best long term solution. Or maybe `crop` can be overloaded to also support 4 numbers, if that's not in conflict with any future plans for that method.

brandonaaron:
It seems really obvious after reading this that it should definitely be in that order. At the time I didn't think much of it. I just followed the example api laid out in this [comment](https://github.com/lovell/sharp/issues/23#issuecomment-53325646) without really thinking about it.

Another potential name for a replacement is `isolate`. However, I wonder if a new method would be more confusing than just making the method take an object. I think I'd vote for just documenting the oddity of argument order and have it take an object.

lovell:
[...]

PR #309, opened 2015-11-18
extract: Support an options object (take 2 against the look branch)

papandreou:
Also stop encouraging the use of .extract(top,left,width,height) as discussed on #276.


lovell:
Thanks again.

Don't worry about the Node v4 and v5 Windows CI test failures - there are a couple of Buffer-related bug fixes due in their respective next releases that I hope will prevent the heap-trampling we're seeing.

Issue #250, opened 2015-08-12
image input size limit


anandthakker:
Fantastic library--thanks for creating/maintaining!

I'm using it to resize some very large TIFs (> 1.5-2 billion pixels) and running into the input size limit.  I'm able to use `limitInputPixels()` to increase the pixel limit, but the highest value that works for me is `2^32 - 1 = 2147483647`.  My undersatnding is that this is because on the C++ side, the limit is read as a 32-bit integer [here](https://github.com/lovell/sharp/blob/01435977dea515d2fd371f444e4062c4a65cc4ff/src/pipeline.cc#L1149).  How feasible would it be to allow a larger value here?

lovell:
Hello Anand, your understanding is correct. At the moment we're using 32-bit signed for which the max is 2^31-1.

It should be a one-line change (plus tests, of course) to switch this to 32-bit unsigned via `->Uint32Value()`, allowing 2^32-1 = 4294967295.

Alternatively/additionally, we could modify `limitInputPixels` to also accept a boolean, where `false` = remove limit and `true` = reset to default limit.

(The limit exists to prevent resource hogging - see #115 for the background - although you do have to trust the image metadata.)

anandthakker:
Thanks for the quick response, @lovell.   I like the boolean option approach here, since, if someone's getting up above 2^31 pixels, it seems pretty likely that, like me, they actually just want no upper limit.  I'll try that out; happy to submit it back as a PR if you're interested.

lovell:
Great, PR gladly accepted, thank you for your kind offer!

kentongray:
@anandthakker did you have any luck with a pull request? I'd love to have this feature

anandthakker:
Ah, I unfortunately got pretty swamped soon after this, so I never did get
a chance to work up a PR. Sorry :-(

kentongray:
@anandthakker no problem I know how that is! I'll make a stab at it unless @lovell has any other thoughts :godmode:

lovell:
@kentongray Thank you, go for it! Feel free to post questions here if you need help.

mitselek:
I will do it unless I change my mind!

lovell:
This feature will be in v0.14.1, thanks @kentongray.

lovell:
v0.14.1 now available via npm, thanks both for your help with this.

PR #316, opened 2015-11-30
Allowing disabling of pixel limit

kentongray:
As discussed in #250 allowing boolean values for limitInputPixels this partially removes the limit. Unfortunately there are other limits based upon the int32 for width/height that I stumbled on shortly after so I'd like to take a stab at that but this gets you significantly larger images. Let me know your thoughts


lovell:
Hello, thanks for adding this feature. I'll add comments inline.

Are you able to update the [API docs](https://github.com/lovell/sharp/blob/master/docs/api.md#limitinputpixelspixels) to mention the acceptance of a boolean value?

I'm happy for any always-positive parameters to change from signed to unsigned integer if that helps too.

test/unit/io.js
```diff
@@ -805,6 +805,19 @@ describe('Input/output', function() {
       });
     });
 
+    it('Disabling limit works', function(done) {
+      sharp(fixtures.inputJpg).metadata(function(err, metadata) {
```
lovell: Is the use of `metadata` required here?

test/unit/io.js
```diff
@@ -805,6 +805,19 @@ describe('Input/output', function() {
...
+        if (err) throw err;
+        sharp(fixtures.inputJpg)
+          .limitInputPixels(false)
```
lovell: Please can you add the test case for `limitInputPixels(true)` too.

kentongray:
@lovell should be able to handle the noted things and docs, would you be opposed to adding a really large image to the test suite to test these things better?

lovell:
> "would you be opposed to adding a really large image to the test suite"

Great idea. As long as the compressed file size is not "really large", this would also provide a good example of libvip's low memory usage.

A single-colour, non-interlaced PNG (with row filtering disabled?) should compress the best. Decompression memory usage relates to image width rather than height, so perhaps try something like 2000px wide and whatever-you-need px high.

kentongray:
just an update my holidays were a little hectic but I do plan on getting back to wrapping this up in January

lovell:
@kentongray No worries, thanks for the update. Let me know if you need any guidance.

lovell:
The master branch is now using the vips8 C++ bindings so you'll notice a conflict or two when rebasing.

webbrandon:
This would be a great feature.  Is anyone currently working on @lovell comments?

How is the current permitted value calculated for `limitInputPixels`?  

I get a exception telling me the permitted range but assume it varies depending on the machine because I noticed others using higher values than I (current: 268402689).  Is this correct?

lovell:
@webbrandon Please go ahead and work on this if you're willing and able. `limitInputPixels` is currently set [here](https://github.com/lovell/sharp/blob/master/index.js#L34) to the maximum value allowed for WebP, which is the lowest common denominator when it comes to limits in supported formats.

kentongray:
Hey guys sorry for dropping off the planet on this one, but nothing will make progress like needed a feature (can't wait to use the padding). OK so I made the following changes.

1) Update docs
2) Added a large image  (20,000x14,000) and crushed it so it is 43k.
3) Updated tests to use the large image with true and false values

The only problem I potentially see is the large image test takes a bit (16267ms) on my machine. Not sure if that is acceptable or not.

kentongray:
i'm guessing the failing test is due to the large image... let me know your thoughts on how to proceed blarg really was trying to wrap this up today

lovell:
Loving the cyan, thanks for the updates @kentongray.

It looks like the new test is timing out on Windows CI. Are you able to double the `test-win` [limit](https://github.com/lovell/sharp/blob/master/package.json#L28) to 60000ms?

Please can you squash to a single commit (I've recently enabled PR squashing as the default behaviour but this PR was created before that, so you'll have to do it manually). More importantly, please add your good self as a [fully-fledged contributor](https://github.com/lovell/sharp/blob/master/package.json#L5).

kentongray:
unfortunately still failed it seems :sob:

lovell:
No worries, I'll look after the failing test (either add more time or skip on Windows).

Thanks again Kenton for all your work on this feature - this will be included in v0.14.1.

lovell:
@kentongray your test is now <50s on Windows CI with https://github.com/lovell/sharp/commit/71fb839e2ba46327f8fc6a69eb811f6f72b79205

kentongray:
yay thanks @lovell

webbrandon:
no problem @kentongray .  thank you very much for your contribution.

Issue #299, opened 2015-11-14
Switch to libvips C++ bindings


lovell:
The new `vips8` [C++ bindings](http://www.vips.ecs.soton.ac.uk/supported/current/doc/html/libvips/using-from-cpp.html) now expose all of the libvips' features this module uses so we can migrate away from the C bindings.

This will simplify the code and ease memory management.

Other benefits:

* [x] Detach the alpha channel before applying a gamma correction - see 36ac882 - fixed in 4dfbf89

binarykitchen:
yeah, that would be awesome to migrate away from the C bindings

lovell:
Commit 12d66ee on the `mind` branch is the first pass at this. It removed around 450 lines of code.

lovell:
Lots of warnings and linker errors on Windows, sigh. https://ci.appveyor.com/project/lovell/sharp/build/304/job/rct60twmmp2gds1a

lovell:
Thanks to the name mangling car crash that is cross-linking C++ libraries on Windows, we'll need to either include and compile libvips C++ wrapper from source or build libvips-cpp.dll via MSVC instead of mingw-w64.

lovell:
The vips8 C++ bindings are now being built as a DLL on Windows machines at `npm install` time. Looking much happier as a result. https://ci.appveyor.com/project/lovell/sharp/build/322

lovell:
v0.13.0 now available.

binarykitchen:
cool, well done!

Issue #128, opened 2014-11-27
Add the ability to `pad` an image


blowsie:
It would be great to see something much like the implementation here
https://github.com/EyalAr/lwip#user-content-pad

bennlich:
I am also in need of the pad functionality. I'd like to make square image tiles from rectangular images, adding transparent pixels for padding where necessary (i.e. to pad the shorter of the two dimensions). Is there a work-around with the current API?

lovell:
@bennlich you should be able to to do with the existing API via a combination of the `embed` and `background` methods. https://github.com/lovell/sharp/blob/master/test/unit/embed.js#L32-L37 provides an example of this.

bennlich:
Ah! Awesome, thanks a bunch for the quick reply.

bennlich:
Is there a way to specify where the image is embedded in the larger image (bottom left, top right, etc.)? Kind of like a gravity option for embed?

lovell:
@bennlich Reusing `gravity` to apply to the `embed` operation also would make a good addition, thanks for the idea.

@blowsie Does any of what @bennlich and I are discussing meet your needs, or should I move this new suggestion to a separate task?

blowsie:
I'm not sure without testing, which I don't have time to do right now.
I personally would love to see `.pad(10,10,10,10)` or similar as some sort of mixin rather than combining various other functions.

The embed function however does look useful for one of my use cases, which matches @bennlich s requirements of fitting rectangles in squares.

mikemliu:
I'm trying to achieve something similar, where I have a "zoom" or "crop" feature that lets the user specify how to process their image.

To embed a smaller image on a larger background, resorting to below. Wondering if it can be reduced to 1 pipeline instead of 2?

So, starting with a 50x50 image, i want to center it in a 256x256 background:
```
sharp(in)
.resize(50, 256)
.embed()
.pipe(sharp().resize(256, 256).embed())
.pipe(out)
```

The first resize essentially centers the image vertically on a 50x256 background, then a second pass centers it in the 256x256 background. Inefficient, but gets the job done. hoping for better alternative...

blowsie:
So after testing, using `embed` in combination with `background` caters for @bennlich s use case, however it doesnt let you add some padding  / whitespace to the image, as desired

bennlich:
Because of the lack of something like a "centered" gravity option?

blowsie:
[...]

PR #335, opened 2016-01-08
Overlay from buffer and overlay scaling

chrisriley:
Hi Lovell,

We had a use case for loading overlays from a buffer and also needed to be able to scale the overlay to a ratio of the main scaled image. These patches allow for overlays to be loaded from a node.js buffer in addition to a file and add `overlayGravity()`, `overlayRatio()`, and `interpolateOverlayWith()` methods. I wanted to put this out there in case anyone else might find it useful.

Regards,
Chris


lovell:
Thanks Chris, this is great start on the work required for #239. I need to think a bit more about the public API for this, e.g. does this need its own interpolator?

I'm currently moving sharp to use libvips' new C++ bindings as part of #299 and will take a closer look at how to manually merge in your work using its newer API once that's done. Cheers!

chrisriley:
Hi Lovell, I have an updated pull request (now with unit tests! :smile:) that includes a two additional methods, `overlaySize()` and `padOverlay()`. `overlaySize()` takes a width and height parameter and is used for exact control of the resized overlay, as opposed to `overlayRatio()` (which is still in there but is trumped by `overlaySize()` if both are used. `padOverlay()` accepts a boolean and adds a minimal amount of padding to the overlay's bounding box to keep it from running into the edges of the main image.

As for overlay's having their own interpolator, that was something I added because I wanted to be able to use bicubic for the background image and bilinear for the overlay.

Thanks!

lovell:
I've updated #239 with my thoughts on the API additions to improve the existing `overlayWith` method.

I'm not keen on including the ability to resize/interpolate/ratio the overlay image separately given there are already methods to achieve this. There are so many edge cases to resizing that I'd rather provide one way to do this.

The ability to pad an image, overlay or otherwise, should be satisfied by the long-standing feature request of #128.

The migration to the vips8 C++ bindings is almost complete. The [mind](https://github.com/lovell/sharp/tree/mind) branch contains this work and will form the basis of the future v0.13.0 release. By way of example, look how much simpler the internal [Composite](https://github.com/lovell/sharp/blob/mind/src/operations.cc#L14) method has become :)

if you'd like to have a go at getting the API outlined in #239 working with the C++ bindings on the `mind` branch, that would be most fantastic. I'm also happy to adapt and merge the work you've done so far here in this PR.

Most importantly, thanks again @chrisriley for all your ideas, code and time on this feature.

SimonSelg:
Hey @chrisriley,

Your code seems to be missing an 

```c
#include <math.h>
```

in `operations.c`, otherwise `round` (line 30 and 31) is not defined. Thanks for the code - you saved me some hours implementing the same functionality!

lovell:
I'm going to close this PR now as the features missing from sharp have now been reworked as #239, with the exception of `pad` covered by #128. Thanks again @chrisriley for all your work on this!

Issue #350, opened 2016-02-04
Crop: the docs imply that gravity is optional while it actually isn't


joelmukuthu:
```js
var sharp = require('sharp');
sharp().crop(); // throws "Error: Unsupported crop gravity undefined"
```
According to the [current docs](http://sharp.dimens.io/en/stable/api/#cropgravity), the above should work but it throws this error `Error: Unsupported crop gravity undefined`.

PR #351, opened 2016-02-04
Fix: default crop gravity to sharp.gravity.center

joelmukuthu:
Closes https://github.com/lovell/sharp/issues/350


lovell:
Great spot, thanks for fixing this Joel.

joelmukuthu:
Happy to do it :)

Issue #358, opened 2016-02-12
likely typo in clone()


jardakotesovec:
Hi, I was trying understand more how the clone function works. And seems that it just shares reference to buffer once original sharp instance finishes buffering of stream.

https://github.com/lovell/sharp/blob/677b2b90898a813cfca2f87fb9f8effbc487f83d/index.js#L887

```
Sharp.prototype.clone = function() {
  // Clone existing options
  var clone = new Sharp();
  util._extend(clone.options, this.options);
  clone.streamIn = false;
  // Pass 'finish' event to clone for Stream-based input
  this.on('finish', function() {
    // Clone inherits input data
    clone.options.bufferIn = this.options.bufferIn;
    clone.emit('finish');
  });
  return clone;
};
```

And seems to me that  `clone.streamIn = false;` does nothing, as its not changing `options` object and I don't see that this would be used anywhere. Maybe its just left over?

lovell:
Ahoj Jarda, that line looks like it was left over after developing the clone feature, sorry if it caused confusion. Thanks for the code review :smile:

lovell:
Thanks for the fix!

PR #361, opened 2016-02-17
Just small clean up. Addresses #358

jardakotesovec:


lovell:
Great, thanks for cleaning this up Jarda.

(The Circle CI packaging test timeout can be ignored for now - I need to sort this separately.)

PR #368, opened 2016-02-27
[ci skip] Improved Linux docs, Lambda instructions

### LINKED ISSUE #365: How to force external libvips
On MacOS X sharp uses the system provided libvips, but on linux it uses it's own vendored libs in `node_modules/sharp/lib`.

How do I force using the system provided libvips on Linux?

My use case is that I would like to use mozjpeg instead of the bundle libjpeg-turbo.

I couldn't find anything in the docs about this, they just tell that on most Linux systems no external dependencies are needed.

felixbuenemann:
* Add a more detailed explanation of how sharp discovers libvips
* Add a section on packaging sharp for running on AWS Lambda

This resolves #365 and is related to #169.


felixbuenemann:
Hmm, I wonder why CircleCI is ignoring the `[ci skip]` instruction, it [should be supported](https://circleci.com/docs/skip-a-build)…

lovell:
Marvellous, thank you Felix!

PR #374, opened 2016-03-08
Add libpango1.0-dev for trusty 14.04

daemonfire300:
I tried using the script on my my clean 14.04 trusty server, and I had to install libpango1.0-dev again, otherwise the Watermark tests etc would fail. See https://github.com/h2non/bimg/issues/63

### DIFF
--- a/preinstall.sh
+++ b/preinstall.sh
@@ -147,7 +147,7 @@ if [ $enable_openslide -eq 1 ] && [ -z $vips_with_openslide ] && [ $openslide_ex
       trusty|utopic|qiana|rebecca|rafaela|freya)
         # Ubuntu 14, Mint 17
         echo "Installing libopenslide dependencies via apt-get"
-        apt-get install -y automake build-essential curl zlib1g-dev libopenjpeg-dev libpng12-dev libjpeg-dev libtiff5-dev libgdk-pixbuf2.0-dev libxml2-dev libsqlite3-dev libcairo2-dev libglib2.0-dev sqlite3 libsqlite3-dev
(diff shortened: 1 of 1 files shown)


lovell:
Hi Julius, the `preinstall.sh` script is almost-deprecated for users of `sharp` as only 32-bit x86 Linux users still require it (and Centos 6 users, but only for a few more months).

I do however realise it's being promoted and used by other libraries that rely on `libvips` such as `bimg` and friends.

If `pango` is required for `bimg`, we should be consistent and install it on all flavours of Linux, not just for the use case of a specific Ubuntu version when also installing `openslide`.

Are you also able to add the correct `pango-dev`, `pango-devel` etc. packages for the other Linux OSs?

daemonfire300:
@lovell So what is the 'new' way of setting up `libvps` for use with libraries like sharp, bimg etc? I am sorry, I didn't notice, that I made the pull request to `sharp` and not `bimg`.

lovell:
`sharp` ships with pre-compiled binaries (Linux and Windows, x64 and ARM architectures) to allow non-root installation and keep up with the latest `libvips` improvements. These won't work out-of-the-box for the golang-based `bimg`.

You should be able to get away with using `sudo apt install libvips-dev` on Ubuntu [14.04](http://packages.ubuntu.com/vivid/libvips-dev) if the claimed libvips v7.40.0+ requirement of `bimg` is correct.

lovell:
Closing as I believe this is no longer required - feel free to re-open if I've got this wrong.

PR #384, opened 2016-03-25
Fix orientation that does not accept 8 and incorrectly accepts 0

jtobinisaniceguy:
Orientation value can be between 1 and 8, library is incorrect in an acceptable range 0 to 7


lovell:
Hello, the source branch for this PR is incorrect - I suspect you meant to select a branch from your own fork.

Github doesn't allow you to change the branches of a PR after it's created, so I'll close this and let you reopen a new PR from the correct branch/fork.

Issue #382, opened 2016-03-24
Setting orientation to 8 causes an exception


jtobinisaniceguy:
Library states that only values 0-7 are allowed.
Values 1-8 are allowed for orientation field

lovell:
The [logic](https://github.com/lovell/sharp/blob/master/index.js#L627-L628) and [docs](https://github.com/lovell/sharp/blob/master/docs/api.md#withmetadatametadata) both use 0-7 rather than 1-8. D'oh.

Great spot on this - I'm amazed no-one has run into this problem before - thanks for reporting.

lovell:
Happy to accept a PR if you're willing and able :)

lovell:
v0.14.0 now available via npm with your patch, thanks again for reporting this!

PR #385, opened 2016-03-26
Fix for orientation values 1-8

jtobinisaniceguy:
Included mocha locally to facilitate VS2015 with NTVS 1.1

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -519,7 +519,7 @@ This will also convert to and add the latest web-friendly v2 sRGB ICC profile.
 The optional `metadata` parameter, if present, is an Object with the attributes to update.
 New attributes cannot be inserted, only existing attributes updated.
 
-* `orientation` is an integral Number between 0 and 7, used to update the value of the EXIF `Orientation` tag.
+* `orientation` is an integral Number between 1 and 8, used to update the value of the EXIF `Orientation` tag.
 This has no effect if the input image does not have an EXIF `Orientation` tag.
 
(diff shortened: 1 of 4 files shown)


lovell:
Hello John, thanks for this fix for #382.

Are you able to update the relevant values in the [withMetadata docs](https://github.com/lovell/sharp/blob/master/docs/api.md#withmetadatametadata) also?

I'd very much prefer not to introduce `mocha` to `dependencies`; test dependencies belong in the `devDependencies` section. If there's a specific platform this doesn't work on, I suggest the use of `npm install -g mocha`.

jtobinisaniceguy:
Will do today

jtobinisaniceguy:
Will push later today

jtobinisaniceguy:
Interesting if I have mocha installed locally, all test pass
When installed globally I get this
  1) Input/output Handle Stream to Stream error :
     Uncaught Error: EPERM: operation not permitted, open 'C:\Users\User Name\Documents\OpenSource\Sharp\test\fixtures\output.jpg'
      at Error (native)

Some funny with mocha I believe

lovell:
Sorry, I think I may have confused things. Please leave `mocha` where it is, in `devDependencies`. This PR should not require any changes to `package.json`.

Thanks for updating the docs. Please can you rebase against upstream/master and squash into a single commit and we can get this into the forthcoming v0.14.0.

lovell:
Thanks for the changes, looking great. Are you able to squash into a single commit? I notice the commits are from (or at least are linked to) a different Github account - is this intentional?

lovell:
Landed in ebfc897, thank you!

PR #397, opened 2016-04-06
tweak xcode_settings

salzhrani:
This allows for a more verbose error message on mac os x
I have no idea why or how but it worked
```
libc++abi.dylib: terminating with uncaught exception of type vips::VError
```
becomes
```
libc++abi.dylib: terminating with uncaught exception of type vips::VError: VipsImage: memory area too small --- should be 1191960 bytes, you passed 1189440
```


salzhrani:
It seems that travis doesn't like this ... I don't think it is worth the trouble.

lovell:
I think the CI thing is only temporary - I'm re-running the failed builds.

lovell:
Thanks Samy, this is a useful debugging feature for OS X users.

Issue #387, opened 2016-03-28
Wrong output: resize (ignore aspect ratio) -> rotate 90 degrees


kleisauke:
Thanks for this awesome library, it's a great way to discover the speed and power of libvips.

After some testing I found a bug. See this testcase:
```javascript
var path = require('path');
var sharp = require('sharp');
var assert = require('assert');

sharp(path.resolve(__dirname, '2569067123_aca715a2ee_o.jpg')).resize(300, 300).ignoreAspectRatio().rotate(90).toFile(path.resolve(__dirname, '2569067123_aca715a2ee_o_2.jpg'), function(err, info) {
      if (err) throw err;
      assert.strictEqual('jpeg', info.format);
      assert.strictEqual(300, info.width);
      assert.strictEqual(300, info.height);
});
```

(No `AssertionErrors` are thrown but the output is wrong)

Output:
![Resize to 300x300 (ignore aspect-ratio) -> rotate 90 degrees - WRONG](https://cloud.githubusercontent.com/assets/12746591/14080615/94a337e4-f506-11e5-9516-df811a417327.jpg)

Expected output:
![Resize to 300x300 (ignore aspect-ratio) -> rotate 90 degrees - GOOD](https://cloud.githubusercontent.com/assets/12746591/14080649/cb99e518-f506-11e5-9928-bf04dfca2210.jpg)

Tested on Windows 10 x64, Node.js v5.9.1 with the latest Sharp (v0.13.1)

lovell:
Hallo Kleis, great spot, thanks for reporting this. It looks like the x and y ratios need to be swapped when ignoring aspect with a 90 or 270 degree rotation. Should be an easy one to fix - thanks for the test case.

lovell:
Commit 25b63a2 adds a test case that would previously have failed and the fix so it doesn't.

kleisauke:
Thanks for the fix! It's working when the width and height are equal. However when the width and height and not the same then width === height. See this testcase (method order is first rotate then resize):
```javascript
var path = require('path');
var sharp = require('sharp');
var assert = require('assert');

sharp(path.resolve(__dirname, '2569067123_aca715a2ee_o.jpg'))
  .rotate(270)
  .resize(500, 250)
  .ignoreAspectRatio()
  .toBuffer(function(err, data, info) {
    if (err) throw err;
    assert.strictEqual(500, info.width);
    assert.strictEqual(250, info.height);
    sharp(data).metadata(function(err, metadata) {
[...]

PR #399, opened 2016-04-06
Add a failing test for vips::VError exception

salzhrani:
There is a discrepancy between the raw buffer size that is emitted and expected. it is my understanding that a raw image buffer size should be the product of `width * height * channels` but it seems that there is a difference in rounding of the image dimensions. in the specific case in this test the buffer length should be `840 * 473 * 3 = 1191960` but instead `1189440` is emitted which seems to be the product of `840 * 472 * 3`


lovell:
Hi Samy, there was a recent problem highlighted by https://github.com/lovell/sharp/issues/387#issuecomment-203114575 where it was possible to get incorrect values for `info.width` and `info.height` passed to the callback. I think this might be related, in that `info.size` is correct but `info.height` is off-by-one due to a rounding that takes place after the height is calculated.

If so, the fix here is probably to ensure the `width` and `height` are based on the actual dimensions of the final output image rather than an intermediate calculation.

salzhrani:
It might be best if vips is the single source of image metadata if possible

lovell:
I agree, thanks Samy. Setting width and height on the `baton` approximately [here](https://github.com/lovell/sharp/blob/master/src/pipeline.cc#L658) should fix this. Happy to accept a PR for this if you're willing and able :)

salzhrani:
Thanks for the opportunity ... I also took the opportunity to add a try/catch.

lovell:
This looks great, thanks for the work on this.

What's the origin of the `flower.jpeg` test image? Are you able to add a source URL (you'll see these against most of the other test fixtures), assuming it's been made available for us to use under a suitable licence?

salzhrani:
The image can be found [here](http://eskipaper.com/phalaenopsis-wallpapers.html). Not sure about the licence. I can replace it with a another with the same dimensions and more explicit license if you want.

salzhrani:
Added a blank image with the correct dimensions.

lovell:
Fantastic work, than you very much for this Samy!

salzhrani:
@lovell If appropriate please cut us a 0.14.1 release. It seems all issues on the milestone are covered.

lovell:
@salzhrani v0.14.1 now available via npm, thanks for your additions!

PR #402, opened 2016-04-08
Add support for writing dz format to zip container

felixbuenemann:
This adds support for the `container` option of `vips dzsave`, which writes all generated deep zoom files to a ZIP container instead of the filesystem.

To enable it you can either use the `.zip`/ `.szi` file extensions or use `.tile({container: 'zip'})` with `.toFormat('dz')` or the `.dzi` extension.

My original idea was to also implement writing the dz format to a buffer using the ZIP container, but it's currently not possible, because Vips lacks the required `dzsave_buffer` function (see jcupitt/libvips#415).

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -527,13 +527,16 @@ The default behaviour, when `withMetadata` is not used, is to strip all metadata
 
 #### tile(options)
 
-The size, overlap and directory layout to use when generating square Deep Zoom image pyramid tiles.
+The size, overlap, container and directory layout to use when generating square Deep Zoom image pyramid tiles.
 
 `options` is an Object with one or more of the following attributes:
 
 * `size` is an integral Number between 1 and 8192. The default value is 256 pixels.
 * `overlap` is an integral Number between 0 and 8192. The default value is 0 pixels.
+* `container` is a String, with value `fs` or `zip`. The default value is `fs`.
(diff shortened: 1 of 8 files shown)


felixbuenemann:
@lovell Do you think the test coverage is sufficient? I could add some more in depth tests, but they would likely require adding a dev dependency on something like node-unzip or at the very least do some filetype magic checks. Without that I can't really prove that `.tile({container: 'zip'})` is working as expected.

lovell:
This looks great, thank you Felix.

I'm always happy to add modules to `devDependency` and improve test coverage. If you've got the time to add extra tests, that would be marvelous, but no worries if not.

felixbuenemann:
Great, I'll look into adding a test for the new tile option.

felixbuenemann:
OK, I've reworked the tests to unzip the created archive and inspect it using the existing `assertDeepZoomTiles` helper. This allowed me to add a test for `.tile({container: 'zip'})`.

It seems you can't use a custom file extension when writing to a ZIP file, because libvips always overrides it to `.zip`. Still the test shows that the container option is working as expected.

felixbuenemann:
The Travis build failure seems to be caused by an unrelated timeout on an overloaded os x worker, see:

https://travis-ci.org/lovell/sharp/jobs/121738571

Restarting the build on Travis CI should fix that.

lovell:
Thanks for adding the extra test cases Felix, this is a useful addition, great work.

PR #404, opened 2016-04-08
Speed up slow pixel limit tests

felixbuenemann:
This replaces the single color 20000x14000 PNG with a 20000x14000 Grayscale JPG.

On a Broadwell Core i7 @ 3.1 GHz this speeds up the "Disabling limit works" test from ~6.2s to ~0.15s because JPEG is generally faster than PNG and it can use the libjpeg resize-on-load optimizations.

This should fix occasional timeouts on Travis CI introduced by #250.


lovell:
Great, thank you Felix, it doesn't even reach the "slow" test limit now.

kentongray:
nice!

PR #419, opened 2016-04-27
Update Lambda instructions for Node.js 4.3

felixbuenemann:
Amazon introduced [Node.js 4.3 support for Lambda](https://aws.amazon.com/de/blogs/compute/node-js-4-3-2-runtime-now-available-on-lambda/), which is now the recommended runtime instead of the old Node.js 0.10. This commit revises the Lambda docs to build Node.js 4.3 compatible binaries using the latest stable [Node.js 4.x packages from Nodesource](https://nodejs.org/en/download/package-manager/#enterprise-linux-and-fedora).


lovell:
Thanks Felix. When Meteor finally gets its act together we can drop support for Node v0.10 completely.

Issue #437, opened 2016-05-24
Add an alpha channel before extending with a transparent background


frulo:
I'm trying to extend an image with a transparent background. But sharp extend the image with a black background. What is my mistake?

```
sharp( 'test.jpg' )
    .background( 0, 0, 0, 0 )
    .toFormat( sharp.format.png )
    .extend( {top: 0, bottom: 256, left: 0, right: 256} )
    .toFile( 'test.png' )
```

kind regards,

Orlando

lovell:
Hello, please can you try `.background({r: 0, g: 0, b: 0, a: 0})` - see http://sharp.dimens.io/en/stable/api/#backgroundrgba

frulo:
Hello, 
thanks for the fast response. I tried it with .background({r: 0, g: 0, b: 0, a: 0}) and have the same problem. If the  source image have already an Alpha value, than I can extend with a transparent  background. But with a jpeg as source I have this problem...

lovell:
Sorry, yes you're quite correct, `extend` doesn't currently add an alpha transparency channel.

It should be easy to add this feature, thank you for highlighting this.

frulo:
I have create a pull request therefore #439

lovell:
Thanks for the fix, this will be in v0.15.1.

lovell:
v0.15.1 now available, thanks again.

perekrestov-igor:
Looks like issue described above is still actual for webp format

PR #439, opened 2016-05-25
add Alpha Chane before extend imaging

frulo:
add alpha channel if source has no alpha channel before extending image.
Issue 437 [https://github.com/lovell/sharp/issues/437]


lovell:
Excellent, thank you very much for this. It's always a pleasure to see tests in a PR :)

PR #440, opened 2016-05-26
Adds the option of tiling/replicatng the overlay image across the image.

lemnisk8:
USAGE: .overlayWith(overlayimage.png, { tile: true} )

### DIFF
--- a/index.js
+++ b/index.js
@@ -92,6 +92,7 @@ var Sharp = function(input, options) {
     overlayFileIn: '',
     overlayBufferIn: null,
     overlayGravity: 0,
+    overlayTile: false,
     // output options
     formatOut: 'input',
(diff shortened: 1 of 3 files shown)


lovell:
Thank you for tackling this useful feature, this is a great start.

I think we should `ceil` rather than `floor` the number of tiles so they completely cover the image to be overlaid. This would mean supporting the case where the overlayImage, when tiled, is larger than the image to be overlaid, so vips_extract might help here.

Please can you add some test cases. This feature involves calculations based on width and height, so for example include the case where there's a 90 degree rotation in addition to a tiled overlay.

lemnisk8:
Thanks for the inputs, will do that...
Cheers

PR #443, opened 2016-05-26
Adds the option of tiling/replicatng the overlay image across the image.

lemnisk8:
USAGE: 
`.overlayWith(overlayimage.png, { tile: true, gravity: northwest} )`

Note: When using the tile option, the gravity option is applied to the extracted part of the tiled overlay image.


lovell:
Fantastic, thank you Chintan for this excellent addition.

Issue #435, opened 2016-05-22
Support for clipping (clipWith?)


kleisauke:
Hi there!

It would be awesome if this library supports clipping out-of-the-box. Currently I'm doing this with these steps:
- Convert the image to PNG (using sharp)
- Encode the image to a base64 encoded image (`data:image/png;base64`)
- Put the base64 encoded image inside in the SVG `xlink:href=` attribute (inside the `<image>` element which has for example the `clip-path="url(#clipCircle)"` attribute)
- Change the width and height of the SVG to the original PNG dimensions
- Save the SVG (example [here](https://gist.githubusercontent.com/kleisauke/16863e1b4aa49f2a8eb5c1357f9ab98c/raw/aac0225ea02858fe88aab4561d30b9deb0d18f17/circle.svg))
- Load the SVG into sharp and convert it back to PNG.

To simplify these steps it seems to me a good idea to come with a new API feature (this is just a draft):
> #### clipWith([options])
> 
> Clips an image to a certain path. The parts of the shape inside the path are visible, and the parts outside are invisible.
> 
> `options`, if present, is an Object, Buffer or String with the following optional attributes:
> 
> * `shape` is a String or an attribute of the `sharp.shape` Object e.g. `sharp.shape.circle` or `sharp.shape.ellipse`
> * Buffer containing SVG image data
> * String containing the path to an SVG image

For example I want to crop [this image](https://cloud.githubusercontent.com/assets/12746591/15454181/0e8a4ca0-2031-11e6-8cbe-22b687434d42.png) to a circle.
Then I just do this:
```javascript
sharp('500x500.png')
  .clipWith({ gravity: sharp.shape.circle })
  .toFile('500x500-circle.png', function(err) {
    // 500x500-circle.png is a 500 pixels wide and 500 pixels high image
    // containing a clipped to circle version of 500x500.png
  });
```

Which outputs [this image](https://cloud.githubusercontent.com/assets/12746591/15454184/364b8678-2031-11e6-8f67-cbbac66f9de0.png).

I don't know how much demand there is for this feature, but this will help for me to write more efficient and less hackish code. (The way I'm doing it right now is way too complicated :expressionless:)

URL's which might be helpful in order to accomplish this feature:
[draw: VIPS Reference Manual](http://www.vips.ecs.soton.ac.uk/supported/current/doc/html/libvips/libvips-draw.html)
[RSVG Libary Reference Manual](https://developer.gnome.org/rsvg/stable/)
[...]

PR #448, opened 2016-05-29
Add support for clipping/cutting out (#435)

kleisauke:
Usage: `overlayWith('overlayimage.png', { cutout: true } )`

Only the API Docs and changelog needs to be updated. Test cases are included.

Thanks to @jcupitt for the help provided [here](https://github.com/jcupitt/libvips/issues/59#issuecomment-222351004).


src/operations.cc
```diff
@@ -82,6 +82,70 @@ namespace sharp {
...
   /*
+    Cutout src over dst with given gravity.
+   */
+  VImage Cutout(VImage mask, VImage dst, const int gravity) {
+    using sharp::CalculateCrop;
+    using sharp::HasAlpha;
+    using sharp::MaximumImageAlpha;
+
+    bool maskHasAlpha = HasAlpha(mask);
+    bool dstHasAlpha = HasAlpha(dst);
```
lovell: I think `dstHasAlpha` will always be `true` as there's logic elsewhere to ensure the presence of an alpha channel.

src/operations.cc
```diff
@@ -82,6 +82,70 @@ namespace sharp {
...
+
+    if (!maskHasAlpha && mask.bands() > 1) {
+      throw VError("Image to be overlaid must have an alpha channel or one band");
```
lovell: Should this start "Overlay image must ..."?

src/common.cc
```diff
@@ -277,4 +277,17 @@ namespace sharp {
     return std::make_tuple(left, top);
   }
 
+  /*
+    Return the image alpha maximum. Useful for combining alpha bands. scRGB
+    images are 0 - 1 for image data, but the alpha is 0 - 255.
+  */
+  int MaximumImageAlpha(VipsInterpretation interpretation) {
```
lovell: Thanks for externalising this logic - something similar appears in quite a few places - I can now switch the other occurrences to use this.

src/pipeline.cc
```diff
@@ -693,10 +694,15 @@ class PipelineWorker : public AsyncWorker {
           // the overlayGravity was used for extract_area, therefore set it back to its default value of 0
           baton->overlayGravity = 0;
         }
-        // Ensure overlay is premultiplied sRGB
-        overlayImage = overlayImage.colourspace(VIPS_INTERPRETATION_sRGB).premultiply();
-        // Composite images with given gravity
-        image = Composite(overlayImage, image, baton->overlayGravity);
+        if(baton->overlayCutout) {
+          // 'cut out' the image, premultiplication is not required
```
lovell: The image to be overlaid has been pre-multiplied, so we'll either need to also pre-multiply here or not pre-multiply the image to be overlaid via the logic that calculates `shouldPremultiplyAlpha`.

lovell:
Fantastic, thank you for this Kleis. It's always a pleasure to see tests and I'm happy to update the docs etc. I've added a few comments inline.

src/operations.cc
kleisauke: You're right. https://github.com/lovell/sharp/blob/master/src/pipeline.cc#L448-L455

src/operations.cc
kleisauke: I'll fix that.

src/common.cc
kleisauke: :thumbsup:

src/pipeline.cc
kleisauke: What will be the best option? Will there be any difference in the image if we don't pre-multiply it?

lovell:
Updates look great, thank you.

> "What will be the best option? Will there be any difference in the image if we don't pre-multiply it?"

I think the best approach is to avoid premultiplying the image to be overlaid, assuming there are no other reasons for `shouldPremultiplyAlpha` being set to `true`, when using `cutout`.

This will avoid (a) a performance hit, (b) a source of rounding/clipping errors, and (c) potential light/dark artefacts at the "edges" of what was the discarded alpha channel.

jcupitt:
A general Porter-Duff compose operator could solve this problem too. vips will need PD at some point.

lovell:
@kleisauke Would you like to remove the premultiply logic here or would you prefer I take a look?

kleisauke:
@lovell I already did that. See: [pipeline.cc#L463-L464](https://github.com/kleisauke/sharp/blob/feature-cutout/src/pipeline.cc#L463-L464) and [pipeline.cc#L698-L700](https://github.com/kleisauke/sharp/blob/feature-cutout/src/pipeline.cc#L698-L700).

lovell:
@kleisauke You are absolutely correct - sorry I completely missed that. Thank you again for adding this very useful feature!

kleisauke:
@lovell No problem! Thanks for merging the pull request! :thumbsup:

Issue #442, opened 2016-05-26
corrupted double-linked list since version 0.13.0


kapouer:
Hi, here some back trace:

```
#15 0x00007ffff0d54534 in g_object_new () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#16 0x00007ffff10e2595 in vips_operation_new (name=0x7ffff111bb29 "VipsForeignLoadJpegBuffer") at operation.c:667
#17 0x00007ffff154f921 in vips::VImage::call_option_string (operation_name=operation_name@entry=0x7ffff111bb29 "VipsForeignLoadJpegBuffer", 
    option_string=option_string@entry=0x0, options=options@entry=0x7fffc80009f0) at VImage.cpp:463
#18 0x00007ffff1564981 in vips::VImage::new_from_buffer (buf=<optimized out>, len=2498, option_string=option_string@entry=0x0, 
    options=options@entry=0x7fffc80009f0) at VImage.cpp:558
#19 0x00007ffff1784e98 in PipelineWorker::Execute (this=0x1308fc0) at ../src/pipeline.cc:134
```

I'm using debian/sid (that currently has vips 8.3.1 and libjpeg62-turbo-1.4.2).
I'll have more time later this week: i could debug it with libasan, but it would be faster to have a clue about which lib i should rebuild with the address sanitizer option.

lovell:
I think this might relate to the ABI changes in newer versions of the g++ compiler.

http://sharp.dimens.io/en/stable/install/#linux

> Beware of Linux OS upgrades that introduce v5.1+ of the g++ compiler due to changes in the C++11 ABI. This module assumes the previous behaviour, which can be enforced by setting the _GLIBCXX_USE_CXX11_ABI=0 environment variable at libvips' compile time.

Are you able to share more of the back trace to help confirm?

luciferous:
@kapouer I ran into this same issue when updating from 0.12.0 to 0.15.0 on debian/sid. Ensure you're not installing libvips-dev via apt-get, just let the sharp installation fetch and build it instead. I'm not sure what the underlying issue is, but doing this fixed the reported issue for me.

lovell:
Thanks @luciferous, that helps confirm my suspicion the pre-compiled libvips binaries provided by Debian use `_GLIBCXX_USE_CXX11_ABI=1` (the default in gcc v5+), which conflicts with sharp's use of `_GLIBCXX_USE_CXX11_ABI=0` for backwards compatibility.

[...]

PR #456, opened 2016-06-07
Allow node-gyp configure --sharp-cxx11=1

kapouer:
Can also be set in .npmrc.
Closes: https://github.com/lovell/sharp/issues/442


lovell:
Merci beaucoup Jérémy, je vais changer la docs.

PR #465, opened 2016-06-14
Document undocumented BC break in v0.14.0

teohhanhui:
Deprecated style of calling extract was removed in 2034efcf558bbb0a6936b1804c0422103d43ad69

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -106,6 +106,10 @@ Requires libvips v8.2.3
   [#387](https://github.com/lovell/sharp/issues/387)
   [@kleisauke](https://github.com/kleisauke)
 
+* Remove deprecated style of calling extract API. Breaks calls using positional arguments.
+  [#276](https://github.com/lovell/sharp/issues/276)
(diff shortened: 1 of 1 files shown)


docs/changelog.md
```diff
@@ -106,6 +106,10 @@ Requires libvips v8.2.3
   [#387](https://github.com/lovell/sharp/issues/387)
   [@kleisauke](https://github.com/kleisauke)
 
+* Remove deprecated style of calling extract API. Breaks calls using positional arguments.
+  [#276](https://github.com/lovell/sharp/issues/276)
+  [@papandreou](https://github.com/papandreou)
```
lovell: Probably shouldn't blame papandreou for doing this - it was entirely my fault!

lovell:
Thanks for spotting and correcting this - sorry I failed to document this previously.

(A couple of the tests are currently failing on OS X due to image similarity thresholds being broken as a result of 7d261a1. I'll fix these in the next day or so.)

PR #473, opened 2016-06-23
Feature overlay top and left offset support

rnanwani:


lovell:
Thank you very much for this, looks very useful. I'll make time over the weekend to review in more detail.

index.js
```diff
@@ -364,7 +365,10 @@ Sharp.prototype.overlayWith = function(overlay, options) {
       throw new Error(' Invalid Value for tile ' + options.tile + 'Only Boolean Values allowed for overlay.tile.');
     }
     
-    if (isInteger(options.gravity) && inRange(options.gravity, 0, 8)) {
+    if(isInteger(options.left) && options.left >= 0 && isInteger(options.top) && options.top >= 0) {
```
lovell: Given the strictness in the rest of this function, I think if people try to define either `left` or `top` and either are invalid, it should also throw an Error.

test/unit/overlay.js
```diff
@@ -264,6 +264,146 @@ describe('Overlays', function() {
...
+    });
+
+    it('Overlay with offset and gravity and tile', function(done) {
+      var expected = fixtures.expected('overlay-offset-with-gravity-tile.jpg');
+      sharp(fixtures.inputJpg)
+        .resize(400)
+        .overlayWith(fixtures.inputPngWithTransparency16bit, {
+          left: 10,
+          top: 10,
+          gravity : 4,
```
lovell: Given the conflicting inputs, should this test fail with an Error?

lovell:
It's always a pleasure to see a PR with tests, thank you for adding these, it makes future maintenance of this feature so much easier.

I've added a couple of minor comments/nitpicks inline.

Are the `overlay-offset-with-tile.jpg` and `overlay-very-large-offset.jpg` test cases used?

index.js
```diff
@@ -8,7 +8,6 @@ var events = require('events');
 var semver = require('semver');
 var color = require('color');
 var BluebirdPromise = require('bluebird');
-
```
lovell: Nitpick: I like the separation between third and first-party dependencies.

src/operations.h
```diff
@@ -15,6 +15,22 @@ namespace sharp {
...
   /*
+    Alpha composite src over dst with given x and y offsets.
+    Assumes alpha channels are already premultiplied and will be unpremultiplied after.
+   */
+  VImage Composite(VImage src, VImage dst, const int x, const int y);
+
+  /*
+    Check if the src and dst Images for composition operation are valid
+  */
+  bool isInputValidForComposition(VImage src, VImage dst);
```
lovell: Nitpick: capitalisation on `IsInputValidForComposition`.

src/operations.h
```diff
@@ -15,6 +15,22 @@ namespace sharp {
...
+
+  /*
+    Given a valid src and dst, returns the composite of the two images
+  */
+  VImage compositeImage(VImage src, VImage dst);
```
lovell: Nitpick: `CompositeImage`

test/unit/overlay.js
rnanwani: if both top/left and gravity are present and valid, top/left takes preference over gravity. Neither of the inputs individually are invalid. So IMO, the test shouldn't fail. But should work as per the set preference.

lovell: Thanks, we'll need to document this :)

rnanwani:
The Appveyor CI is failing and I am not sure why. Let me know if something has to be fixed.

lovell:
The latest version of npm should fix the AppVeyor problem - see https://github.com/npm/npm/pull/13199

lovell:
This is looking great now, thank you.

Please can you remove the unused test expectations:
```
overlay-negative-offset.jpg
overlay-only-left-offset.jpg
overlay-only-top-offset.jpg
```
then rebase against upstream/master and squash to a single commit ready for merging.

rnanwani:
I created a new branch with all the changes in a single commit. I still get a few merge commits from the master branch. Any ideas how do I squash them?

lovell:
Squashed as 278273b - thanks again Rahul for all you work on this handy addition.

Issue #35, opened 2014-05-23
Survey: what are you using 'sharp' for?


lovell:
I'm truly humbled by the recent surge in support for `sharp`.

Whilst :star: and the trending status they bring are both amazing, thank you, the real measure of open source success is happy users.

I'd love to find out more about how you or your organisation are using (or plan to use) this module.

Feel free to add a comment with details of who you are, what you're doing with it and why you chose it. Alternatively let me know via [@lovell](https://twitter.com/lovell) on Twitter.

Thank you!

aronallen:
We are using sharp to deliver dynamically resized and retina optimized movie posters for the streaming services youseeplay.dk and blockbuster.dk.

Zewy:
I am working on replacing my php script that get the images from the ftp server and resize it. Nodejs and Sharp seems to use less resources then php.

gasi:
@FiftyThree uses **sharp** on our [Mix](https://mix.fiftythree.com/) backend to pyramid our user image uploads. We previously used ImageMagick and improved our image processing performance by several factors once we switched to **sharp**. Thanks for all your hard work on this, @lovell :heart:

Robert-Wett:
We are using Sharp at InVisionApp in increasing amounts as we replace legacy ColdFusion code with node microservices. We are using it for thumbnails as well as general resizes, and chose to use Sharp due to the performance gains compared to Image Magick

pajtai:
On the fly local and s3 image resizer.

x-cray:
Image validation/resize on upload and tile() for breaking big floor plans into tiles to show them with Leaflet.

cassieramen:
We're using sharp as the guts of an on demand image resizing service. Signal is a group video chatting app that allows you to browse photos, youtube videos, and the like synchronously with your friends. Internet together! https://signal.is/

asilvas:
Hey Lovell, great module!

We're using sharp at GoDaddy via an open-source image server, [ImageSteam](https://github.com/asilvas/node-image-steam). This allows us to serve only the bytes required by the device viewing the page, enabling much better user experiences.

masimakopoulos:
Thank you Lovell, for the great module. 
We at https://app.covve.com are using sharp as part of our avatar micro-service both for creating standard sizes for our common cases as well as transforming avatars on the fly for the not so common requests. [Covve](http://covve.com) is serving millions of contacts to its users and most of them have avatars, so we are putting your module to a good use.

[...]

PR #479, opened 2016-06-27
Add conv command to apply arbitrary convolution kernel to image

mhirsch:
This patch adds a `conv` operator to the sharp API so that an arbitrary convolution kernel can be applied to an image. The `conv` command has one argument that takes a structure containing all the ingredients for a vips convolution kernel, eg.:

`{ 'width': N
  , 'height': M
  , 'scale': Z
  , 'offset': Y
  , 'kernel':
             [[ 1, 2, 3],
              [ 4, 5, 6],
              [ 7, 8, 9]]
}`

One part of the patch that might need a little work: I couldn't find any documentation on testing whether a `VImage` object is empty, so I made a `bool` called `convKernelValid` that keeps track of whether the user has issued a `conv` command when building the pipeline. It would be cleaner to track this with the `VImage` object itself.


lovell:
Hello, thanks for making a good start on what will be a useful feature.

The instantiation of the `VImage` for the kernel needs to be within the `PipelineWorker::Execute` method to ensure any thread-local data lives on a libuv-managed thread (that we later invoke `vips_thread_shutdown` on), rather than the V8 main thread that the `NAN_METHOD(pipeline)` method runs on. This should avoid the need for the `convKernelValid` attribute.

Although the underlying operation is called `vips_conv`, I would prefer this behaviour be exposed via a function named `convolve`. The “conv" abbreviation can easily be confused with “conversion”, e.g. ImageMagick provides the `convert` tool.

I’d prefer the `kernel` attribute to be a flat array rather than an array of arrays. This more closely matches the underlying vips API and keeps the code simpler, e.g. removes inner loops. The API you’ve added (correctly) expects a width and height, so this can be used to validate the array length. Looping over the values in the kernel is still useful as we can add number validation here.

Please can you add tests to cover this work. I notice you added a couple of “make CI happy” commits. Were you able to run the tests locally as I’d expect these failures to show up there?

mhirsch:
I'll move the initialization of the kernel `VImage`. Thanks.

I'm happy to change the name to `convolve`. I'll look into adding tests -- I haven't done that before. I also have now figured out how to run the tests locally, so I hope I can avoid future commits chasing the test results.

src/operations.cc
```diff
@@ -221,10 +221,8 @@ namespace sharp {
 
     VipsImage *vips_matrix = kernel.get_image();
 
-    int i = 0;
-    for(int k : kernel_v) {
```
lovell: C++11 range based loops are [used elsewhere](https://github.com/lovell/sharp/blob/master/src/utilities.cc#L141) so this should work. What was the error?

mhirsch: I was trying to fix the mysterious appveyor failure. I suppose this has nothing to do with it. I can go back on this commit.

mhirsch:
@lovell thanks for all your feedback. Is there anything else left to do before you would be comfortable merging this?

src/operations.cc
```diff
@@ -212,6 +212,25 @@ namespace sharp {
   }
 
   /*
+   * Convolution with a kernel.
+   */
+  VImage Convolve(VImage image, int width, int height, double scale, double offset, std::vector<double> kernel_v) {
+    VImage kernel = VImage::new_matrix(width, height);
```
lovell: Might it be possible to use [new_from_memory](https://github.com/jcupitt/libvips/blob/master/cplusplus/include/vips/VImage8.h#L440) instead here? `NAN_METHOD(pipeline)` already contains a loop over the values of the kernel array - if it built a `char*` instead of a `std::vector` the loop and pointer logic would no longer be required here.

lovell:
Thanks for all the updates - one comment inline.

Are you able to use more extreme (or at least less tame) test cases, e.g. edge detection? The expected test cases are almost the same as the input so `fixtures.assertSimilar` won't be able to detect regression.

src/pipeline.cc
```diff
@@ -1149,6 +1160,22 @@ NAN_METHOD(pipeline) {
...
+  if(Has(options, New("convKernel").ToLocalChecked()).FromJust()) {
+    Local<Object> kernel = Get(options, New("convKernel").ToLocalChecked()).ToLocalChecked().As<Object>();
+    baton->convKernelWidth = attrAs<int32_t>(kernel, "width");
+    baton->convKernelHeight = attrAs<int32_t>(kernel, "height");
+    baton->convKernelScale = attrAs<double>(kernel, "scale");
+    baton->convKernelOffset = attrAs<double>(kernel, "offset");
+
+    size_t kernelSize = baton->convKernelWidth * baton->convKernelHeight;
+
+    baton->convKernel = std::unique_ptr<double[]>(new double[kernelSize]);
```
lovell: Does this need to be `delete`d after use?

mhirsch: No, `unique_ptr` will delete the pointer it's managing when it goes out of scope: http://en.cppreference.com/w/cpp/memory/unique_ptr

lovell: Thanks for confirming. I've not had the pleasure of using `unique_ptr` until now and seeing a `new` without a `delete` gave me the usual twitch. It looks like a rather useful feature of C++11, so thanks for introducing it to sharp!

lovell:
Thanks again Matt for all your work on this feature!

lovell:
It's always interesting to discover more about what people are using this module for in the survey at #35 should you be able to share details publicly.

mhirsch:
Thanks!

Issue #491, opened 2016-07-05
Feature request: Trim support


kevin1024:
Would love to have this feature, similar to ImageMagick.  It removes borders around images:

![](http://2.bp.blogspot.com/-qqiFk9xjlDo/UvpYznotYAI/AAAAAAAAAc4/5c13ph5acxg/s1600/x.jpg)

I noticed there is a blog post about how to implement it in libvips here, might be a good starting point!

http://libvips.blogspot.com/2014/02/trim-auto-crop-with-ruby-vips.html

lovell:
Hello, this feature, including sample code, has been mentioned previously at https://github.com/lovell/sharp/issues/295#issuecomment-203442516

The trim operation would probably make sense occurring before any resize operation.

As always I'm happy to accept a PR and/or provide help to anyone who'd like to attempt to implement this.

kleisauke:
I was trying to implement the trim operation with the [example](https://gist.github.com/jcupitt/eecd6d24692c388799e5) provided by @jcupitt. However it's currently not working (the input image is the same as the output). It would be great if someone could look at it. See: [tree#feature-trim](https://github.com/kleisauke/sharp/tree/feature-trim)

jcupitt:
Hi @kleisauke, you use `.project()` like this:

```C++
    VImage rows;
    VImage columns = mask.project(&rows);
```

Same for `.profile()`. It's a little awkward, unfortunately.

kleisauke:
Thanks @jcupitt! A pull request has been made for this feature at #492.

lovell:
In what will probably be the biggest "patch" release ever, the `trim` feature will be available in v0.15.1 thanks to the marvellous @kleisauke.

kevin1024:
This is amazing, thanks so much!

lovell:
v0.15.1 now available, thanks @kleisauke for the code and @kevin1024 for the suggestion.

PR #492, opened 2016-07-05
Add support for trimming (#491)

kleisauke:
Usage: `trim()` or defining a ~~threshold~~ _percentaged tolerance_ between ~~0-255~~ _0-100_: `trim(20)`

2 test cases included (a test with a 16-bit PNG and a normal PNG, both with an alpha channel). Trimming is always performed before any resize operation.

~~I was working on an optional argument defining the border(s) that should be trimmed away. By default the trimming is performed on all borders. Let me know if you want such additional argument.~~


src/operations.cc
```diff
@@ -392,4 +392,46 @@ namespace sharp {
...
+    // An equivalent of ImageMagick's -trim in C++ ... automatically remove
+    // "boring" image edges.
+
+    // We use .project to sum the rows and columns of a 0/255 mask image, the first
+    // non-zero row or column is the object edge. We make the mask image with an
+    // amount-different-from-background image plus a threshold.
+
+    // find the value of the pixel at (0, 0) ... we will search for all pixels
+    // significantly different from this
+    std::vector<double> background = image(0, 0);
```
kleisauke: Maybe we should let the users define the the point from where the trimming color is picked. Don't know how we should handle this (by a color or a position). For now it's always the top-left color.

jcupitt:
`Trim()` will scan the whole image once to find the crop area, so it won't work with sequential mode images. 

I guess you'd need to scan to find the crop area, then close and reopen to crop and shrink. This would halve speed for things like PNG.

lovell:
Wow, that was fast, thanks Kleis!

I don't think this needs to be overly-configurable for now, so let's have it (attempt to) trim all edges and always start with the top-left pixel.

A test with a 16-bit input image is probably a must given lossless PNG will likely be the format of choice here.

Making the `trim` operation mutually-exclusive with `sequentialRead` should alleviate the problem John mentions.

src/operations.cc
```diff
@@ -392,4 +392,46 @@ namespace sharp {
...
+    VImage profileRightV;
+    VImage profileRightH = columns.fliphor().profile(&profileRightV);
+
+    VImage profileTopV;
+    VImage profileTopH = rows.profile(&profileTopV);
+
+    VImage profileBottomV;
+    VImage profileBottomH = rows.flipver().profile(&profileBottomV);
+
+    int left = (int) profileLeftV.min();
```
lovell: Please can these use `static_cast<int>(floor( ... ))`.

jcupitt:
I suppose the threshold could be a percentage. That would avoid having to change it for 8 and 16-bit images.

src/operations.cc
```diff
@@ -406,7 +406,9 @@ namespace sharp {
 
     // we need to smooth the image, subtract the background from every pixel, take
     // the absolute value of the difference, then threshold
-    VImage mask = (image.median(3) - background).abs() > threshold;
+    // TODO: Not sure if this calculation supports tolerance. Or should we do this?:
+    // VImage mask = (image.median(3) - background).abs() > (MaximumImageAlpha(image.interpretation()) / 100 * tolerance);
```
jcupitt: I think I'd have this as `max * tolerance / 100`. I guess this should be uncommented too?

kleisauke: Thanks! Committed by mistake.

src/pipeline.cc
```diff
@@ -206,6 +207,12 @@ class PipelineWorker : public AsyncWorker {
         RemoveExifOrientation(image);
       }
 
+      // Trim
+      // TODO(kleisauke): Not sure if we should check if accesMethod is VIPS_ACCESS_SEQUENTIAL
```
lovell: How about adding logic to `NAN_METHOD(pipeline)` that when `trimTolerance` is set then we also force the `accessMethod` to `VIPS_ACCESS_RANDOM`?

test/unit/trim.js
```diff
@@ -0,0 +1,39 @@
...
+        assert.strictEqual(450, info.width);
+        assert.strictEqual(322, info.height);
+        fixtures.assertSimilar(expected, data, done);
+      });
+  });
+
+  it('16-bit PNG with alpha channel', function(done) {
+    sharp(fixtures.inputPngWithTransparency16bit)
+      .resize(32, 32)
+      .trim()
```
kleisauke: If I change this to `trim(100)` then it will fail:
```
Uncaught Error: extract_area: parameter width not set
extract_area: parameter height not set
```

Debug:
```
(sharp:6296): GLib-GObject-WARNING **: value "-32" of type 'gint' is invalid or out of range for property 'width' of type 'gint'

(sharp:6296): GLib-GObject-WARNING **: value "-32" of type 'gint' is invalid or out of range for property 'height' of type 'gint'
[Error: extract_area: parameter width not set
extract_area: parameter height not set
]
```

It looks like only values from 1-99 are permitted.

index.js
```diff
@@ -553,14 +554,31 @@ Sharp.prototype.threshold = function(threshold, options) {
...
+
+/*
+  Automatically remove "boring" image edges.
+    tolerance - if present, is a percentaged tolerance level between 0 and 100 to trim away similar color values
+      Defaulting to 10 when no tolerance is given.
+ */
+Sharp.prototype.trim = function(tolerance) {
+  if (typeof tolerance === 'undefined') {
+    this.options.trimTolerance = 10;
+  } else if (isInteger(tolerance) && inRange(tolerance, 0, 100)) {
```
kleisauke: TODO: Should I change this to:
```javascript
} else if (isInteger(tolerance) && inRange(tolerance, 1, 99)) {
```

Because it fails on tolerance 100 and it doesn't call the trim function on tolerance 0.

src/pipeline.cc
kleisauke: :+1:

kleisauke:
Squashed, rebased, added logic to `NAN_METHOD(pipeline)` and added some inline notes.

index.js
lovell: Yes please to a 99% maximum. 100% would suggest the whole image should be trimmed, probably not what most people want :)

src/operations.cc
```diff
@@ -399,4 +399,49 @@ namespace sharp {
...
+    VImage profileBottomV;
+    VImage profileBottomH = rows.flipver().profile(&profileBottomV);
+
+    int left = static_cast<int>(floor(profileLeftV.min()));
+    int right = columns.width() - static_cast<int>(floor(profileRightV.min()));
+    int top = static_cast<int>(floor(profileTopH.min()));
+    int bottom = rows.height() - static_cast<int>(floor(profileBottomH.min()));
+
+    // and now crop the original image
+    return image.extract_area(left,  top, right - left, bottom - top);
```
lovell: I guess even with a 99% tolerance we could see rounding errors that lead to a 100%-like failure. How about a non-zero check for the width and height passed to `extract_area`?

kleisauke: Should we throw a `VError` if width or height is less than or equal to 0? Or should we return the 'untouched' input image?

lovell: Good question. Not failing here could produce unexpected results so I think throwing a `VError` is the right thing to do.

lovell:
I "fixed" the Appveyor CI builds via 8dd554b so after a rebase against master this PR should start passing again :)

kleisauke:
All builds are passing, thanks!

lovell:
Dank je!

PR #496, opened 2016-07-06
Bandbool - channel-wise boolean operation

mhirsch:
This pr adds a `bandbool` operation to perform bitwise boolean operations on image channels.  I find this a useful feature for an application I'm working on. I realize not many people will use this, but it's also a native vips function, and the implementation in sharp is straightforward and lightweight.

What do you think about naming the operator in the call to `bandbool` like `bandbool('and')` vs. something like `bandbool({'operation':'and'})`? Since the operator function is self-descriptive I opted for the simpler calling convention.


test/unit/bandbool.js
```diff
@@ -0,0 +1,48 @@
...
+var sharp = require('../../index');
+
+describe('Bandbool per-channel boolean operations', function() {
+
+  it('\'and\' Operation', function(done) {
+    sharp(fixtures.inputPngBooleanNoAlpha)
+      .bandbool('and')
+      .toBuffer(function(err, data, info) {
+        if (err) throw err;
+        assert.strictEqual(200, info.width);
```
lovell: Would an assertion on the number of channels in the output image be useful/valid here?
```javascript
assert.strictEquals(1, info.channels)
```

lovell:
Looks good, thank you Matt.

As you say, probably a limited audience for this feature, but given it's already supported by libvips then it's worth exposing.

I like the simple string-based approach for the API. Perhaps additionally supporting an enum-esque value might also be useful, e.g. `sharp.bool.and` (following the `sharp.kernel.lanczos3` and friends).

lovell:
Thanks for the updates Matt.

Given the channel vs band naming consistency thing, this should probably be called something boring and obvious like `channelBoolean`, but `bandbool` is a much cooler name :)

PR #497, opened 2016-07-07
Add extractChannel command to allow for extracting a color channel from an image

mhirsch:
This pr adds a `band` parameter to the options object passed into `extract` to allow the user to extract a color channel from an image.


lovell:
Overloading region extraction with band separation doesn't quite feel like the right API abstraction to me.

Looking in the [API docs](http://sharp.dimens.io/en/stable/api/), the word "channel" is more frequently used than "band", so how about something like a new `channel()` method that takes a single parameter defining the channel number to include in the output image?

mhirsch:
It's easy enough to scrap this and start over with the `channel` approach. However, just so you understand my initial reasoning, I felt that the verb `extract` clearly indicates to me that I'm leaving the remainder of the image behind. I could see someone, especially someone used to working with imagemagick, thinking that `channel` implies that subsequent operations will apply to a channel, not that the resulting image will contain only that channel. 

I was initially going to add another command like `extract_band`, but since `extract` already existed it seemed neater to give it another type of thing to extract. Anyway, its your project, let me know what you think.

lovell:
Agreed, `channel` on its own would be confusing. How about `extractChannel`?

mhirsch:
This revised pr adds a new command called `extractChannel` which accepts either an integer or one of 'red', 'green', or 'blue' to identify the channel to extract.

lovell:
Looking great, thanks Matt. Please can you add tests for the other named channels (as well as red) then we're good to merge.

test/unit/extractChannel.js
```diff
@@ -0,0 +1,72 @@
...
+
+  it('Red channel', function(done) {
+    sharp(fixtures.inputJpg)
+      .extractChannel('red')
+      .resize(320,240)
+      .toBuffer(function(err, data, info) {
+        if (err) throw err;
+        assert.strictEqual(320, info.width);
+        assert.strictEqual(240, info.height);
+        fixtures.assertSimilar(fixtures.expected('extract-red.jpg'), data, { threshold: 8 }, done);
```
lovell: Is the `{ threshold: 8 }` needed in these tests?

lovell:
Thanks again!

PR #498, opened 2016-07-07
Add a unit test for extract area out of bounds

mhirsch:
This unit test checks to see that an assertion is thrown if `extract` is called for an area outside the image size.

This test doesn't pass right now. I'm not sure if this is even a good test to have -- but I'm submitting a pr in order to understand better. What's the recommended way to throw a proper exception from within native code in node? Obviously it isn't possible to know at the time `execute` is called in the js layer whether the area falls within the image bounds. So this will be caught in `pipeline.cc, Execute()`. There's this `(baton->err).append(...); return Error();` pattern, but that results in the same type of error handling as an error that is thrown from vips, e.g.:

```
Unhandled rejection Error: extract_area: bad extract area

    at Error (native)
```

The test suite doesn't see that as a valid exception, so the unit test fails.


lovell:
The test needs to include `toBuffer( ... )` to invoke the pipeline.

mhirsch:
With this test, there is an `Unhandled rejection Error` thrown --  you can see this on the node command line, but that doesn't register as the assert that the test suite is looking for. Are you aware if there's a way to catch this type of error and throw a proper assert?

lovell:
Once you're in C++ land, any error becomes the first parameter of the invoked callback function, e.g.:
```javascript
it('Bad image area', function(done) {
  sharp(fixtures.inputJpg)
    .extract({ left: 3000, top: 10, width: 10, height: 10 })
    .toBuffer(function(err) {
      assert(err instanceof Error);
      done();
    });
});
```

mhirsch:
Ok! That's what I was looking for. Thanks.

lovell:
Thank you!

PR #500, opened 2016-07-07
Add ability to read and write native vips .v files

mhirsch:
This pr adds the native vips .v file type to the formats that are file read/write-able in sharp. Vips doesn't seem to support reading/writing vips format to/from buffers, so that capability is not included in this pr.


lovell:
Please can you add a test case for saving a vips-format file, otherwise (after a rebase) this looks great, thank you very much.

mhirsch:
Wow, look at that. Unit tests work! Added the test you asked for and caught a bug with it ;)

lovell:
Thanks for the updates Matt.

Given vips format is filesystem-only, we won't need any of the additions to `index.js` as libvips can infer what it needs from the file extension (the `toFormat` logic is for Buffer/Stream-based output).

src/utilities.cc
```diff
@@ -139,7 +139,7 @@ NAN_METHOD(format) {
   // Which load/save operations are available for each compressed format?
   Local<Object> format = New<Object>();
   for (std::string f : {
-    "jpeg", "png", "webp", "tiff", "magick", "openslide", "dz", "ppm", "fits", "gif", "svg", "pdf"
+    "jpeg", "png", "webp", "tiff", "magick", "openslide", "dz", "ppm", "fits", "gif", "svg", "pdf", "vips"
```
lovell: For consistency this should probably be `"v"`.

lovell:
Thank you!

PR #501, opened 2016-07-08
Add a boolean operator that performs bitwise boolean operations between images

mhirsch:
This pr adds a `boolean` operator to sharp that takes a second input image, and performs a selected bitwise boolean operation on a pixel-per-pixel basis between the two images. `boolean` accepts both files and buffers, much like `overlayWith`, and uses the same boolean operators as `bandbool`. It relies on the underlying vips function `boolean`, which handles casting data types and padding the right-hand operand with zeros to match the size of the left-hand operand.


src/common.cc
```diff
@@ -324,4 +324,18 @@ namespace sharp {
     }
   }
 
+  /*
+    Get VIPS Boolean operatoin type from string
+  */
+  VipsOperationBoolean GetBooleanOperation(std::string opStr) {
+    if(opStr == "and" ) {
```
lovell: The rather handy `vips_enum_from_nick` function should allow this to be simplified:
```cpp
return static_cast<VipsOperationBoolean>(
  vips_enum_from_nick(nullptr, VIPS_TYPE_OPERATION_BOOLEAN, opStr.data())
);
```

src/pipeline.cc
```diff
@@ -76,18 +77,23 @@ using sharp::FreeCallback;
 using sharp::CalculateCrop;
 using sharp::counterProcess;
 using sharp::counterQueue;
+using sharp::GetBooleanOperation;
 
 class PipelineWorker : public AsyncWorker {
  public:
   PipelineWorker(Callback *callback, PipelineBaton *baton, Callback *queueListener,
-    const Local<Object> &bufferIn, const Local<Object> &overlayBufferIn) :
+    const Local<Object> &bufferIn, const Local<Object> &overlayBufferIn,
```
lovell: How about we make this a `std::vector<Local<Object>>` of buffers that need to be involved in a call to `SaveToPersistent` and `GetFromPersistent`?

src/pipeline.cc
```diff
@@ -777,6 +783,44 @@ class PipelineWorker : public AsyncWorker {
         }
       }
 
+      // Apply bitwise boolean operation between images
+      if (baton->booleanOp != VIPS_OPERATION_BOOLEAN_LAST &&
+          (baton->booleanBufferInLength > 0 || !baton->booleanFileIn.empty())) {
+        VImage booleanImage;
+        ImageType booleanImageType = ImageType::UNKNOWN;
+        if (baton->booleanBufferInLength > 0) {
```
lovell: I feel like there's potential for extracting this `new_from_buffer`/`new_from_file` logic given it appears in a couple of places now. I'm happy to do this later though; this is more of a note-to-self.

test/unit/boolean.js
```diff
@@ -0,0 +1,68 @@
...
+
+var fs = require('fs');
+var assert = require('assert');
+var fixtures = require('../fixtures');
+var sharp = require('../../index');
+
+describe('Boolean operation between two images', function() {
+
+  it('\'and\' Operation, file', function(done) {
+    sharp('/home/mhirsch/sharp/test/fixtures/2569067123_aca715a2ee_o.jpg') //fixtures.inputJpg
```
lovell: path?

test/unit/boolean.js
```diff
@@ -0,0 +1,68 @@
...
+
+  it('\'or\' Operation, buffer', function(done) {
+    sharp(fixtures.inputJpg)
+      .resize(320,240)
+      .boolean(fs.readFileSync(fixtures.inputJpgBooleanTest), 'or')
+      .toBuffer(function(err, data, info) {
+        if (err) throw err;
+        assert.strictEqual(320, info.width);
+        assert.strictEqual(240, info.height);
+        //fixtures.assertSimilar(fixtures.expected('boolean_or_result.jpg'), data, done);
```
lovell: commented out?

lovell:
I've added a few comments inline, thanks for this Matt.

test/unit/boolean.js
mhirsch: wooooooopsssss

mhirsch:
I took a crack at the "passing as a vector of buffers" improvement you suggested. I stored the node name and buffer size so its easier to iterate over the vector in the constructor. I think, since the constructor runs in the main thread, we could get away without storing the buffer length, but I thought I would avoid the ambiguity.

src/pipeline.cc
```diff
@@ -1219,7 +1271,12 @@ NAN_METHOD(pipeline) {
 
   // Join queue for worker thread
   Callback *callback = new Callback(info[1].As<Function>());
-  AsyncQueueWorker(new PipelineWorker(callback, baton, queueListener, bufferIn, overlayBufferIn));
+
+  std::vector<BufferContainer> saveBuffers;
+  saveBuffers.push_back({"bufferIn", bufferIn, baton->bufferInLength});
```
lovell: If we only append buffers to the vector where their length is >0 then we don't need to store the length.

src/pipeline.cc
```diff
@@ -989,6 +1034,9 @@ class PipelineWorker : public AsyncWorker {
     if (baton->overlayBufferInLength > 0) {
       GetFromPersistent("overlayBufferIn");
     }
+    if (baton->booleanBufferInLength > 0) {
```
lovell: How about a single call to `GetFromPersistent` within a for loop over `saveBuffers`?

src/pipeline.cc
```diff
@@ -76,19 +77,25 @@ using sharp::FreeCallback;
...
+                 const std::vector<BufferContainer> saveBuffers) :
     AsyncWorker(callback), baton(baton), queueListener(queueListener) {
-      if (baton->bufferInLength > 0) {
-        SaveToPersistent("bufferIn", bufferIn);
-      }
-      if (baton->overlayBufferInLength > 0) {
-        SaveToPersistent("overlayBufferIn", overlayBufferIn);
+    for (const BufferContainer b : saveBuffers) {
+      if (b.length > 0) {
+        SaveToPersistent(b.name.c_str(), b.nodeBuf);
```
lovell: We can probably auto-generate a suitable "key" (the first parameter, a string) for `SaveToPersistent` based on the memory location, perhaps something like http://stackoverflow.com/a/7850160/4625124

mhirsch:
@lovell I tried a couple of methods of finding a pointer that is constant between the constructor and `HandleOKCallback`, and didn't find something that worked. In 4c7f339 I made the mistake of not passing the buffer by reference, but fixing this, it appears that the iterator object is actually declared on the stack (from the memory address) and differs between range loops while the buffer element is on the heap. I wasn't aware that that's how c++11 range loops worked -- so I learned something. In e2cda35, I expected going to a standard for loop over the length of the `std::vector`, and getting the address of the vector element to work. However, it seems the referenced object has been copied at some point. I realize there is some ambiguity about which vector is being referenced in the constructor due to the class member `saveBuffers` having the same name as the constructor argument, but I changed the names and the pointers still differed. I'm not sure if this has to do with some obscure fact of a `Local<Object>` object, or copying `std::vector`s but I couldn't find a solution and I'm out of time for tonight. The most recent commit goes back to the `BufferContainer` method of filling the vector.

I think this is going a little off topic from the boolean operator. Can we merge this one, and then work out a better system for passing in buffers to be saved in a later commit?

I would like to be able to reference this mechanism, whatever it is, in a future commit to pass buffers in for other operators that take images.

lovell:
Thank you Matt!

lovell:
I was able to simplify* the SaveTo/GetFromPersistent logic a little in fee3d88 (and e10aeb2 for reasons of d'oh) by using an index-based approach.

_\*assuming people consider a more functional style to be simpler :-/_

Issue #218, opened 2015-05-12
Expose setting ICC profile via withMetadata


keithpaton:
When I resize a CMYK TIFF,  and output to TIFF the output file is RGB not CMYK. 

Do I need to configure the output somehow to enable CMYK output?

Here is the code I am using:

``` javascript
var outputImage = sharp(source)
                .resize(outputSize.w,outputSize.h)
                .sharpen()
                .withMetadata()
                .interpolateWith(sharp.interpolator.bicubic)
                .toFile(outputFile,function(err) {
                    ...
                });
```

lovell:
Hi Keith, the most common use case of sharp is generating images for the web so the current behaviour is to always convert images to device-independent, non-linear sRGB colour space. Using `withMetadata` will also [add the latest web-friendly v2 sRGB ICC profile](https://github.com/lovell/sharp#withmetadata).

It sounds like you're hoping to use it for a different purpose, so this sounds like an excellent enhancement to make.

Would an extension to `withMetadata` to allow it to accept a named profile help, something like `withMetadata(sharp.profile.cmyk)`? If the profile is not provided, it would default to `sharp.profile.srgb` to maintain the existing behaviour. This would mean the only way to get CMYK images out would be to explicitly request them; there would be no support for "pass-through" of the input profile.

keithpaton:
Hi Lovell, 

That approach sounds ideal. Especially since it preserves your existing standard behaviour.

I am working with images for print applications, where the source and destination are CMYK and chose sharp because it can handle TIFF files. I need to handle CMYK JPEG input/output too, probably not as often as TIFF.

As part of the process, I am also generating RGB thumbnails and other images for use in mobile & web applications from the source TIFF and that part works perfectly, so I would love to be able to avoid having to use a different node package (or something more "low level") alongside to create the print/CMYK output.

If you were up for considering adding this enhancement, that would be fantastic. :-)

lovell:
Thanks for confirming Keith. Perhaps a separate `toColourspace`/`toColorspace` method would be more appropriate here, to support things like profile-less CMYK JPEG images. This could then be extended in future to handle other colour spaces such as linear "Adobe RGB".

keithpaton:
That sounds good too. I'm sure there are a few cases, particularly for print use where it's important that the source Colour Space is preserved in the output. I'm conscious that this is a very specific requirement that most people using sharp may not need.

lovell:
Note to implementer/self:

[...]

PR #503, opened 2016-07-13
Add option to greyscale() to enable grayscale output

mhirsch:
This pr adds an options object to the `grayscale()` command to enable output of a single color channel image. The default remains to convert grayscale images to RGB images in sRGB colorspace.


lovell:
Hi Matt, would the proposed `toColourspace` operation in #218 provide a more generic/powerful method of achieving this?

mhirsch:
It would certainly be more powerful to be able to assign arbitrary color profiles. I think something like that ought to be in the API. It was one of the things I was considering working on. However because of the near impossibility, for most people, of creating custom color profiles and the difficulty of finding color profiles that do exactly what is wanted, I think it would be frustrating to create an API where you have to find a color profile for a grayscale color space to get an image with a single channel out. Also, as it is now, you can put an unmanaged image into sharp and get an unmanaged image out -- sometimes you want to know that a red pixel value of 123 in the input maps to a gray pixel value of 123 in the output -- also nearly impossible using managed colorspaces, even with saturation intent.

mhirsch:
This is obviated by #513.

PR #504, opened 2016-07-13
Change the order of operations: boolean, extractChannel, then sRGB conversion

mhirsch:
This small pr moves the `boolean()`, `bandbool()`, and `extractChannel()` operations to before the code that performs sRGB conversion. It doesn't make sense to be able to extract a channel that doesn't appear in the input image. It also doesn't make a lot of sense to perform boolean operations on channels that were implicitly added after the image was loaded. This does raise the somewhat odd potential for colorspace conversion to be performed after a bitwise boolean operation -- perhaps this should be explicitly prevented? It can happen if the input image has a non-sRGB profile.

### DIFF
--- a/src/pipeline.cc
+++ b/src/pipeline.cc
@@ -774,20 +774,6 @@ class PipelineWorker : public AsyncWorker {
         image = Normalize(image);
       }
 
-      // Convert image to sRGB, if not already
-      if (Is16Bit(image.interpretation())) {
-        image = image.cast(VIPS_FORMAT_USHORT);
-      }
-      if (image.interpretation() != VIPS_INTERPRETATION_sRGB) {
-        image = image.colourspace(VIPS_INTERPRETATION_sRGB);
-        // Transform colours from embedded profile to sRGB profile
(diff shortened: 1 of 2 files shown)


lovell:
Thanks Matt. Might this be considered a breaking change? Should this land in v0.16.0 or v0.15.2?

mhirsch:
I can see a circumstance where it might break existing code. The intent is to avoid confusing situations where a boolean operation occurs on a channel that you weren't expecting to exist, or you are able to extract a channel that didn't exist in the input image. So to me the question is whether it is more likely that people run into problems with the code as is, or with the change. Given that we just added these features three or four days ago, it seems unlikely that someone would have already written code to exploit one of the existing corner cases. The overlap between using bitwise boolean operations and working with images that contain a color profile seems like it would be small.

I guess I can see both sides here. Perhaps there's some additional behavior that could be enforced to make this an easier decision? In the end though it doesn't matter much to me which version the change lands in -- whatever you think is best.

lovell:
Thanks Matt, I agree this has minimal impact - will be in the next point release.

Issue #509, opened 2016-07-14
0.15.1: Additional error from libvips when attempting to extract an invalid area


papandreou:
```
$ node -e "require('sharp')('testdata/turtle.jpg').extract({left: 10000, top: 10000, width: 10, height: 10}).toBuffer(console.log);"
Error: enum 'VipsOperationBoolean' has no member 'null', should be one of: and, or, eor, lshift, rshift
enum 'VipsOperationBoolean' has no member 'null', should be one of: and, or, eor, lshift, rshift
extract_area: bad extract area

    at Error (native) null null
```

(turtle.jpg is a 481x424 image)

With sharp 0.15.0 only the last part ("extract_area: bad extract area") was logged. Is that reason for concern?

papandreou:
Also happens on master (15a5778).

mhirsch:
I submitted a pr for this. I'm not sure what happened -- these assignments used to be surrounded by a check for whether the operation in question was requested. The pr takes the approach of checking whether the operations exist in the options object rather than special casing the 'null' operation. Does that sound ok to you?

lovell:
@papandreou Are you able to confirm @mhirsch's fix (and associated test case) fixes this?

papandreou:
@lovell Confirmed -- doesn't happen on master now.

mhirsch:
I think this could be closed now, unless anyone objects.

papandreou:
@mhirsch I didn't close it because @lovell usually likes to keep issues open until the a fix has been released.

lovell:
v0.16.0 is now available via npm.

PR #511, opened 2016-07-15
Prevent boolean errors during extract operation (#509)

mhirsch:
This pr addresses the error message seen in #509


lovell:
Marvellous, thank you Matt.

PR #516, opened 2016-07-19
boolean() and overlayWith() accept raw buffers as input

mhirsch:
This PR allows `boolean()` and `overlayWith()` to accept raw buffers, just as `sharp()` does. It uses the same `options.raw` syntax as `sharp()`.


index.js
```diff
@@ -382,13 +389,32 @@ Sharp.prototype.boolean = function(operand, operator) {
...
+  }
+
   return this;
 };
 
 /*
   Overlay with another image, using an optional gravity
 */
 Sharp.prototype.overlayWith = function(overlay, options) {
+  /* jshint maxcomplexity:16 */
```
lovell: We'll need much higher test coverage at this complexity level. However a better approach here is probably to abstract the common logic for the various file/Buffer/raw input to a separate function and keep the `boolean` and `overlayWith` functions themselves simpler.

lovell:
Commit e6bfa52 on the `pencil` (v0.16.0) branch introduces quite a significant refactor to abstract the common image opening logic, which just so happens to allow the use of raw pixel data with *any* method that takes an input image.

I took the liberty of using the test cases from this PR to prove this is the case :)

I'll close this PR as it is no longer necessary but, most importantly, thank you for kicking me into action to clean up the code and make it more extensible for others to contribute.

PR #550, opened 2016-08-23
add withIcc command to override input file ICC profile

mhirsch:
This PR adds a `withIcc()` command which overrides the input ICC profile.

I think the unit test is not quite sufficient, as the difference between the colorspace transformed image and the original input image are very subtle (the RGB values differ by 1 or 2 in most cases) with the ICC profile I'm using. `fixtures.assertSimilar()` will miss the difference if no transform occurred. The `Bluish.icc` file was taken from the lcms project via colord. I wonder if there's an ICC profile that would create a more dramatic difference? I hesitate to do something like an `assert.strictEqual()` on a couple of pixels of the image, as different systems might have different rounding errors during the colorspace conversion.


lovell:
Hello, there's been some previous discussion in #218 about adding the ability to inject an ICC profile via the existing `withMetadata` operation. I'll attempt to update it with some thoughts on what the API might look like.

lovell:
Hello, thank you for your time on this so far. I'm going to close this PR as it has drifted a long way behind the master branch - let's use #218 to track improvements.

PR #408, opened 2016-04-12
Support for saving lossless WebP

dhotson:
Hello. :)

This PR adds initial support for saving lossless WebP.

Sharp usually saves WebP in lossy mode with a default quality of 80.
A bug that was fixed in libvips 8.3 means lossless WebP output is now possible.

To enable this in sharp, I've introduced an options object parameter to the webp method. e.g.

```javascript
var image = sharp('file.png')
  .webp({ lossless: true });
```

I'd be happy to make changes to the API—please just let me know.

Also, I haven't added test cases since my test environment isn't working quite right. Once I've got that worked out—any thoughts on the kinds of tests you would like to see?

### Todo

- [ ] extra options parameter for `toFormat()` and other methods?
- [ ] unit tests
- [ ] documentation?

#### PNG (lossless)
image

---

#### WebP (lossy - quality:80)
image

---

#### WebP (lossless)
image

PR #569, opened 2016-09-12
WebP advanced output options

smremde:
Please see this pull request for three additional WebP output option: lossless, alpha_q and near_lossless. See http://www.vips.ecs.soton.ac.uk/supported/8.3/doc/html/libvips/VipsForeignSave.html#vips-webpsave-buffer

I am having trouble getting near_lossless and alpha_q working

```
vips warning: webpsave_buffer: no property named `near_lossless'
vips warning: webpsave_buffer: no property named `alpha_q'
```

The version of libvips downloaded is 8.3.3 so I'm not sure why these are not working - any advice?

I have added documentation and a simple unit test for alpha_q, which should pass when this property can be set. I'll add unit tests for the other parameters after.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -626,6 +626,23 @@ The output quality to use for lossy JPEG, WebP and TIFF output formats. The defa
 
 `quality` is a Number between 1 and 100.
 
+#### alphaQuality(alphaQuality)
+
+The output quality to use for the alpha channel of WebP output formats. The default quality is `100`.
+
+`alphaQuality` is a Number between 1 and 100.
+
+#### lossless(lossless)
+
+Set the use off lossless output of WebP output formats. The default quality is `false`.
+
+`lossless` is ais a boolean where `true` enables and `false` disables lossless output.
+
+#### nearLossless(nearLossless)
+
+Enable preprocessing on lossy compression of WebP formats to increase compression by up to 50%.
+
+
 #### progressive()
 
 Use progressive (interlace) scan for JPEG and PNG output. This typically reduces compression performance by 30% but results in an image that can be rendered sooner when decompressed.
--- a/docs/api.md
+++ b/docs/api.md
@@ -642,6 +642,7 @@ Set the use off lossless output of WebP output formats. The default quality is `
 
 Enable preprocessing on lossy compression of WebP formats to increase compression by up to 50%.
 
+`nearLossless` is a boolean where `true` enables and `false` disables near lossless output.
 
 #### progressive()
 
--- a/index.js
+++ b/index.js
@@ -98,6 +98,9 @@ var Sharp = function(input, options) {
     fileOut: '',
     progressive: false,
     quality: 80,
+    alphaQuality: 100,
+    lossless: false,
+    nearLossless: false,
     compressionLevel: 6,
     withoutAdaptiveFiltering: false,
     withoutChromaSubsampling: false,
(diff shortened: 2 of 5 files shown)


lovell:
Hi Stephen, thank you for this PR.

The new WebP options will be in the forthcoming libvips v8.4.0, hence the errors. I expect the future sharp v0.17.0 to have a dependency on libvips v8.4.x.

In terms of the API, rather than the existing method-per-option approach, I'd like to start moving any format-specific options into a method specific to that format (e.g. `webp({nearLossless: true})`). Given the output format may not be known until the input has been read, this might also require a single method that accepts an Object of options for many formats.

I realise this is a much wider change than just the new WebP options here, but this PR feels like a good place to start.

Let me have a go at moving the JPEG-specific `withoutChromaSubsampling` first (with a deprecation route) to explore how the API could behave.

/cc @dhotson who may be interested in these options also - see #408

dhotson:
@lovell @smremde Ah this is brilliant! It's great to see support for these options. 😄

I've been keeping a branch with just basic lossless support: https://github.com/lovell/sharp/compare/master...dhotson:webp-save-lossless

.. the project I was working on got delayed so I never got around to finishing it off with proper tests and docs etc sorry. Thank you very much @smremde for taking the initiative on this PR. 😄 

For what it's worth—I think I agree that it makes sense to have format specific options as params, rather than separate methods. 👍

smremde:
I am confused about alpha_q and near_lossless. The libvips documentation suggests this is in 8.3 as per the link I sent. Checking git, it appears that the documentation is wrong on this one. 

Yes I agree, per format options would be much better. I'll watch this PR and can update it once you have explored the best way of doing this.

lovell:
@jcupitt Could the libvips 8.3 published docs have been generated from 8.4 source?

jcupitt:
Oh dear, I'm very sorry, these are 8.4 options.

The 8.3 docs had a number of serious issues (whole sections missing, broken cross-referencing, argh) and I spent ages fixing up doc generation for 8.4. The poor 8.3 docs were causing people quite a few problems, so I put the much better 8.4 docs up instead. I hoped that the win from having working docs outweighed the confusion of having a few features which didn't actually exist.

8.4 should be out within the next week, so this problem will vanish. I hope I didn't waste too much of your time.

smremde:
@jcupitt Not at all. This PR should work with little or no modification once sharp upgrades to 8.4 I assume.

lovell:
> "Let me have a go at moving the JPEG-specific withoutChromaSubsampling first (with a deprecation route) to explore how the API could behave."

See commit 1159c99 for the first part of the groundwork for this.

lovell:
> "In terms of the API, rather than the existing method-per-option approach, I'd like to start moving any format-specific options into a method specific to that format (e.g. webp({nearLossless: true})). "

This approach, with complete JSDoc comments, is now available in the `master` branch when setting output options.

https://github.com/lovell/sharp/blob/master/lib/output.js

Functions relevant to this PR are `webp()`, the `_setBooleanOption()` helper and the `is.*` type-checkers.

rnanwani:
Is this feature merged through some other PR? If not, then can I take this up to fit into the new API?

lovell:
@rnanwani This is not merged yet so alternative PR would be welcome, thank you.

rnanwani:
Can you confirm if this is the right command to test vips near lossless flag
```
vips copy test/fixtures/alpha-premultiply-1024x768-paper.png /tmp/webp-near-lossless-20.webp[near_lossless,Q=30]
```

lovell:
@rnanwani Yes, `near_lossless` is the relevant flag - see also https://github.com/jcupitt/libvips/pull/430

lovell:
Superseded by #68. Thank you for the original inspiration and work on this everyone.

PR #581, opened 2016-09-22
Allow for overriding platform, arch and arm_version

gigafied:
We are using sharp on an embedded platform, and we need to cross compile on Jenkins when we create a debian package. Right now, binding.js reads the host's arch, arm_version and platform, this enables us to specify them as command line args to `npm install` i.e. `npm install --arch=arm --armv=7 --platform=linux`

### DIFF
--- a/binding.js
+++ b/binding.js
@@ -17,6 +17,16 @@ var minimumLibvipsVersion = process.env.npm_package_config_libvips || require('.
 
 var vipsHeaderPath = path.join(__dirname, 'include', 'vips', 'vips.h');
 
+var platform = process.env.npm_config_platform || process.platform;
+
+var arch = process.env.npm_config_arch || process.arch;
+var arm_version = process.env.npm_config_armv || process.config.variables.arm_version;
+
+if (arch === 'arch64' || arch === 'armhf') {
+  arch = 'arm';
+  if (arch === 'arch64') arm_version = '8';
+}
+
 // -- Helpers
 
 // Does this file exist?
@@ -47,9 +57,9 @@ var unpack = function(tarPath, done) {
 };
 
 var platformId = function() {
-  var id = [process.platform, process.arch].join('-');
-  if (process.arch === 'arm') {
-    switch(process.config.variables.arm_version) {
+  var id = [platform, arch].join('-');
+  if (arch === 'arm') {
+    switch(arm_version) {
       case '8':
         id = id + 'v8';
         break;
@@ -79,7 +89,7 @@ module.exports.download_vips = function() {
   // Has vips been installed locally?
   if (!isFile(vipsHeaderPath)) {
     // Ensure Intel 64-bit or ARM
-    if (process.arch === 'ia32') {
+    if (arch === 'ia32') {
       error('Intel Architecture 32-bit systems require manual installation - please see http://sharp.dimens.io/en/stable/install/');
     }
     // Ensure glibc >= 2.15


lovell:
Thanks for this, very useful. We should probably document this too - I'm happy to do that after.

gigafied:
@lovell Still one issue, but not sure how to overcome. The OS conditionals in binding.gyp still check host platform, i.e. https://github.com/lovell/sharp/blob/master/binding.gyp#L121

Not sure if there's a way to pass `--platform` to node-gyp, any ideas? 

For now, not a big issue, since both our build machine and target platform are linux, however it would be nice if I could cross-compile on my laptop (macOS) as well.

binding.js
```diff
@@ -17,6 +17,13 @@ var minimumLibvipsVersion = process.env.npm_package_config_libvips || require('.
 
 var vipsHeaderPath = path.join(__dirname, 'include', 'vips', 'vips.h');
 
+var platform = process.env.npm_config_platform || process.platform;
+
+var arch = process.env.npm_config_arch || process.arch;
+if (arch === 'armhf') arch = 'arm';
```
gigafied: I'll add that later today, can also probably set version to 8 if value is `aarch64`, I guess there is `armel` but think that's considered obsolete now.

lovell:
> "The OS conditionals in binding.gyp still check host platform"

The [GYP_DEFINES](https://www.chromium.org/developers/gyp-environment-variables) env var might solve this.

gigafied:
@lovell thanks! That seems to work (getting other compile errors, but looks like it is respecting `GYP_DEFINES`.

Really wish all of this was cohesive and you didn't have to specify these values in multiple places, but I guess nothing to be done about it at the module's level.

gigafied:
@lovell added support for  `aarch64`, also defaulted to `7` for `arm_version` if not specified. Once you sign off, I'll rebase.

binding.js
```diff
@@ -20,9 +20,14 @@ var vipsHeaderPath = path.join(__dirname, 'include', 'vips', 'vips.h');
 var platform = process.env.npm_config_platform || process.platform;
 
 var arch = process.env.npm_config_arch || process.arch;
-if (arch === 'armhf') arch = 'arm';
-
-var arm_version = process.env.npm_config_armv || process.config.variables.arm_version;
+var arm_version = process.env.npm_config_armv || process.config.variables.arm_version || '7';
```
lovell: The default value for `arm_version` is currently 6 - see https://github.com/lovell/sharp/blob/master/binding.js#L59

binding.js
```diff
@@ -20,9 +20,14 @@ var vipsHeaderPath = path.join(__dirname, 'include', 'vips', 'vips.h');
...
+
+switch (arch) {
+  case 'aarch64':
+    arm_version = '8';
+  case 'armhf':
```
lovell: Missing `break;`

lovell:
> "Really wish all of this was cohesive and you didn't have to specify these values in multiple places, but I guess nothing to be done about it at the module's level."

I agree. Are you able to provide a worked example, perhaps as a comment here, on which I can base the docs?

binding.js
gigafied: The missing `break;` is intentional so that the arch gets set to `arm` for both `aarch64` and `armhf` cases.

gigafied:
```
$ export GYP_DEFINES=OS=linx target_arch=arm arm_version=7
$ npm install --arch=arm --armv=7 --platform=linux
```

the `npm install` line definitely works, `GYP_DEFINES` seems to work fine, but I haven't gotten mac os cross compiling working yet.

binding.js
lovell: Sorry, yes, that makes sense. Please can you add an explicit `/* falls through */` comment to keep the linter happy.

gigafied: Just replaced with to `ifs` I hate linter exceptions. Should be good now.

gigafied:
Rebased so should be ready for merge.

lovell:
Thank you!

PR #596, opened 2016-10-08
enhancement/#472-generate-typescript-typings

### LINKED ISSUE #472: Generate TypeScript typings, preferably automagically
first of all thx for your amazing library. I am writing an application using sharp. I would like to use [TypeScript](https://github.com/Microsoft/TypeScript) but unfortunately your library doesn't come with typings included.

I already started implementing some of them and would like to know if its worth the work to complete them so you can include them into your package. You could also think about generating the docs from the type definitions.

```
declare module "sharp" {
  function s(input: string | Buffer, options?: any): s.ISharpImage;

  namespace s {
    interface ISharpImage {
      /**
       * Fast access to image metadata without decoding any compressed image data.
       */
      metadata(cb?: (error?: Error, metadata?: IMetadata) => void): void;
      /**
       * Takes a "snapshot" of the instance, returning a new instance. Cloned instances inherit the input of their parent instance.
       * 
       * This allows multiple output Streams and therefore multiple processing pipelines to share a single input Stream.
       */
      clone(): ISharpImage;
      resize(width?: number, height?: number, options?: any): ISharpImage;
      max(width?: number, height?: number): ISharpImage;
      min(width?: number, height?: number): ISharpImage;
      rotate(angle?: number): ISharpImage;
      toFile(path: string, callback?: (err?: Error, info?: any) => void): any;
    }

    /**
     * image metadata
     */
    interface IMetadata {
      /**
       * Name of decoder used to decompress image data e.g. `jpeg`, `png`, `webp`, `gif`, `svg`
       */
      format: 'jpeg' | 'png' | 'webp' | 'gif' | 'svg';
      /**
       * Number of pixels wide
       */
      width: number;
      /**
       * Number of pixels high
       */
      height: number;
      /**
       * Name of colour space interpretation e.g. srgb, rgb, scrgb, cmyk, lab, xyz, b-w ...
       */
      space: string;
      /**
       * Number of bands e.g. `3` for sRGB, `4` for CMYK
       */
      channels: number;
      /**
       * Boolean indicating the presence of an embedded ICC profile
       */
      hasProfile: boolean;
      /**
       * Boolean indicating the presence of an alpha transparency channel
       */
      hasAlpha: boolean;
      /**
       * Number value of the EXIF Orientation header, if present
       */
      orientation: number;
      /**
       * Buffer containing raw EXIF data, if present
       */
      exif: Buffer;
      /**
       * Buffer containing raw ICC profile data, if present
       */
      icc: Buffer;
    }
  }

  export = s;
}
```

BrianDGLS:
closes #472


lovell:
Thanks for this. A couple of questions:

1. Is a `types` entry in `package.json` required to use this?
2. Are you aware of a suitable module/tool we can use to add a unit test that verifies/lints the typings file (like we have for [jshint](https://github.com/lovell/sharp/blob/master/test/unit/jshint.js) and [cpplint](https://github.com/lovell/sharp/blob/master/test/unit/cpplint.js))?

arusakov:
Too many `any` in definition :(

lovell:
Having evaluated `dts-gen` further, it doesn't feel like the right tool for the job.

> "dts-gen is meant to be a starting point for writing a high-quality definition file"

What we need is a tool to consistently and repeatedly generate the "high-quality" definition file from JSDoc (or very similar) in an automated manner.

Thanks for your submission though - it's definitely helped me assess the (lack of) tooling currently available.

PR #617, opened 2016-11-01
Test the file path with Chinese characters.

natural-law:
@lovell 
I encountered a problem on Windows:
Once a image file path contains Chinese characters, sharp will can't handle the image on Windows. But it works well on Mac.

This PR is a test case for reproduce the problem. Could you please solve the problem? Thanks very much!

### DIFF
--- a/.jshintrc
+++ b/.jshintrc
@@ -6,6 +6,8 @@
   "globals": {
     "beforeEach": true,
     "afterEach": true,
+    "before": true,
+    "after": true,
     "describe": true,
     "it": true
   }
--- a/test/unit/chinesePath.js
+++ b/test/unit/chinesePath.js
@@ -1,0 +1,77 @@
+'use strict';
+
+var assert = require('assert');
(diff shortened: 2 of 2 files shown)


lovell:
Hello, libvips v8.4.0+ is required for Windows unicode support - see https://github.com/jcupitt/libvips/blob/master/ChangeLog#L57

sharp v0.17.0 will provide pre-compiled binaries for libvips v8.4.2 (or later).

natural-law:
@lovell Thanks for your reply.

> sharp v0.17.0 will provide pre-compiled binaries for libvips v8.4.2 (or later).

Could you please tell me when the sharp v0.17.0 will be released? Or whether there is a way/document can guide me to update the libvips myself? Thank you very much!

lovell:
https://github.com/lovell/sharp/milestone/25

The work-in-progress branch is unstable but can be used for early testing:

```sh
npm install lovell/sharp#quill
```

Issue #337, opened 2016-01-12
Deep Zoom / .tile() quality setting not used?


homerjam:
Hi Lovell

I think this might be a vips issue but the quality setting doesn't seem to effect the output when using the `.tile()` method.

I've looked (bear in mind I don't have any C experience) and can't find any references to quality in the sharp/vips code relating to dz images so I'm assuming its using a default setting?

Thanks

James

lovell:
Hi James, although there's currently no support for setting the image quality of tiles in sharp, libvips does expose a [suffix](https://github.com/jcupitt/libvips/blob/master/libvips/foreign/dzsave.c#L2065-L2066) option that could be used to control this.

I suspect other existing options relating to image formats and options could be passed through using this method, e.g. for line art drawings `.png[compression=9,filter=none]` might be more suitable.

Let's file this under enhancement - probably one to tackle at the same time as #223 - thanks for the suggestion!

lovell:
PNG output is definitely required as part of this task - see #386

lovell:
Added in #622 and now available in v0.17.0.

PR #622, opened 2016-11-11
propagate format to tile output

ppaskaris:
This PR adds format propagation to the tile output, fixing https://github.com/lovell/sharp/issues/337. Specifically, it forwards the output format and output format options through the suffix argument to vips. Format propagation only occurs if explicitly set (i.e. not `input`), otherwise it defaults to the current behavior of `jpeg`.

I've only added support for `jpeg`, `png` and `webp` since those seem like the most desirable tile formats. Also, when using `jpeg` with Google or Zoomify layouts, the `.jpg` extension is used instead of `.jpeg` as per [libvips defaults](https://github.com/jcupitt/libvips/blob/6bff578fd978bace5959e894341809628928c0b5/libvips/foreign/dzsave.c#L1677) and to maintain compatibility.

Usage looks like this:

```js
sharp('huge-map.svg')
  .background('transparent')
  .resize(8162, 8162)
  .webp({ quality: 1 }) // new
  .tile({ layout: 'google' })
  .toFile('./tiles', function (err, info) {
    if (err) throw err;
    // ./tiles/0/0/0.webp contains a very ugly webp encoded tile
  });
```

PR is against master because "ridge" branch does not appear to exist.

### DIFF
--- a/lib/output.js
+++ b/lib/output.js
@@ -285,7 +285,15 @@ const tile = function tile (tile) {
       }
     }
   }
-  return this;
+  // Format
+  if (is.defined(this.options.formatOut) && is.string(this.options.formatOut)) {
+    if (is.inArray(this.options.formatOut, ['jpeg', 'png', 'webp'])) {
+      this.options.tileFormat = this.options.formatOut;
+    } else if (this.options.formatOut !== 'input') {
+      throw new Error('Invalid tile format ' + this.options.formatOut);
+    }
+  }
+  return this._updateFormatOut('dz');
 };
 
 /**
--- a/lib/output.js
+++ b/lib/output.js
@@ -286,12 +286,10 @@ const tile = function tile (tile) {
     }
   }
   // Format
-  if (is.defined(this.options.formatOut) && is.string(this.options.formatOut)) {
-    if (is.inArray(this.options.formatOut, ['jpeg', 'png', 'webp'])) {
-      this.options.tileFormat = this.options.formatOut;
-    } else if (this.options.formatOut !== 'input') {
-      throw new Error('Invalid tile format ' + this.options.formatOut);
-    }
+  if (is.inArray(this.options.formatOut, ['jpeg', 'png', 'webp'])) {
+    this.options.tileFormat = this.options.formatOut;
+  } else if (this.options.formatOut !== 'input') {
+    throw new Error('Invalid tile format ' + this.options.formatOut);
   }
   return this._updateFormatOut('dz');
 };
--- a/src/pipeline.cc
+++ b/src/pipeline.cc
@@ -849,37 +849,31 @@ class PipelineWorker : public Nan::AsyncWorker {
           // Forward format options through suffix
           std::string suffix;
           if (baton->tileFormat == "png") {
-            suffix = ".png[interlace="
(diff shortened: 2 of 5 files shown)


lovell:
Hello, thanks for this Patrick, looks good. I'll add a couple of comments inline. Are you able to add more test cases, for example there's nothing to verify WebP-based tile output works.

@homerjam As the originator of #337 are you able to help test this also?

src/pipeline.cc
```diff
@@ -846,13 +846,49 @@ class PipelineWorker : public Nan::AsyncWorker {
           if (isDzZip) {
             baton->tileContainer = VIPS_FOREIGN_DZ_CONTAINER_ZIP;
           }
+          // Forward format options through suffix
+          std::string suffix;
+          if (baton->tileFormat == "png") {
+            suffix = ".png[interlace="
```
lovell: Might it be worth externalising this string concatenation? I'm thinking of something like a function that accepts a `vector<pair<string, string>>` and returns a single `string`.

test/unit/tile.js
```diff
@@ -128,6 +128,22 @@ describe('Tile', function () {
...
+  it('Valid formats pass', function () {
+    ['jpeg', 'png', 'webp'].forEach(function (format) {
+      assert.doesNotThrow(function () {
+        sharp().toFormat(format).tile();
+      });
+    });
+  });
+
+  it('Invalid formats fail', function () {
+    ['zoinks', 1, 'tiff', 'raw'].forEach(function (format) {
```
lovell: I think the `zoinks` and `1` values here end up testing the existing `toFormat` function rather than your additions to the `tile` function.

test/unit/tile.js
```diff
@@ -224,6 +240,31 @@ describe('Tile', function () {
...
+          layout: 'google'
+        })
+        .toFile(directory, function (err, info) {
+          if (err) throw err;
+          assert.strictEqual('dz', info.format);
+          assert.strictEqual(2725, info.width);
+          assert.strictEqual(2225, info.height);
+          assert.strictEqual(3, info.channels);
+          assert.strictEqual('number', typeof info.size);
+          fs.stat(path.join(directory, '0', '0', '0.png'), function (err, stat) {
```
lovell: Perhaps use `sharp(path.join(directory, '0', '0', '0.png')).metadata()...` to verify the tile really is a PNG image?

lib/output.js
```diff
@@ -285,7 +285,15 @@ const tile = function tile (tile) {
       }
     }
   }
-  return this;
+  // Format
+  if (is.defined(this.options.formatOut) && is.string(this.options.formatOut)) {
```
lovell: `this.options.formatOut` should always have a string value and therefore will be defined.

src/pipeline.cc
ppaskaris: Done! It's much more readable now.

test/unit/tile.js
ppaskaris: I've also added an extreme quality setting along with a file size check (with a liberal margin of error) to verify that the format options are indeed making it through.

lovell:
Fantastic, thanks for all the updates, this will be in v0.17.0

Issue #411, opened 2016-04-14
"Error: linear: vector must have 1 or 2 elements" raised when resizing a b&w png image


janaz:
Hi Lovell,

Sharp is failing to resize a png file with the following metadata:
```
{ format: 'png',
  width: 1653,
  height: 2339,
  space: 'b-w',
  channels: 1,
  density: 200,
  hasProfile: false,
  hasAlpha: false }
```

To reproduce, please run the following:
```javascript
var sharp = require('sharp');
var i = sharp('image.png');

i.embed() // works without embed
  .resize(100, 100)
  .background({r:0,g:0,b:0,a:0}) // works with a:1
  .toBuffer(function(err, output, info) {
    if (err) {
      throw err;
    } else {
      console.log("All good")
    }
  });
```

The error thrown is `Error: linear: vector must have 1 or 2 elements`

It works when 'embed()' is not used or the alpha channel in the background color is set to `1`

I created a repo that showcases the issue and includes the b&w png file: https://github.com/janaz/sharp-png-issue

lovell:
That looks like a bug - the background colour embedding logic has probably never been tested with 1 channel images - I'll take a look.

lovell:
Commit e0d5826 adds support for embed with 1 and 2 channel images. Thanks for reporting this!

lovell:
v0.14.1 now available via npm. Please re-open this if it doesn't solve the problem Tomasz.

PR #623, opened 2016-11-14
fix extend for 2 channel sources

ppaskaris:
This PR fixes `extend` when using a 2 channel source (e.g. greyscale png w/ alpha). I've simply copied the fix for https://github.com/lovell/sharp/issues/411 from https://github.com/lovell/sharp/commit/e0d58266be05da72c1072c5c7048637113b172c0 and added a test to verify it.

Previously this resulted in the following error from libvips: `linear: vector must have 1 or 2 elements`.


lovell:
Fantastic, thank you!

PR #644, opened 2016-12-02
Better Error Messages for Strings in Resize

spikeon:
Added Error Message "Invalid width [width].  Cannot be string."
Added Error Message "Invalid height [height].  Cannot be string."

This will make it easier for developers to identify when they have mis-cast their variables before sending them into sharp.resize()

### DIFF
--- a/lib/resize.js
+++ b/lib/resize.js
@@ -103,6 +103,8 @@ const resize = function resize (width, height, options) {
   if (is.defined(width)) {
     if (is.integer(width) && is.inRange(width, 1, this.constructor.maximum.width)) {
       this.options.width = width;
+    } else if (is.string(width)) {
+      throw new Error('Invalid width ' + width + '. Cannot be string.');
     } else {
       throw new Error('Invalid width (1 to ' + this.constructor.maximum.width + ') ' + width);
     }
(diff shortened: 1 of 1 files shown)


lovell:
Thank you! How about we consolidate this improvement with the existing Error and make it more positive, taking more of a "here's what you should do" approach rather than "here's one possible thing that won't work".

I'm thinking something along the lines of:
```diff
- Invalid height (1 to ' + this.constructor.maximum.height + ') ' + height
+ Expected integral Number for height between 1 and ' + this.constructor.maximum.height + ', received ' + height + ' (' + typeof height +')'
```

spikeon:
Placing the term "Integral number" doesn't tell the user that they have sent the wrong data-type.  Perhaps instead of "Cannot be a string" and validating to see if it's a string, it should say: "Must be an Integral Number" if any non numeric datatype is sent into the function?

So, I'm all for the enhancement to the existing error, but there should also be an error specific to receiving an incorrect data type.

lovell:
> Perhaps ... it should say: "Must be an Integral Number" if any non numeric datatype is sent into the function?

Agreed, this sounds like what I was thinking too.

How about we externalise the responsibility for creating the Error and its message to a separate, generic function, something like:

```javascript
const invalidParameterError = function (name, expectedDescription, actual) {
  return new Error(
    `Expected ${expectedDescription} for ${name} but received ${actual} of type ${typeof actual}`
  );
};
```
which could then be used by the `resize()` function via something like:
```javascript
throw invalidParameterError('height', `integral Number between 1 and ${maximum.height}`, height);
```

spikeon:
That... looks just about perfect.  It tells the developer specifically what's wrong, as well as exactly what's expected.

lovell:
I've added this as part of commit 70a3d4f - thanks for your work on this.

Issue #655, opened 2016-12-17
debian multiarch libvips-cpp.so path


kapouer:
Hi,

on debian/stretch, the path is `/usr/lib/x86_64-linux-gnu/libvips-cpp.so`,
thus the check in binding.gyp fails.
Otherwise it compiles all right with system-installed vips 8.4.4.
That script is obscure to me, so i can't propose a PR.

kapouer:
Haha
> pkg-config --variable libdir vips-cpp
> /usr/lib/x86_64-linux-gnu

kapouer:
Could you release that ? I might find a minute to upload it in debian.

lovell:
I should get some time for this after the new year. 🎆

lovell:
@kapouer v0.17.1 est disponible maintenant.

kapouer:
Merci

PR #656, opened 2016-12-17
Simpler expression for finding vips-cpp libdir

kapouer:
Fixes: #655

However, since i did not understand all the line i simplified, i may have removed something.


lovell:
Merci, c'est beaucoup moins horrible! :)

PR #663, opened 2016-12-31
Remove parentheses from sharp.format example

bhank:
With parentheses it gets "sharp.format is not a function"


lovell:
Good catch, thank you.

Issue #673, opened 2017-01-09
Docs: extend background example needs to use alpha attribute rather than a


zilongshanren:
tested sharp version: 0.17

On Windows platform the `background` option {r:0, g:0, b:0, a:0} is OK, alpha is properly handled.

But on macos, I must use {r:0, g:0, b:0, alpha:0}.

If I stick to {r:0, g:0, b:0, a:0} , I will get this error message:

![image](https://cloud.githubusercontent.com/assets/785541/21761653/520eb42e-d68f-11e6-8e94-e1d1091e4e24.png)

I found the docs here http://sharp.dimens.io/en/stable/api-operation/#extend is also using the **a** for alpha.

I think this might be a bug

zilongshanren:
This is my bad

lovell:
Hello, it looks like the example in the [extend jsdoc](https://github.com/lovell/sharp/blob/master/lib/operation.js#L189) is incorrect. As you correctly point out, `alpha` should be used here. I'm happy to accept a PR to correct this or can fix it myself when I get a moment.

zilongshanren:
@lovell  thanks, 😄  I could make one

zilongshanren:
@lovell  done.

PR #675, opened 2017-01-09
rename a to alpha in background color option

zilongshanren:
fixed issue: https://github.com/lovell/sharp/issues/673


lovell:
Thank you / 谢谢

PR #685, opened 2017-01-17
advanced webp output options - alpha quality, lossless and near lossless

rnanwani:


lib/constructor.js
```diff
@@ -145,6 +145,8 @@ const Sharp = function (input, options) {
     pngCompressionLevel: 6,
     pngAdaptiveFiltering: true,
     webpQuality: 80,
+    webpAlphaQuality: 100,
+    webpLossless: false,
     tiffQuality: 80,
```
lovell: Is this missing a `webpNearLossless: false` default? This might be the reason for the "Partial image extraction WebP" test failure.

lib/output.js
```diff
@@ -168,6 +168,9 @@ const png = function png (options) {
  * Use these WebP options for output image.
  * @param {Object} [options] - output options
  * @param {Number} [options.quality=80] - quality, integer 1-100
+ * @param {Number} [options.alphaQuality=100] - quality of alpha layer, integer 0-100
+ * @param {Boolean} [options.lossless=false] - use lossless compression mode
+ * @param {Boolean} [options.nearLossless=false] - use near_lossless compression mode
```
lovell: Are `lossless` and `nearLossless` mutually exclusive?

src/pipeline.cc
```diff
@@ -1209,6 +1218,11 @@ NAN_METHOD(pipeline) {
   baton->pngCompressionLevel = AttrTo<uint32_t>(options, "pngCompressionLevel");
   baton->pngAdaptiveFiltering = AttrTo<bool>(options, "pngAdaptiveFiltering");
   baton->webpQuality = AttrTo<uint32_t>(options, "webpQuality");
+  baton->webpAlphaQuality = AttrTo<uint32_t>(options, "webpAlphaQuality");
+  baton->webpLossless = AttrTo<bool>(options, "webpLossless");
+  if(HasAttr(options, "webpNearLossless")) {
```
lovell: Does this attribute need the `HasAttr` check given there is (or should be) a default?

test/unit/io.js
```diff
@@ -372,6 +372,50 @@ describe('Input/output', function () {
...
     });
+
+    it('should work for webp alpha quality', function (done) {
+      sharp(fixtures.inputPngAlphaPremultiplicationSmall)
+        .webp({alphaQuality: 80})
+        .toBuffer(function (err, data, info) {
+          if (err) throw err;
+          assert.strictEqual(true, data.length > 0);
+          assert.strictEqual('webp', info.format);
+          fixtures.assertSimilar(fixtures.expected('webp-alpha-80.webp'), data, function (err) {
```
lovell: The inline function can be rolled up to use `done`, something like:
```javascript
fixtures.assertSimilar(fixtures.expected('webp-alpha-80.webp'), data, done);
```

test/unit/io.js
```diff
@@ -735,6 +779,12 @@ describe('Input/output', function () {
     });
   });
 
+  it('Invalid WebP alpha quality throws error', function () {
```
lovell: Please can you add tests for invalid `lossless` and `nearLossless` attributes also.

rnanwani: Invalid `lossless` and `nearLossless` attributes are ignored and are set to default boolean `false`. Similar to what is done for `jpegOvershootDeringing` and other such params

lib/output.js
rnanwani: `nearLossless` has to be used in conjunction with Q. `nearLossless` with Q=100 is equivalent to  `lossless` . But with other quality settings, it will lead to a smaller file size than `lossless`

lib/constructor.js
rnanwani: Tests on my system passed without this as well. Still added it.

src/pipeline.cc
rnanwani: Removed the check since the default has been added

test/unit/io.js
rnanwani: Thanks for pointing this out

rnanwani:
If it looks good now, I can squash it into a single commit?

docs/api-operation.md
```diff
@@ -169,7 +169,7 @@ This operation will always occur after resizing and extraction, if any.
 // to the top, left and right edges and 20 to the bottom edge
 sharp(input)
   .resize(140)
-  .background({r: 0, g: 0, b: 0, a: 0})
+  .background({r: 0, g: 0, b: 0, alpha: 0})
```
lovell: This change was included in commit 4730554 so I think this file can be removed from the PR.

lib/output.js
lovell: Sorry, my question wasn't very clear. What happens when trying to combine the two?
```javascript
.webp({
  lossless: true,
  nearLossless: true
})...
```
Is this valid?

rnanwani:
Since this is valid `vips copy clovia.jpg clovia_vips_1.webp[lossless,near_lossless]` and uses near_lossless with default Q setting, then I guess even 
```
.webp({
  lossless: true,
  nearLossless: true
})
```
should be considered valid.

lovell:
Fantastic, thank you very much for this Rahul.

Issue #731, opened 2017-03-12
Could reducing the size of an image support nearest-neighbor?


alice0meta:
I have some retina screenshots of a non-retina app, and I want to get the non-retina version of them.

In general, nearest neighbor enlargement is invertible (easily if it's an integer enlargement, maybe less easily if it's like 1.6 or something?) so I want its inverse function.

lovell:
Hello, this should be easy to add as libvips already supports the "nearest" kernel for image reductions. Happy to accept a PR for this.

We'd need to add the "nearest" entry to the [list of known kernels](https://github.com/lovell/sharp/blob/master/lib/resize.js#L38), update the [jsdocs](https://github.com/lovell/sharp/blob/master/lib/resize.js#L64) and add a [reducer test](https://github.com/lovell/sharp/blob/master/test/unit/interpolation.js#L9).

alice0meta:
:) I tried looking through that code and got a bit lost - I couldn't figure out where the code sends the kernel string to libvips & which kernels libvips actually supports. Are you saying it's just like three lines of code necessary to add this?

lovell:
Yes, this is probably as easy as you're thinking :)

> "the code sends the kernel string to libvips"

https://github.com/lovell/sharp/blob/8b5d8a0577120109311a96f2b30e78b41519387d/src/pipeline.cc#L393-L394

> "which kernels libvips actually supports"

http://www.vips.ecs.soton.ac.uk/supported/current/doc/html/libvips/libvips-resample.html#VipsKernel

lovell:
This will be in v0.17.3 - thanks for the PR.

lovell:
v0.17.3 now available, thanks for your help!

PR #732, opened 2017-03-13
Add support for the "nearest" kernel for image reductions

alice0meta:
As recommended in [#731](https://github.com/lovell/sharp/issues/731)


lovell:
Marvellous, thank you!

PR #748, opened 2017-03-30
Add some more heroku info

sagivf:
folder size checked using `heroku run bash`

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -84,7 +84,7 @@ cd /usr/ports/graphics/vips/ && make install clean
 ### Heroku
 
 libvips and its dependencies are fetched and stored within `node_modules\sharp\vendor` during `npm install`.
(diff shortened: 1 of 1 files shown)


lovell:
Hello, thanks for this PR.

> This involves an automated HTTPS download of approximately 22MB.

22MB refers to the on-disk space rather than the number of bytes that have to be download, which is a highly-compressed tarball of about 6.5MB, so this sentence is a bit confusing.

The Linux section in the same file already contains this information. I think it's fairly well-known that Heroku is Ubuntu and therefore Linux-based so is the existing content enough?

sagivf:
Updated the size. Amm, i'm not sure it's that well known. I think it's good to keep all the platforms consistent, users don't always read everything. But what ever you see fit works for me 👍

lovell:
Thank you!

PR #754, opened 2017-04-04
Fixes API docs: sharp.format is not a function

cspotcode:


lovell:
Hello, thanks for this.

```
```

Please can you update the incorrect JSDoc at [constructor.js#L214](https://github.com/lovell/sharp/blob/master/lib/constructor.js#L214) then use `npm run docs` to automagically update the docs.

lovell:
Added via 19dd6a9, thanks for the original submission.

PR #760, opened 2017-04-06
Add gentoo support to glibc detection

kub1x:


lovell:
Fantastic, thank you for testing this.

PR #769, opened 2017-04-14
Constructor should fail on single undefined input. 

kub1x:
Perhaps I lack some programming imagination, but I can't imagine a situation, where I would intentionally pass a single undefined variable into this constructor and expect it to suddenly behave differently. That seems like it would generate a unreadable piece of code hidden as intentional magic. I might as well be living in a bubble ;)

Cheers, 
Jakub


lovell:
Thanks for this, I've used it as the basis for commit c41d755, which also "flattens" the two argument form of the constructor so e.g. `sharp({create: ...})...` can be used too.

Issue #774, opened 2017-04-20
Enhancement: save 1 bit-depth TIFF images


YvesBos2:
I'm trying to create a TIFF file that contains only black or white, nothing in between. In the end result there were  always some very light grey pixels. I've fixed this by setting the quality to 100 but it still has a bit depth of 24. 
Is there a way to force using a bit depth of 1?

lovell:
Hello, did you see the [tiff](http://sharp.dimens.io/en/stable/api-output/#tiff) options, specifically `compression`?

YvesBos2:
Hello, thanks for answering.

Yes I've looked at the compression options, this gives me the best result:

```
sharp(Buffer.from(data), {raw: {width: W, height: H, channels: C}})
    .resize(H*cell_in_pix,W*cell_in_pix, {kernel: sharp.kernel.nearest,
                interpolator: sharp.interpolator.nearest})
    .toColourspace('b-w')
    .tiff({quality: 100, compression: 'deflate', predictor: 'horizontal'})
    .toFile(dir+'/'+uuidV4()+'.tiff', function(err, info) {
      ...
    });
```
But the images still have a bit-depth of 8, is there no way I can force it to 1?

lovell:
For 1-bit output, we'd need to expose the `squash` property of libvips' TIFF writer. http://jcupitt.github.io/libvips/API/current/VipsForeignSave.html#vips-tiffsave

lovell:
v0.18.0 now available, thanks for the PR!

PR #783, opened 2017-04-24
Added support for squashing TIFF files.

YvesBos:
In reference to #774 
This PR exposes the squash property from libvips.
Squashing allows 8-bit tiffs to be squashed into 1 bit tiffs.


lib/output.js
```diff
@@ -224,6 +225,13 @@ function tiff (options) {
       throw new Error('Invalid quality (integer, 1-100) ' + options.quality);
     }
   }
+  if (is.object(options) && is.defined(options.squash)) {
+    if (is.bool(options.squash)) {
+      this.options.tiffSquash = options.squash;
+    } else {
+      throw new Error('Invalid Value for squash ' + options.squash + ' Only Boolean Values allowed for options.squash.');
```
lovell: Please can you add a test for this `Invalid value...` code path.

YvesBos: Sure, thanks for your feedback.

lovell:
Fantastic, thank you. This will be in v0.18.0.

PR #790, opened 2017-04-30
Feat: add support for any 90 angle

ncoden:
**Allow to provide any positive or negative multiple of 90 to `.rotate(...)`.**

Negative angles and angles above 360 are converted to valid 0/90/180/270
rotations (0 rotations are still ignored).

#### Changes:
- [Node] Add `useExifOrientation` internal variable to know if the Exif
  orientation must be used instead of the provided angle. This allows to save a
  negative angle in the `angle` option, because the `-1` special case is not
  needed.

- [C++] Separate `CalculateRotationAndFlip` into two generic functions:
  - `CalculateExifRotationAndFlip`: Calculate the angle of rotation and
    need-to-flip for the given Exif orientation.
  - `CalculateAngleRotation`: Calculate the rotation for the given angle.

  One or the other function is used to calculate the rotation, depending on
  wether the Exif orientation tag or the provided angle must be used.

- Add unit tests for `-3690`, `-450`, `-90`, `90`, `450`, `3690` and `-3780`,
  `-540`, `0`, `180`, `540`, `3780` rotations


ncoden:
Poke @lovell

ncoden:
Made some fixes. Closing to open a new PR with a clear history.

ncoden:
https://github.com/lovell/sharp/pull/791

PR #791, opened 2017-04-30
Add support for any rotation angle

ncoden:
**Allow to provide any positive or negative multiple of 90 to `.rotate(...)`.**

Negative angles and angles above 360 are converted to valid 0/90/180/270
rotations (0 rotations are still ignored).

#### Changes:
- [Node] Add `useExifOrientation` internal variable to know if the Exif
  orientation must be used instead of the provided angle. This allows to save a
  negative angle in the `angle` option, because the `-1` special case is not
  needed.

- [Node] Change check for planed-rotation in extract, to prepare a
  rotation before extraction: check with both `angle` and `useExifOrientation`
  options.
  I think this check contains a bit too much logics on rotation options. Maybe
  we could move this condition to a dedicated function.

- [C++] Separate `CalculateRotationAndFlip` into two generic functions:
  - `CalculateExifRotationAndFlip`: Calculate the angle of rotation and
    need-to-flip for the given Exif orientation.
  - `CalculateAngleRotation`: Calculate the rotation for the given angle.

  One or the other function is used to calculate the rotation, depending on
  wether the Exif orientation tag or the provided angle must be used.

- Add unit tests for `-3690`, `-450`, `-90`, `90`, `450`, `3690` and `-3780`,
  `-540`, `0`, `180`, `540`, `3780` rotations
- Add `320x240` fixture image for tests.

#### Unrelated changes (squashed):
- Add ncoden to the list of contributors


lovell:
Hello, thanks for this, am I right in thinking the underlying logic being added here appears to be the following?
```
angle = angle % 360;
if (angle < 0)
  angle = 360 + angle;
```
Could this logic conceivably live outside sharp and be used to calculate the angle before any call to `rotate(angle)`? What use case would require this logic to live inside sharp?

ncoden:
@lovell Well, it could. It would avoid a call to sharp for `rotate(720)` for example. I put it inside sharp to improve the sharp API itself, with this architecture in mind : the core API in C++, protected by the Node API which check for arguments and throw errors. `720` is a "angle" valid argument but must be internally transformed to a "rotate" value.

It's up to you. If you want this logic to be outside the C++, I will make a commit.

lovell:
Thanks for the update. This change feels like unnecessary code to maintain and provide tests for what is effectively the `modulo(angle, 360)` operation.

```javascript
const modulo = (a, n) => (a % n + n) % n;

const normaliseAngle = angle => modulo(angle, 360);

sharp(input)
  .rotate(normaliseAngle(angle))
  ...
```

Perhaps there's an existing (tested) math library for your use case?

ncoden:
I don't think this is "unnecessary code to maintain". An embeded angle processing is practical, intuitive (rotation by 540deg is a valid rotation), easy-to-use and require no configuration or even to think about this particular case.

This is only a modulo operation with 360, nice, so it shouldn't be hard to maintain ? :)

lovell:
💯 on the use of the `useExifOrientation` flag instead of a magic value :)

Go on then, you've sold me, although I've a slight preference for the modulo stuff to live in JS-land.

ncoden:
> the core API in C++, protected by the Node API which check for arguments and throw errors. 720 is a "angle" valid argument but must be internally transformed to a "rotate" value.

So ?

ncoden:
@lovell Poke.

lovell:
Bonsoir, I'll be merging the work-in-progress [ridge](https://github.com/lovell/sharp/tree/ridge) branch to the master branch soon. Once that's done we can revisit this PR for inclusion in v0.18.0.

lovell:
Hello, are you able to rebase against master?

ncoden:
Done @lovell

lovell:
Merci beaucoup.

ncoden:
🇫🇷  <3

Issue #607, opened 2016-10-20
Integrate with forthcoming libvips v8.5.0 logging improvements


puzrin:
When i process a lot of images, sometime something like this is printed to console:

```
vips warning: VipsJpeg: bad exif meta "exif-ifd3-GPSVersionID"
vips warning: icc_transform: embedded profile incompatible with image
```

Unexpected console texts are not acceptable - that can destroy terminal output (progress bars etc). Is it possible to suppress such things (i think that should be default behaviour)?

lovell:
Hello, these are from libvips - set the `VIPS_WARNING` environment variable to suppress them, e.g. https://github.com/lovell/sharp/blob/master/package.json#L35

puzrin:
I'm looking for long-term solution. It looks strange that additional actions should be applied to make module work right (and uncontrolled console writes is a bug for server side IMHO). May be we should report to libvips that warnings should be off by default (then anyone can enable those via env)?

puzrin:
Reported to libvips. I hope, all will agree that writing to console from library is not a good practice.

lovell:
Thanks Vitaly, I'll leave this open to track integration with any new libvips API when available.

lovell:
I expect the sharp API to access libvips' log entries will probably look something like:

```javascript
sharp.log.on('warning', (err) => {
  console.log(`libvips warning ${err.message}`);
});
```

Note that this event handler is registered against the `sharp.log` singleton rather than a sharp instance as there's no obvious way to tie a log entry back to a given input/pipeline.

puzrin:
@lovell IMHO `sharp.on('warning', ...)` would be more simple, since `sharp` is already EventEmitter instance.

Anyway, those warnings looks useless. Probably it worth to disable those at all or dump to console via `debug` package. You can add interface later, if anyone request it, but i doubt this will ever happens.

I think, real problem is that now warnings poison console output, not "how to read those".

thiagomorato:
By the way I'm one affected by the warning messing up with the logs.

lovell:
An alternative (better?) "API" would be to use [util.debuglog](https://nodejs.org/api/util.html#util_util_debuglog_section).

puzrin:
I'm for `util.debuglog` or `debug`. See no difference, because i have no plans to check those warnings.

lovell:
[...]

PR #798, opened 2017-05-05
Sketch out solution to PR#793

mceachen:
In this diff (which is not in a mergable state, this is for review and guidance before it'll be ready), I've

* added an example truncated jpg for testing, 
* added two locations that might be reasonable to add a new `failOnError` flag, and 
* plumbed those changes (mostly) into the c++ codebase.

The following are incomplete in this branch:

* Plumbing the baton into `OpenInput` needs to be done, but I didn't know if you'd rather add a new options parameter to the function, or add the flag to the `InputDescriptor` struct.

* Test needs to be written that verifies that by default, only a warning is emitted by reading the truncated image

* Test needs to be written that verifies that with the failOnError set, the promise is rejected with a reasonable error message

* Documentation needs to be regenerated. I didn't bother to include the documentation changes until I got your yay/nay on where this flag should live on the javascript side

FWIW, loading from a file or buffer with this branch does make `sharp` raise a reasonable error for corrupt images, so libvips does what it says on the side of the tin 🎉 👍 

```
VipsJpeg: Premature end of JPEG file
VipsJpeg: out of order read at line 272
```

So, my questions for you:

1. Do you have a better name for this than "failOnError"? (I didn't think "fail" was a good name for the flag)
2. Where do you want to add this flag? Can/should it live in both the constructor and in the fluent configuration?
3. On the c++ side, what's the most reasonable way to plumb the flag into `OpenInput`?
4. Does it need to be added to any other libvips calls?

Thanks!

### DIFF
--- a/lib/constructor.js
+++ b/lib/constructor.js
@@ -73,6 +73,7 @@ let versions = {
  *  a String containing the path to an JPEG, PNG, WebP, GIF, SVG or TIFF image file.
  *  JPEG, PNG, WebP, GIF, SVG, TIFF or raw pixel image data can be streamed into the object when null or undefined.
  * @param {Object} [options] - if present, is an Object with optional attributes.
+ * @param {Boolean} [options.failOnErrors] - halt processing and raise an error if the image is found to be corrupt. Defaults to false.
  * @param {Number} [options.density=72] - integral number representing the DPI for vector images.
  * @param {Object} [options.raw] - describes raw pixel input image data. See `raw()` for pixel ordering.
  * @param {Number} [options.raw.width]
@@ -93,6 +94,7 @@ const Sharp = function (input, options) {
   stream.Duplex.call(this);
   this.options = {
     // input options
+    failOnErrors: false,
     sequentialRead: false,
     limitInputPixels: maximum.pixels,
     // ICC profiles
--- a/lib/input.js
+++ b/lib/input.js
@@ -262,6 +262,26 @@ function sequentialRead (sequentialRead) {
 }
 
 /**
+ * The `libvips` default is to try to process corrupt images, and on the
+ * assumption that some output (or wrong output) is better than none. Note that
+ * these warnings are disabled (in sharp v0.18+), and are emitted by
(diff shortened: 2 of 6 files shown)


jcupitt:
The libvips side looks OK to me.

`fail` doesn't work correctly for PNG images with current stable, there's a patch in 8.5.5 which should make it all function.

lovell:
Hello, thanks for this PR Matthew, looking good so far.

I think adding the `failOnError` flag as a new property of the `InputDescriptor` JS Object created by the internal `_createInputDescriptor()` function (and therefore the `InputDescriptor` C++ struct) is the way to go here as it keeps all the input-related options together and allows fine-grained control over e.g. an initial input vs overlay images. This means the "fluent" option/function wouldn't be required. 

I prefer the singular `failOnError` as I guess this will fail as soon as the first error, if any, is reached.

mceachen:
Thanks for the quick replies! 

Lovell, sounds good to me. Would you like the PR against master or ridge?

lovell:
Target the `ridge` branch please.

mceachen:
Will do. Also, where (if anyplace) would you like to add this comment?

> The `libvips` default is to try to process corrupt images, and on the assumption that some output (or wrong output) is better than none. Note that these warnings are disabled (in sharp v0.18+), and are emitted by [debuglog](https://nodejs.org/api/util.html#util_util_debuglog_section) if enabled via `NODE_DEBUG=sharp`. See [this comment thread](https://github.com/lovell/sharp/issues/607#issuecomment-298678935) for more details.
> 
> If you'd rather have processing halt with an error when image corruption is detected, add `.failOnErrors()` to your input.

lovell:
A summary of this comment as the jsdoc for the new option in the constructor should hopefully suffice. We probably don't need to mention the v0.18+ bit as the docs are already versioned.

I plan to update the `install.md` docs before the next release so will add details of exposing warnings via `NODE_DEBUG=sharp` there.

mceachen:
Github is saying this PR is from an "unknown repository", so I'm opening a new RP with the next diff revision.

PR #813, opened 2017-05-21
add Debian jessie based dockerfiles links to the documentation

gmaliar:
Hey guys, we've created over at TailorBrands a few docker files based on Debian Jessie for libvips 8.4.6 and 8.5.5 and also pre-baked node.js images as well,
We're happy to share them and maintain them as we use Sharp ourselves!

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -105,6 +105,13 @@ docker pull marcbachmann/libvips
 docker pull wjordan/libvips
 ```
 
+[Tailor Brands](https://github.com/TailorBrands) maintain
+[Debian-based Dockerfiles for libvips and nodejs](https://github.com/TailorBrands/docker-libvips).
+
+```sh
+docker pull tailor/docker-libvips
+```
+
 ### AWS Lambda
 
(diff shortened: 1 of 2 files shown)


docs/install.md
```diff
@@ -105,6 +105,13 @@ docker pull marcbachmann/libvips
 docker pull wjordan/libvips
 ```
 
+[Tailor Brands](https://github.com/TailorBrands) maintains an
```
lovell: Nitpick:
```diff
- Tailor Brands maintains an Debian-based Dockerfiles ...
+ Tailor Brands maintain Debian-based Dockerfiles ...
```

lovell:
I notice you include libpoppler in your Dockerfile. Please be aware it is GPL licensed and, as it will be running the same process as sharp, means some of your code will become GPL. I am not a lawyer but you may need to take out a commercial licence for libpoppler should client-side JavaScript be served from the same process.

gmaliar:
@lovell fixed!
Regarding libpoppler, I ran it with the relevant people, we're cool with that, thanks for the heads up though!

PR #814, opened 2017-05-21
Support toFormat('jpg')

jingsam:


lovell:
Hello, thanks for this, please can you fix the linting errors.

lovell:
Please can you add a unit test for this.

jingsam:
It seems there is no independent unit test for `output.js`. Where should I start writing a test for this?

lovell:
Thanks for the update, please add a test to [test/unit/io.js](https://github.com/lovell/sharp/blob/master/test/unit/io.js) for now (sorry that file is so big, it needs breaking up).

test/unit/io.js
```diff
@@ -253,6 +253,21 @@ describe('Input/output', function () {
       });
   });
 
+  it('Support output to jpg format', function (done) {
+    sharp(fixtures.inputJpg)
```
lovell: Is it worth selecting an input image in a format other than JPEG?

lovell:
Thank you!

PR #825, opened 2017-05-26
Added support for Tinting images using a colour

rikh42:
This works in a similar way to the Photoshop Colour Blend mode, as it preserves the image luminance, while tinting it with the specified colour.


lovell:
Hi Rik, thanks for updating the destination branch.

Perhaps we could pass a chroma value directly from JS-land? This should be a bit faster as it will avoid the colourspace conversion and band extraction of the pixel/labPixel image.

```javascript
const colour = color(rgb);
this.options.tint = [
  colour.a(),
  colour.b()
];
```

```cc
VImage chroma = image.new_from_image(colour);
// Convert to LAB colourspace
VImage lab = image.colourspace(VIPS_INTERPRETATION_LAB);
// Extract luminance from the image
VImage luminance = lab[0];
// Create the tinted version by combining the L from the original, and the Chroma from the tint colour pixel
// convert back to the original colour space
VImage tinted = luminance.bandjoin(chroma).colourspace(typeBeforeNormalize);
```

src/operations.cc
```diff
@@ -153,6 +153,40 @@ namespace sharp {
   }
 
   /*
+   * Tint an image using the colour specified, preserving the original image luminance
+   */
+  VImage Tint(VImage image, std::vector<double> colour) {
+    // Get original colourspace
+    VipsInterpretation typeBeforeNormalize = image.interpretation();
```
lovell: Is `typeBeforeNormalize` copypasta? :)

rikh42: Totally is :-0

rikh42:
Taken me a while to understand your suggestion, but yes, that should work. I'll update...

lovell:
@rikh42 Would you like to make further changes to this? I'm very happy for it to land as-is then we can create a possible future enhancement task to cover potential performance gains.

Issue #823, opened 2017-05-25
Manually set DPI TIFF image


YvesBos:
In reference to #181 .

Because we're developing an application which will be used for printing, it is absolutely necessary to have the correct dpi.

How straightforward would it be to add this functionality, is it just a matter of exposing the xres and yres in vips_tiffsave?

Could you point me in the right direction?

Thanks in advance!

lovell:
Hello, yes exposing the `xres` and `yres` parameters of libvips' `tiffsave` operation should do what you need. Happy to accept a PR for this like #783.

http://jcupitt.github.io/libvips/API/current/VipsForeignSave.html#vips-tiffsave

lovell:
v0.18.2 now available, thanks for the PR.

Andriy-Kulak:
@lovell @YvesBos is it possible to do this for jpg/png images too? I have a similar need to specify DPI for printing but I am always dealing with JPG and PNG images, not TIFF ones. Thanks!

lovell:
@Andriy-Kulak JPEG does not contain DPI information for printing. PNG can store pixels/mm in its metadata but not sure how widespread support for that is.

https://stackoverflow.com/questions/27323561/do-pngs-or-jpgs-have-a-dpi-or-is-it-irrelevant-when-building-for-retina

PRs accepted, as always!

jcupitt:
I don't know if this is any help, but the official libvips way to do this is with `copy`:

http://jcupitt.github.io/libvips/API/current/libvips-conversion.html#vips-copy

So something like:

```c++
VImage b = a.copy(VImage::option()->set("xres", 99)->set("yes", 99)); 
```

Now when `b` is written, the file format writers will attempt to tag it with the appropriate resolution metadata.

lovell:
Thanks John, the (future) ability to set xres and yres on JPEG images is now at #967

kevinswarner:
I think the instructions for tiff xres and yres are off. It indicates pixels/mm. I think is is behaving like pixels/cm. An image saved as a tiff with no res specified shows in Photoshop as 2.54 pixels/inch. Also, to get my image out at 300 pixels per inch, I have to set xres and yres to 118.1102. Maybe my math is off. Just an observation. If the values were pixels/mm, wouldn't I need to use 11.81102 as the xres and yres values to get a 300 pixels/inch image?

lovell:
@kevinswarner Yes, those numbers do sound like px/cm.

[...]

PR #828, opened 2017-05-29
Added support for explicitly setting the dpi(xres,yres) on TIFF images.

YvesBos:
In reference to #823 .
This PR exposes the xres and yres properties from libvips.

[http://jcupitt.github.io/libvips/API/current/VipsForeignSave.html#vips-tiffsave](url)


test/unit/io.js
```diff
@@ -933,6 +933,46 @@ describe('Input/output', function () {
...
+    const res = 300;
+    const image = sharp(fixtures.inputTiff8BitDepth)
+      .tiff({
+        xres: (res),
+        yres: (res)
+      })
+      .toFile(fixtures.outputTiff, (err, info) => {
+        if (err) throw err;
+        assert.strictEqual('tiff', info.format);
+        assert.strictEqual(image.options.tiffXres, res);
```
lovell: Perhaps the returned `info` Object should include `xres` and `yres` properties? Or maybe the `metadata()` operation could provide this? I'd prefer to avoid testing the "private" `options` internals of a sharp instance.

src/pipeline.h
```diff
@@ -182,6 +184,8 @@ struct PipelineBaton {
     tiffCompression(VIPS_FOREIGN_TIFF_COMPRESSION_JPEG),
     tiffPredictor(VIPS_FOREIGN_TIFF_PREDICTOR_NONE),
     tiffSquash(false),
+    tiffXres(10),
```
lovell: It looks like libvips uses a default of `1.0` - see https://github.com/jcupitt/libvips/blob/d1ca808769339a8856be69b281b72d674a9c0f75/libvips/foreign/

YvesBos: I assume you mean 10.0 right? I'll update it.

lovell: Sorry, the URL was truncated, see https://github.com/jcupitt/libvips/blob/d1ca808769339a8856be69b281b72d674a9c0f75/libvips/foreign/tiffsave.c#L305-L306 for the default of 1.

test/unit/io.js
lovell: `metadata()` provides a `density` property that might be suitable. http://sharp.dimens.io/en/stable/api-input/#metadata

src/pipeline.cc
```diff
@@ -809,6 +809,9 @@ class PipelineWorker : public Nan::AsyncWorker {
             ->set("squash", baton->tiffSquash)
             ->set("compression", baton->tiffCompression)
             ->set("predictor", baton->tiffPredictor)));
+            ->set("predictor", baton->tiffPredictor)
```
YvesBos: I also added xres, and yres when writing to buffer. However, these values do not get saved when calling the ToBuffer method on a tiff image.
e.g. when I write a test like this:
```
 it('TIFF setting xres and yres on buffer', function (done) {
    const res = 1000.0; // inputTiff has a dpi of 300 (res*2.54)
    const image = sharp(fixtures.inputTiff)
      .tiff({
        xres: (res),
        yres: (res)
      })
      .toBuffer(function (err, data, info) {
        if (err) throw err;
        sharp(data).metadata(function (err, metadata) {
          if (err) throw err;
          assert.strictEqual(metadata.density, res*2.54); // convert to dpi
          done();
        });
      });
  });
```
And then write it to disk afterwards or read the metadata, the density is still the same as the original(300dpi).
@lovell do you have any idea why?

YvesBos: The position of the brackets are fixed by the way. But the dpi on the buffer is still not working.

lovell:
Thanks for the updates, looking good.

> "I also added xres, and yres when writing to buffer. However, these values do not get saved when calling the ToBuffer method on a tiff image."

Based on the code in thie PR, it sounds like a possible bug in libvips or libtiff itself. I should have a chance to investigate this further at the weekend if you haven't been able to work it out in the meantime.

lovell:
I had a quick look at this branch and added your proposed `toBuffer` based scenario and the test passed. This was on Linux, using the pre-compiled libvips v8.5.5 binaries.

YvesBos2:
A reinstall of sharp did the trick, pretty weird as I was also on libvips v8.5.5.
Glad it's working though, thanks for your help.

YvesBos:
A reinstall of sharp did the trick, pretty weird as I was also on libvips v8.5.5.
Glad it's working though, thanks for your help.

lovell:
Fantastic, thank you very much for this Yves.

Issue #887, opened 2017-07-28
Skip shrink-on-load when trimming


kleisauke:
I forgot to skip shrink-on-load when trimming occurs in pull request https://github.com/lovell/sharp/pull/492  (we should check for `baton->trimTolerance == 0` [here](https://github.com/lovell/sharp/blob/49297d6afbbaa747895cb92ef8fafb48f70a65a8/src/pipeline.cc#L225-L229)). Sorry for that. :sweat_smile:

Test image:
![alpha-layer-2-ink](https://user-images.githubusercontent.com/12746591/28712781-b62f05b8-738c-11e7-9793-48b318a3de33.jpg)
<sup>([This image](https://github.com/lovell/sharp/blob/64f7f1d6627339e55a18c169c442f15644a845ec/test/fixtures/alpha-layer-2-ink.png) converted to JPG)</sup>

Testcase:
```js
const sharp = require('sharp');

sharp('alpha-layer-2-ink.jpg')
  .trim(10)
  .resize(300)
  .toFile('trim-resize.jpg');
```

Excepted output:
![trim-resize-excepted](https://user-images.githubusercontent.com/12746591/28712828-ea71e7dc-738c-11e7-8b06-3ab7656582ce.jpg)

Current output:
![trim-resize-current](https://user-images.githubusercontent.com/12746591/28712836-f13b64f8-738c-11e7-9ef3-45197f223663.jpg)

lovell:
Good spot, thanks for reporting. Happy for a PR to fix this if you're able.

kleisauke:
A pull request has been made to fix this bug at #888.

lovell:
PR merged, thank you very much. This will be in v0.19.0 (and a possible v0.18.x patch release should anything urgent come along).

lovell:
v0.18.3 now available, thanks for reporting and fixing this.

PR #888, opened 2017-07-28
Skip shrink-on-load when trimming (#887)

kleisauke:
I forgot to skip shrink-on-load when trimming occurs. 

The test case ensures that this will not happen again. :smile:


lovell:
Fantastic, thank you!

PR #906, opened 2017-08-07
Implement tilt by arbitrary angles

gargsms:
sharp has had no functionality to tilt an image by any arbitrary angle yet.
This PR utilizes `libvips`'s `similarity` to do an affine transformation on the image and result in an image rotated by some angle.

- [x] Code is linted
- [x] Code confirms to style guide
- [x] Contribution guide is followed
- [x] Code is accompanied by supporting documentation
- [x] Added tests
- [x] Single squashed commit


gargsms:
@lovell I had a minor typo in the commits. Should I redo the PR?

gargsms:
There does seem to an issue with the tests as well. I might need to redo this. @lovell Help needed!

lovell:
Thanks Sukant, this looks like a very useful addition, I'll make time to review it properly in the next couple of days.

Couple of initial thoughts: Do the tests pass locally for you? Could this functionality be exposed via the existing `rotate` operation?

gargsms:
Yes, the tests do pass locally for me.It is possible to expose it via rotate itself. I did not want to change the existing API, because the libvips functions for rotate and tilt are different. rotate uses `rot` and `rot45`, while tilt uses `similarity`.It is easy to accomplish both of these with `sharp().rotate()` itself.  I was unable to decide if we should overload rotate. Should we?



    On Tuesday, 8 August 2017 2:56 PM, Lovell Fuller <notifications@github.com> wrote:
 

 Thanks Sukant, this looks like a very useful addition, I'll make time to review it properly in the next couple of days.Couple of initial thoughts: Do the tests pass locally for you? Could this functionality be exposed via the existing rotate operation?—

gargsms:
*Clarification*: `sharp().rotate()` uses `rot`.

On a second thought, I think it would be better if we don't expose just a single effect that vips `similarity` can accomplish. It is possible to apply different sorts of affine transforms using `similarity`, like skew, shear, scale, and rotate. We should be able to expose all of them at once. Something like `sharp().affine({tilt: 15, skew: 2, scale: 0.3})` can be a candidate API for this.

@lovell Let me know what you think of that.

lovell:
@gargsms Sorry for the delay replying.

Yes, I prefer the `affine` API you propose (as `vips_similarity` is a thin wrapper around the rotation feature of `vips_affine` anyway).

gargsms:
@lovell Perfect. Thanks for the update. I will try to renew this PR soon.

    On Wednesday, 16 August 2017 4:52 PM, Lovell Fuller <notifications@github.com> wrote:
 

 @gargsms Sorry for the delay replying.Yes, I prefer the affine API you propose (as vips_similarity is a thin wrapper around the rotation feature of vips_affine anyway).—

gargsms:
@lovell Anything about the failed tests? They happened to all pass on my local machine. The generated PNGs had 4 channels locally, but they have 6 on the test server.

I will close this and implement affine.

lovell:
@gargsms Sorry, I haven't had a chance to check your work-in-progress branch out to see what might have been causing this.

One thing I did notice was that when a transparency channel is added, make sure the value matches the bit depth.

```cc
if (!HasAlpha(image)) {
  double const multiplier = sharp::Is16Bit(image.interpretation()) ? 256.0 : 1.0;
  image = image.bandjoin(255 * multiplier);
}
```

bentona:
@lovell it looks like this never went anywhere. Are you open to introducing this either using the current `tilt` api or overloading `rotate`? I'd love to clean up this PR & get arbitrary rotation in, but I'm not confident I could implement the full `affine` API.

lovell:
@bentona Please see #998

PR #915, opened 2017-08-18
.stats() implementation

rnanwani:
Please review the code. Especially the C++ part as I am not very confident with the language. Let me know if something else or some other helper value like `isOpaque` needs to be exposed in the output.


lovell:
Looks great, and with lots of tests too, thank you very much Rahul!

In terms of the sporadically failing values, perhaps there should be a margin of error for these? For example the difference in the sum of pixel values appears to be <0.1% across all platforms.

I realise the min and max coords will vary across test runs. Perhaps we could verify their values are positive integers that fall within the bounds of the image?

test/unit/stats.js
```diff
@@ -0,0 +1,311 @@
...
+      assert.strictEqual(32151543524, stats.channels[2]['squaresSum']);
+      assert.strictEqual(61.51724663436759, stats.channels[2]['mean']);
+      assert.strictEqual(38.96702865090125, stats.channels[2]['stdev']);
+
+      done();
+    });
+    readable.pipe(pipeline);
+  });
+
+  it('Stream in, Promise out', function (done) {
```
lovell: This test should be able to return the Promise instance, which will then remove the need to use `done` and `catch`.

src/stats.cc
```diff
@@ -0,0 +1,177 @@
...
+    vips::VImage stats;
+    sharp::ImageType imageType = sharp::ImageType::UNKNOWN;
+
+    try {
+      std::tie(image, imageType) = OpenInput(baton->input, VIPS_ACCESS_SEQUENTIAL);
+    } catch (vips::VError const &err) {
+      (baton->err).append(err.what());
+    }
+    if (imageType != sharp::ImageType::UNKNOWN) {
+      stats = image.stats();
```
lovell: Can `stats` throw? This might need to live inside the try/catch block.

rnanwani: I don't think `stats()` throws
https://github.com/jcupitt/libvips/blob/master/libvips/arithmetic/stats.c

test/unit/stats.js
rnanwani: `.stats()` does return a promise in this test. I am not sure if I got your point. A similar test exists for `metadata()` as well.

lovell: mocha deals with tests that return (and reject) a `Promise`, e.g.:
```javascript
  it('Stream in, Promise out', function () {
    const pipeline = sharp();

    fs
      .createReadStream(fixtures.inputJpg)
      .pipe(pipeline);

    return pipeline
      .stats()
      .then(function (stats) {
        ...
      });
  });
```

src/stats.cc
lovell: libvips' C++ wrapper passes through [call_option_string](https://github.com/jcupitt/libvips/blob/master/cplusplus/VImage.cpp#L459) that can throw.

rnanwani:
Is there anything I can do to fix this build?

src/stats.cc
```diff
@@ -0,0 +1,182 @@
...
+    using Nan::New;
+    using Nan::Set;
+    using sharp::MaximumImageAlpha;
+
+    vips::VImage image;
+    vips::VImage stats;
+    sharp::ImageType imageType = sharp::ImageType::UNKNOWN;
+
+    try {
+      std::tie(image, imageType) = OpenInput(baton->input, VIPS_ACCESS_SEQUENTIAL);
```
lovell: Probably best to use the configured `baton->accessMode` instead of `VIPS_ACCESS_SEQUENTIAL` as this defaults to random, which avoids over-computation from the source decoder (and should fix those libtiff test errors).

src/stats.cc
```diff
@@ -0,0 +1,182 @@
...
+    } catch (vips::VError const &err) {
+      (baton->err).append(err.what());
+    }
+    if (imageType != sharp::ImageType::UNKNOWN) {
+      try {
+        stats = image.stats();
+        int bands = image.bands();
+        double const max = MaximumImageAlpha(image.interpretation());
+        for (int b = 1; b <= bands; b++) {
+          ChannelStats cStats(stats.getpoint(STAT_MIN_INDEX, b).front(), stats.getpoint(STAT_MAX_INDEX, b).front(),
```
lovell: I think the values returned by `getpoint` are `double` precision so some of these will require `static_cast<int>` conversion.

test/unit/stats.js
```diff
@@ -0,0 +1,488 @@
+'use strict';
+
+const fs = require('fs');
+const assert = require('assert');
+// const exifReader = require('exif-reader');
```
lovell: These unused imports can be removed.

test/unit/stats.js
```diff
@@ -0,0 +1,488 @@
...
+      assert.strictEqual(true, stats.isOpaque);
+
+      // red channel
+      assert.strictEqual(0, stats.channels[0]['min']);
+      assert.strictEqual(255, stats.channels[0]['max']);
+      assert.strictEqual(true, isInAcceptableRange(stats.channels[0]['sum'], 615101275));
+      assert.strictEqual(true, isInAcceptableRange(stats.channels[0]['squaresSum'], 83061892917));
+      assert.strictEqual(true, isInAcceptableRange(stats.channels[0]['mean'], 101.44954540768993));
+      assert.strictEqual(true, isInAcceptableRange(stats.channels[0]['stdev'], 58.373870588815414));
+      assert.strictEqual(true, isInteger(stats.channels[0]['minX']) && isInRange(stats.channels[0]['minX'], 0, 2725));
```
lovell: Perhaps split these `assert(true, x && y)` statements into two separate `assert` statements to aid debugging and remove conditional logic from tests.

docs/api-input.md
```diff
@@ -4,6 +4,7 @@
 
 -   [clone](#clone)
 -   [metadata](#metadata)
+-   [stats](#stats)
```
lovell: The `docs/api-*.md` files are autogenerated so this content needs to live in the jsdocs above the new `stats` function.

src/stats.cc
rnanwani: These are the types used in the ChannelStats struct. 

```
  int min;
  int max;
  double sum;
  double squaresSum;
  double mean;
  double stdev;
  int minX;
  int minY;
  int maxX;
  int maxY;
```

min, max will always be integers. Does this look right?

lovell:
This all looks great, thank you for the updates Rahul. As soon as the `suit` WIP branch is merged into master we can look at getting this merged in too for the v0.19.0 release.

lovell:
Merged to master, thanks Rahul, this will be in v0.19.0.

sebastienbarre:
Hi. I'm having issues getting the stats from a PNG image converted to LCH space.

```
  sharp(file)
    .toColorspace('lch')
    .toBuffer()
    .then(data => sharp(data).stats())
    .then(stats => console.log(stats))
    .catch(error => signale.error('Failed getting stats. ' + error.message));
```

This returns the same stats, whether I convert to `'srgb'`, or `'lch'`, or `'lab'`. Am I not understanding the `stats` function correctly? I thought it would give me the stats for the L, C, H channels, which would be different than the stats for R, G, B channels.

Thanks

PR #919, opened 2017-08-18
Clarifying cache documentation

tendermario:
Skimming the documentation I wasn't sure if caching occurred by default. It does. I've updated the documentation to reflect this better.


lovell:
Thank you!

PR #925, opened 2017-08-28
Add failOnError

mceachen:
Note that this patch doesn't currently work--"fail" doesn't seem to apply (even when forced to `TRUE`). I am probably not configuring gyp to relay exceptions properly, or I'm not setting the `fail` option correctly for VIPS. 

Is there something obviously wrong I'm doing here?

Thanks.


lovell:
Hello, thanks for this. I think there may be logic missing from the `CreateInputDescriptor` function in `common.cc` to set the `failOnError` bool of the `descriptor` struct based on the incoming `v8::Boolean` value.

lovell:
Note to self: this requires libvips v8.5.9+

lovell:
@mceachen The `suit` branch is available with a pre-built, pre-release libvips v8.6.0. If you're willing and able we can switch the destination branch for this PR and see what happens! I suspect there may be a couple of conflicts that will need resolving.

PR #940, opened 2017-09-08
Add SHARP_DIST_AUTH_HEADER to pass basic authorization for libvips download

### LINKED ISSUE #938: Support Basic Authorization for custom libvips URLs
There doesn't currently appear to be a mechanism to download the libvips library from a different site using the `SHARP_DIST_BASE_URL` environment variable that also requires BASIC authorization. The `got` request is configured with the following code which doesn't allow for an auth header to be included:
```js
      const gotOpt = {
        agent: caw(null, {
          protocol: 'https'
        })
      };
```
Would you be open to accepting an auth header using an Environment variable like `SHARP_DIST_AUTH_HEADER`? If so, I can submit a pull request.

pbomb:
Fixes #938

Setting `SHARP_DIST_BASE_URL="https://username:password@hostname/path/"` doesn't work as the `got` library throws the error: `Error: Basic authentication must be done with auth option`. This pull request allows the user to specify `SHARP_DIST_AUTH_HEADER="username:password"` which will set the auth header on the `got` request for libvips.

### DIFF
--- a/binding.js
+++ b/binding.js
@@ -19,6 +19,8 @@ const platform = process.env.npm_config_platform || process.platform;
 
 const arch = process.env.npm_config_arch || process.arch;
 
+const distAuthHeader = process.env.SHARP_DIST_AUTH_HEADER;
+
 // -- Helpers
 
 // Does this file exist?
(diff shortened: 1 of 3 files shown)


lovell:
I think I'd prefer to have this parsed from the existing single URL. The switch to `simple-get` discussed in #938 should achieve this.

pbomb:
Closing in favor of #945

PR #945, opened 2017-09-11
Migrate from got to simple-get

### LINKED ISSUE #938: Support Basic Authorization for custom libvips URLs

pbomb:
Fixes #938

### DIFF
--- a/binding.js
+++ b/binding.js
@@ -5,7 +5,7 @@ const os = require('os');
 const path = require('path');
 
 const caw = require('caw');
-const got = require('got');
+const simpleGet = require('simple-get');
 const semver = require('semver');
 const tar = require('tar');
 const detectLibc = require('detect-libc');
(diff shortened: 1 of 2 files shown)


binding.js
```diff
@@ -95,19 +95,22 @@ module.exports.download_vips = function () {
           } catch (err) {}
         });
       });
-      const gotOpt = {
+      const url = distBaseUrl + tarFilename;
+      const simpleGetOpt = {
+        url,
```
lovell: This will probably need to be `url: url,` as Node v4 lacks destructuring support.

binding.js
```diff
@@ -95,19 +95,22 @@ module.exports.download_vips = function () {
...
         agent: caw(null, {
           protocol: 'https'
         })
       };
-      const url = distBaseUrl + tarFilename;
-      got.stream(url, gotOpt).on('response', function (response) {
+      simpleGet.stream(simpleGetOpt, function (err, response) {
```
lovell: Is this correct? Perhaps `simpleGet(simpleGetOpt, ...`?

binding.js
pbomb: will do

binding.js
pbomb: Oops, yeah. You're right. Fixing.

lovell:
Brilliant, thank you!

pbomb:
The v0.19.0 milestone still has 9 open issues, but I'd really like to get this change released to unblock my work. Any chance of making a patch release with just this change? It should be transparent to users.

lovell:
> "get this change released to unblock my work"

No worries.

Your profile suggests "work" in this context might relate to "Workday, Inc". If so, do you think your employer might be able to make a suitable donation to a local homelessness charity by way of thanks?

pbomb:
Sure. I just made a donation which they'll match. Thanks!

PR #954, opened 2017-09-20
WIP Using vips_resize and less aggressive shrink_on_load

jardakotesovec:
Here is my attempt to make progress on #848.

It consists of two adjustments. One is to use image.resize() instead of directly image.reduce() and image.shrink(). And second is leaving at least a factor of two for the final resize step for jpeg load on shrink, which is what vipsthumbnail does and without it I was still getting shifted results.

Only with combination of these two adjustments I was getting consistent position and look when generating small thumbnails from big resolutions.

My test is basically resizing from 24 different resolutions to target resolution of 35x25.

I attach zip of result images of original sharp implementation and after these adjustments to illustrate the difference.

[res_jpegshrink_vipsresize.zip](https://github.com/lovell/sharp/files/1317236/res_jpegshrink_vipsresize.zip)
[res_original.zip](https://github.com/lovell/sharp/files/1317237/res_original.zip)

These changes likely has some performance impact - so next step would be to see if performance is acceptable before we decide what to do and final clean up would be necessary. There are currently 3 failing tests - visually they seem ok (just small shift which breaks threshold).. or better than previous.

Do you have some benchmark to tell objectively how significant is it?

PR #956, opened 2017-09-21
Expose shrink on load in resize

rnanwani:


rnanwani:
Not very sure how to write a test case for this. The two images are visibly different, but after resizing in assertSimilar, no difference is left.

test/fixtures/index.js
```diff
@@ -66,6 +66,7 @@ module.exports = {
   inputJpgLarge: getPath('giant-image.jpg'),
   inputJpg320x240: getPath('320x240.jpg'), // http://www.andrewault.net/2010/01/26/create-a-test-pattern-video-with-perl/
   inputJpgOverlayLayer2: getPath('alpha-layer-2-ink.jpg'),
+  inputJpgShrinkOnLoad: getPath('shrink-on-load-test.jpg'),
```
lovell: What is the source of this test image? Are we granted the right to distribute it, e.g. via a suitable Creative Commons licence?

lovell:
The change being proposed in #954 may supersede this.

lovell:
This has been superseded by #977, which will be in v0.19.0. It exposes a `fastShrinkOnLoad` option to the `resize()` function, defaulting to `true`.

PR #960, opened 2017-09-25
Facilitate bundling of sharp

mika-fischer:
Dynamic require statements make it more difficult for a bundler to figure out the needed modules. In this case there's not much benefit in code reduction either, so just revert to static strings inside the require statement.

This (together with providing a list of dlls to include) makes it possible to bundle a nodejs application using sharp into a stand-alone executable using nexe.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,17 +1,13 @@
 'use strict';
 
 const Sharp = require('./constructor');
-[
-  'input',
-  'resize',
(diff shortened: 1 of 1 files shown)


lovell:
Hooray for simplification, thank you Mika!

(I'm sure there was a perfectly good reason for the original approach at the time, not that I can remember it now.)

PR #977, opened 2017-10-05
vips_resize for resizing

jardakotesovec:
This is follow-up for #954.

Bit struggle with failing tests. So intentionally posted tests adjustments in second commit so its easy to go back and forward to double check these adjustments.

I got
```
  1) Embed PNG with alpha channel:
     Error: Expected maximum similarity distance: 5. Actual: 12.
      at test/fixtures/index.js:157:27
      at test/fixtures/index.js:33:9
```
If you compare new expected image with original - I think it could be considered as improvement, or at least not worse result. It might have something to do with that vips_resize combine somewhat differently shrink and reduce. So in this case I just updated the expected result.

```
  2) Partial image extraction Extract then rotate:
     Error: Expected maximum similarity distance: 5. Actual: 8.
      at test/fixtures/index.js:157:27
      at test/fixtures/index.js:33:9
```
Here it gets confusing for me, because test is comparing jpeg with png which already makes difference and not sure what the intention if it has something to do with alpha handle testing. I could not tell any difference visually and by comparing jpeg with jpeg I got good similarity. But not sure if its good fix.

```
  3) Gamma correction alpha transparency:
     Error: Expected maximum similarity distance: 20. Actual: 23.
      at test/fixtures/index.js:157:27
      at test/fixtures/index.js:33:9
```
Similar situation as in 2) as its comparing png with jpeg and expected result is already in pretty poor quality. I tried png and jpeg outputs with and without this PR and could not tell difference apart from some slightly different jpeg compression.

I am happy to make any adjustments based on your feedback, but also feel free to make the adjustments yourself..  I am ok either way :-).

I would make the PR for 'lessShrinkOnLoad' thing once this is sorted out.

Thanks!


src/pipeline.cc
```diff
@@ -279,23 +279,6 @@ class PipelineWorker : public Nan::AsyncWorker {
         }
         xfactor = static_cast<double>(shrunkOnLoadWidth) / static_cast<double>(targetResizeWidth);
         yfactor = static_cast<double>(shrunkOnLoadHeight) / static_cast<double>(targetResizeHeight);
-        xshrink = std::max(1, static_cast<int>(floor(xfactor)));
-        yshrink = std::max(1, static_cast<int>(floor(yfactor)));
-        xresidual = static_cast<double>(xshrink) / xfactor;
-        yresidual = static_cast<double>(yshrink) / yfactor;
-        if (
```
lovell: I'm slightly worried that removing the `swap` here will break something but the tests that were added for the original bug #387 continue to pass. There's probably an edge case that, if ever found, we can add the fix and another test for later :)

src/pipeline.cc
```diff
@@ -279,23 +279,6 @@ class PipelineWorker : public Nan::AsyncWorker {
         }
         xfactor = static_cast<double>(shrunkOnLoadWidth) / static_cast<double>(targetResizeWidth);
         yfactor = static_cast<double>(shrunkOnLoadHeight) / static_cast<double>(targetResizeHeight);
-        xshrink = std::max(1, static_cast<int>(floor(xfactor)));
```
lovell: Can we remove the `xshrink`, `yshrink`, `xresidual` and `yresidual` variables entirely now?

test/unit/extract.js
```diff
@@ -98,6 +98,7 @@ describe('Partial image extraction', function () {
     sharp(fixtures.inputPngWithGreyAlpha)
       .extract({ left: 20, top: 10, width: 380, height: 280 })
       .rotate(90)
+      .jpeg()
```
lovell: Good catch, this is a much better test now.

test/unit/gamma.js
```diff
@@ -48,9 +48,10 @@ describe('Gamma correction', function () {
     sharp(fixtures.inputPngOverlayLayer1)
       .resize(320)
       .gamma()
+      .jpeg()
       .toBuffer(function (err, data, info) {
         if (err) throw err;
-        assert.strictEqual('png', info.format);
+        assert.strictEqual('jpeg', info.format);
         assert.strictEqual(320, info.width);
         fixtures.assertSimilar(fixtures.expected('gamma-alpha.jpg'), data, { threshold: 20 }, done);
```
lovell: I wonder if we can remove the `{ threshold: 20 }` bit now that you've updated this to (correctly) use JPEG output?

lovell:
Oh, I had an idea for the shrink-on-load optional name too: `fastShrinkOnLoad`, defaulting to the current `true` behaviour.

src/pipeline.cc
jardakotesovec: Thats what I wanted to do. But they are still used outside of the resize logic for the shrink_on_load logic.
Particularly xresidual and yresidual is used here 
https://github.com/lovell/sharp/blob/a3457a0b68cff73c9a502e75e28aeb7c074bf157/src/pipeline.cc#L245  this reduce the shrink_on_load.

and xshrink is also used in context of shrink_on_load here https://github.com/lovell/sharp/blob/a3457a0b68cff73c9a502e75e28aeb7c074bf157/src/pipeline.cc#L226

src/pipeline.cc
jardakotesovec: Thats removed because xresidual is not used after that point. Its used only before for the shrink_on_load edge case.

test/unit/gamma.js
jardakotesovec: Will try it out tomorrow.

jardakotesovec:
Sounds good. And if its fastShrinkOnLoad: false it would just use that lower shrink_on_load factor that I stole from vipsthumbnail, right?  That should be easy.. just creating test will be tricky :-).

lovell:
> "if its fastShrinkOnLoad: false it would just use that lower shrink_on_load factor that I stole from vipsthumbnail, right?"

Yes, that's correct, thank you.

In terms of testing variable shrink-on-load, you could try extracting a 9x8 "problem" region that you know will be different based on the `fastShrinkOnLoad` setting. This will prevent the `assertSimilar()` test helper from further scaling before calculating a hash value.

jardakotesovec:
@lovell So here is update. I removed successfully the treshold for `3) Gamma correction alpha transparency:`.

And added the fastShrinkOnLoad logic so you can review that in one go.

lovell:
Thanks for the updates Jarda. As predicted that one test failure is annoying. Perhaps resize to just under half (or quarter) the original size and `extract({ top: 0, left: 0, width: 9, height: 8 })` a 9x8 region in the same pipeline rather than use resize alone?

jardakotesovec:
@lovell Interesting.. Will double check that later on..  it was not failing on my computer..  maybe some oversight..

jardakotesovec:
@lovell It's really passing locally. Only difference I could think of is somewhat different vips version. Locally I compiled from vips-8.6.0-4 package (OS X). 

If I change expect to the other file (fastShrinkOnLoad: false vs fastShrinkOnLoad: true) similarity goes from 0 to 14, which seems that it can recognize the difference between desired and undesired result good enough.

Using extract I don't think it would be make much of the difference, because the issue I am capturing is around 1px offset, which is always around 1px regardless if I resize and extract or just resize.

Any idea why we are getting different result on travis?

lovell:
Are you using libjpeg or libjpeg-turbo? The shrink-on-load feature of libjpeg-turbo is SIMD-enabled and has slightly different rounding logic. The Linux and Windows builds both use the libjpeg-turbo version.

If so, perhaps just increase the threshold to 7 and add a separate test that verifies the result of `fastShrinkOnLoad: true` is *not* similar to the result of `fastShrinkOnLoad: false`?

lovell:
A rebase is probably required too. My fault, sorry. I had to rebase the suit branch against master to pick up recently merged PRs.

jardakotesovec:
@lovell On my system I have only libjpeg. And your libvips build you are providing for sharp -
 are these built also just with libjpeg on OS X? If so, I should update the test as you suggested, otherwise it would make more sense to update the expect image.

jardakotesovec:
I should also do my 'image shifting' test with jpeg-turbo to see how bad it gets there.

jardakotesovec:
@lovell Just recompiled vips with jpeg-turbo and you are right, thats really that makes the difference. So maybe compiling with turbo on os x would also make sense to for consistent result for these nuances situations :-).

Will double check how much difference it makes to my shifting issue.

jardakotesovec:
@lovell ok.. so shifting if very much the same. There were just subtle differences for some pixels shades. Let me know if in this context makes more sense to you make the test happy with both jpeg libraries (with higher treshold) or just update expected image with turbo

lovell:
We should probably support both libjpeg variants for the tests if it's as simple as increasing the threshold a little bit :)

The prebuilt libvips provided for OS X is based on the homebrew default, currently [libjpeg](https://github.com/Homebrew/homebrew-core/blob/master/Formula/vips.rb#L20). We should probably switch to the turbo version as that will make at least a 2x difference for Mac users - leave that bit with me.

jardakotesovec:
@lovell ok.. updated, anything else?

lovell:
Podivuhodný, děkuji!

PR #742, opened 2017-03-24
Added .linear(a, b) operation for levels adjustment

brandonreavis:
This just exposes the libvips' `linear` method which is useful when you need to do a manual levels adjustment and not use the more automatic `normalize` method.

PR #1024, opened 2017-11-14
Linear transform

3epnm:
Hello,

i have done some refactoring for PR at #742 including the suggested changes. However, i have created a new branch with a fresh fork from the master branch in my account. 
rfc.


src/pipeline.h
```diff
@@ -159,6 +161,8 @@ struct PipelineBaton {
     thresholdGrayscale(true),
     trimTolerance(0),
     gamma(0.0),
+    linearA(1.0),
```
lovell: Just spotted that these fields need to be initialised in the same order as they are declared to avoid a compiler warning.
```
../src/pipeline.h:83:10: warning: ‘PipelineBaton::gamma’ will be initialized after [-Wreorder]
   double gamma;
          ^
../src/pipeline.h:81:10: warning:   ‘double PipelineBaton::linearA’ [-Wreorder]
   double linearA;
          ^
```

3epnm: I like to change this - but how do i actually see these compiler warning. I use npm install to build a new release (unfortunately under windows) and there i can't see these warnings. 
(I have to admit that I'm not programming cpp for so long.)

lovell: Thanks for updating - MSVC doesn't appear to warn about this but gcc and clang do.

lovell: https://visualstudio.uservoice.com/forums/121579-visual-studio-ide/suggestions/2553854-c-compiler-should-warn-about-wrong-member-initia

lovell:
Thanks, I'll look at merging this after the suit v0.19.0 WIP branch makes its way into master.

lovell:
@3epnm Hello, there's a small conflict in the `package.json` file. Are you able to fix this?

3epnm:
I have a look

package.json
```diff
@@ -39,6 +39,7 @@
     "Guy Maliar <guy@tailorbrands.com>",
     "Nicolas Coden <nicolas@ncoden.fr>",
     "Matt Parrish <matt.r.parrish@gmail.com>",
+    "Marcel Bretschneider <marcel.bretschneider@gmail.com>"
```
lovell: I think a trailing comma is required here.

3epnm:
I made the required change in the package.json

lovell:
Thank you very much Marcel, your work will be included in the next release.

I notice from your profile that you're based in London so you may be interested in #1104

PR #1111, opened 2018-02-01
Fixed word "center" of options.gravity

edvaldoszy:


lovell:
Hello, thanks for this PR.

The use of both "center" and "centre" are accepted by the API, so the documentation is correct as it is.

(The docs are auto-generated from the source - there's a message at the top of each auto-generated file - "Update this documentation by updating the source code").

Issue #1115, opened 2018-02-06
Question: tiling


BiancoA:
I would like to change the ordering of the saved tiles. 

Using the "google" layout the tiles are stored following rows. For example on level 2, folder 0 contains files 0.jpg, 1.jpg, 2.jpg, 3.jpg that are on the same row on the starting image. I would like them to be extracted from a column.

This could be achieved (I think) by two 90 degrees rotations: a 90 degree rotation, tiling and than -90 degrees to each tile. Looking at http://jcupitt.github.io/libvips/API/current/Making-image-pyramids.md.html I saw that libvisp implements a --angle option for the tiling, could this be implemented in sharp? 

I could contribute with code myself if someone could point me to the part of the code where the interface with libvisp and tiling take place.

lovell:
Hello, a Pull Request to expose the `angle` parameter of libvips' `dzsave` operation would be most welcome if you think it will help in this scenario, thank you.

PR #402 to expose the related `container` parameter should provide some initial guidance.

lovell:
Added via #1121, thank you.

lovell:
v0.19.1 now available, thanks for the PR.

PR #1121, opened 2018-02-09
Tiling angle support

BiancoA:
following back from issue: https://github.com/lovell/sharp/issues/1115

I exposed the angle argument for the .tile() function. It didn't really solve my problem, but it rotates the image. It looks equivalent to do .rotate() before .tile()

### DIFF
--- a/src/pipeline.cc
+++ b/src/pipeline.cc
@@ -1264,7 +1264,7 @@ NAN_METHOD(pipeline) {
   baton->tileSize = AttrTo<uint32_t>(options, "tileSize");
   baton->tileOverlap = AttrTo<uint32_t>(options, "tileOverlap");
   std::string tileContainer = AttrAsStr(options, "tileContainer");
-  baton->tileAngle = AttrTo<uint32_t>(options, "tileAngle");
+  baton->tileAngle = AttrTo<int32_t>(options, "tileAngle");
   if (tileContainer == "zip") {
     baton->tileContainer = VIPS_FOREIGN_DZ_CONTAINER_ZIP;
   } else {
--- a/src/pipeline.cc
+++ b/src/pipeline.cc
@@ -1264,6 +1264,7 @@ NAN_METHOD(pipeline) {
   baton->tileSize = AttrTo<uint32_t>(options, "tileSize");
   baton->tileOverlap = AttrTo<uint32_t>(options, "tileOverlap");
   std::string tileContainer = AttrAsStr(options, "tileContainer");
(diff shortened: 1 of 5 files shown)


src/pipeline.cc
```diff
@@ -1263,6 +1264,7 @@ NAN_METHOD(pipeline) {
   baton->tileSize = AttrTo<uint32_t>(options, "tileSize");
   baton->tileOverlap = AttrTo<uint32_t>(options, "tileOverlap");
   std::string tileContainer = AttrAsStr(options, "tileContainer");
+  baton->tileAngle = AttrTo<uint32_t>(options, "tileAngle");
```
lovell: The use of `uint32_t` here will discard negative values. `int32_t` should suffice.

test/unit/tile.js
```diff
@@ -146,13 +146,38 @@ describe('Tile', function () {
 
   it('Prevent larger overlap than default size', function () {
     assert.throws(function () {
-      sharp().tile({overlap: 257});
+      sharp().tile({
```
lovell: I notice there are code layout changes unrelated to this PR. Is this using `prettier`? (Not a problem, more for interest.)

test/unit/tile.js
```diff
@@ -192,6 +217,26 @@ describe('Tile', function () {
...
+        })
+        .toFile(fixtures.path('output.512_90.dzi'), function (err, info) {
+          if (err) throw err;
+          assert.strictEqual('dz', info.format);
+          assert.strictEqual(2725, info.width);
+          assert.strictEqual(2225, info.height);
+          assert.strictEqual(3, info.channels);
+          assert.strictEqual('undefined', typeof info.size);
+          assertDeepZoomTiles(directory, 512, 13, done);
+        });
```
lovell: I wonder if there's a reliable way to test that the tiles have actually been rotated by 90 degrees? Perhaps inspect one tile against a known fixture?

src/pipeline.cc
BiancoA: fixed! Sorry for the headless copy-paste

test/unit/tile.js
BiancoA: I use atom editor and this is the result of Beautify. I like it, but If you prefer to be consistent with the other convention you were using I can revert these changes.

test/unit/tile.js
BiancoA: this may work, what do you think?

lovell:
Brilliant, thanks for the updates Andrea. This feature will be included in the forthcoming v0.19.1.

Issue #1159, opened 2018-03-15
Feature request median filter


BiancoA:
Hi,

I would like to have the possibility to apply a median filter to an image, as documented on: https://jcupitt.github.io/libvips/API/8.6/libvips-morphology.html#vips-median .

I can try to provide a PR, any suggestion on how to do it?

BiancoA:
I am working on it following what was done for `blur`

lovell:
That's great, thank you Andrea, let me know if you need any help.

lovell:
v0.20.1 now available, thank you for the PR.

PR #1161, opened 2018-03-16
median filter

BiancoA:
following https://github.com/lovell/sharp/issues/1159 I added the median filter using vips_rank()


src/pipeline.cc
```diff
@@ -544,7 +546,10 @@ class PipelineWorker : public Nan::AsyncWorker {
         image = image.embed(baton->extendLeft, baton->extendTop, baton->width, baton->height,
           VImage::option()->set("extend", VIPS_EXTEND_BACKGROUND)->set("background", background));
       }
-
+      // Median - must happen before blurring, due to the utility of blurring after thresholding
+      if (shouldApplyMedian) {
+        image = sharp::Median(image, baton->medianSize);
```
lovell: This can probably use libvips' `median` operation, something like:
```cc
if (baton->medianSize > 0) {
  image = image.median(baton->medianSize);
}
```

src/pipeline.cc
```diff
@@ -1194,6 +1199,7 @@ NAN_METHOD(pipeline) {
   baton->flatten = AttrTo<bool>(options, "flatten");
   baton->negate = AttrTo<bool>(options, "negate");
   baton->blurSigma = AttrTo<double>(options, "blurSigma");
+  baton->medianSize = AttrTo<double>(options, "medianSize");
```
lovell: The size is integral so `uint32_t` would make a better type here.

BiancoA: fixed! sorry - again - for the copy-paste.

src/pipeline.cc
BiancoA: Done, although I am missing something... do we still need 

``` 
 VImage Median(VImage image, int const m) {
      return image.rank(m, m, (m*m)/2);
}
```
at https://github.com/lovell/sharp/pull/1161/files/1cbeabbed071bc6ad7543e2c80bd5989f5e07221#diff-3ad302ad0bfc04ecfaaf7c27e0b92174R362 ?

src/operations.cc
```diff
@@ -355,4 +355,11 @@ namespace sharp {
       return image.linear(a, b);
     }
   }
+
+  /*
+   * Median filter with mask size m
+   */
+  VImage Median(VImage image, int const m) {
```
BiancoA: here, do we still need this?

lovell: Good spot, yes, this code is no longer used. Please go ahead and remove it and its definition in the header file, thank you.

lovell:
Thank you, this will be in v0.20.1.

PR #1165, opened 2018-03-21
Skip libvips compilation with SHARP_IGNORE_GLOBAL_LIBVIPS env var

### LINKED ISSUE #1164: Forcing the use of pre-built binaries anyway
Hello,

I'm glad to discover the existence of pre-built binaries :-)
I'd like to speed up the Continuous Integration build process on a project I'm working on.

Is there any way to use pre-built binaries, regardless of the installed libvips version?

[`node-sass`](https://github.com/sass/node-sass#binary-configuration-parameters) provides a handy config. 

I added 

```
  "nodeSassConfig": {
    "binarySite": "https://github.com/sass/node-sass/releases/download"
  },
```

to my package.json file and it now downloads the pre-built binaries automatically.

Thanks for your help :-)

oncletom:
fixes #1164

### DIFF
--- a/test/unit/libvips.js
+++ b/test/unit/libvips.js
@@ -58,5 +58,13 @@ describe('libvips binaries', function () {
       const hasVendoredLibvips = libvips.hasVendoredLibvips();
       assert.strictEqual('boolean', typeof hasVendoredLibvips);
     });
+    it('useGlobalLibvips can be ignored via an env var', function () {
+      process.env.SHARP_IGNORE_GLOBAL_LIBVIPS = 1;
+
+      const useGlobalLibvips = libvips.useGlobalLibvips();
+      assert.strictEqual(false, useGlobalLibvips);
+
+      delete process.env.SHARP_IGNORE_GLOBAL_LIBVIPS;
+    });
   });
 });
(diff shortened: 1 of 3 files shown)


test/unit/libvips.js
```diff
@@ -12,6 +12,7 @@ const setPlatform = function (platform) {
 
 const restorePlatform = function () {
   setPlatform(originalPlatform);
+  delete process.env.SHARP_IGNORE_GLOBAL_LIBVIPS;
```
lovell: The platform-agnostic tests don't (currently) call `restorePlatform` so I think this env var won't get deleted. Perhaps the `delete` could move to the new test itself, after the assertion?

oncletom: Ah! Good call 😅  That explains why I saw the `process.env.SHARP_IGNORE_GLOBAL_LIBVIPS` being used 5 times on coveralls. I could explain this to myself.

I'll move the delete to the test itself then unless if I find a cleaner solution :-)

oncletom:
@lovell I've made the change. There are 8 jobs, and the coverage mentions the new code branch has been used 8 times. I guess it's fixed :-)

By the way, thank you so much by this library, and for the underlying work you have put into it. I'm amazed by its quality and speed 👍

lovell:
Thank you / merci beaucoup!

oncletom:
You're welcome 🙂

PR #654, opened 2016-12-16
Added support for multi page TIFFs

Brigadier1-9:
This PR adds basic support for multi page TIFF files by supporting a 'page' property on the options object; this allows selecting of a particular page of a multi page TIFF file.

to do this, just add the page to the options:

```
// selects the 5th page of a multi page TIFF file
var image = sharp(file, {page:4})
```
Note the page counts are zero based.

There is no support for page count retrieval included as, in my use case, I need to retrieve all the pages one at a time anyway, so I am dealing with that by looping though the source tiff until I get an error.

PR #1204, opened 2018-04-24
Support page selection for multi-page TIFF

woolite64:
This PR is based off of #654 with changes requested by @lovell. It adds support for zero-indexed page selection from multi-page TIFF's. 

The included test image was generated from the existing G31D.TIF. The second page is a 50% scaled copy of the first page.

The error message for an invalid page index could be more informative. If there is an easy way to determine the TIFF page count with libvips, please let me know. Feedback appreciated. :)


woolite64:
Removed the static_cast and updated contributors.

Thanks for the review, and I'll keep an eye on #1205.

Issue #1207, opened 2018-04-25
Support CCITTFAX4 compression for TIFF write


woolite64:
libvips provides support for lossless CCITT Group 4 compression of 1-bit TIFF's. sharp could expose this with an additional enum value for compression, ccittfax4.

This does not require a predictor, and is only applicable for squashed/b-w images.

lovell:
v0.20.2 now available with this feature, thank you for the PR!

PR #1208, opened 2018-04-25
Add Group 4 (CCITTFAX4) compression to TIFF output

woolite64:
This PR addresses #1207, with one concern mentioned below. Any insight/feedback is appreciated.

It isn't clear to me why the test case required `.toColourspace('b-w')` to compress properly. If we check the input document's metadata, the space is already 'b-w'.

Output of fixtures.inputTiff metadata:
```js
{
  format: 'tiff',
  width: 2464,
  height: 3248,
  space: 'b-w',
  channels: 1,
  depth: 'uchar',
  density: 300,
  hasProfile: false,
  hasAlpha: false,
  orientation: 1
}
```


lovell:
Hello again, thanks for this, looks great. The default colour behaviour is to convert to sRGB, hence the need to use `toColourspace('b-w')`. Perhaps this could/should be inferred from the TIFF output options?

Issue #1223, opened 2018-05-09
Error while running in NW.js


michaelw3:
Hi,
i installed sharp into an `NW.js` project but can't get it to run.
i did rebuild it using `nw-gyp rebuild --target=0.13.0 --arch=x64`, worked for me in the previous versions (target0.12.0) just fine.
here is the error log from the browser
```
Uncaught Error: The specified module could not be found.
\\?\C:\Users\*****\BookViewerApp-rebuil\node_modules\sharp\build\Release\sharp.node
    at Object.Module._extensions..node (internal/modules/cjs/loader.js:749:18)
    at Module.load (internal/modules/cjs/loader.js:613:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:552:12)
    at Function.Module._load (internal/modules/cjs/loader.js:544:3)
    at Module.require (internal/modules/cjs/loader.js:650:17)
    at require (internal/modules/cjs/helpers.js:20:18)
    at Object.<anonymous> (C:\Users\****\code\BookViewerApp-rebuil\node_modules\sharp\lib\constructor.js:10:15)
    at Module._compile (internal/modules/cjs/loader.js:702:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:719:10)
    at Module.load (internal/modules/cjs/loader.js:613:32)
```

node version - 10.1.0
nw-gyp version - 3.6.3
sharp version - 0.20.2

i got a lot of warnings during the rebuild not sure it has anything to do with it because it resulted in:
`gyp info ok`

here is a snippet of the warnings:
```
c:\users\****\code\bookviewerapp-rebuil\node_modules\sharp\vendor\include\vips\VImage8.h(477): warning C4267: 'argum
ent': conversion from 'size_t' to 'int', possible loss of data (compiling source file ..\src\pipeline.cc) [C:\Users\*****\code\BookViewerApp-rebuil\node_modules\sharp\build\sharp.vcxproj]
C:\Users\*****\.nw-gyp\0.13.0\src\node_buffer.h(13): warning C4068: unknown pragma (compiling source file ..\src\shar
p.cc) [C:\Users\*****\code\BookViewerApp-rebuil\node_modules\sharp\build\sharp.vcxproj]
C:\Users\*****\.nw-gyp\0.13.0\src\node_buffer.h(14): warning C4068: unknown pragma (compiling source file ..\src\shar
[...]

PR #1229, opened 2018-05-14
rebuild doc for NW.js

michaelw3:
refers to https://github.com/lovell/sharp/issues/1223


lovell:
Brilliant, thank you.

Issue #846, opened 2017-06-15
Tips for implementing small blurred previews?


wmertens:
Facebook Engineering posted this a while ago: https://code.facebook.com/posts/991252547593574/the-technology-behind-preview-photos/

How could I do the same with sharp? I imagine I'd resize the image, maybe apply blur to it (before or after resize?) and then grab the image from the buffer, as well as the image dimensions, and store these in the image database.

Then convert this to base64, and on the client side make a data url containing the default header, dimensions and image data. When using server-side rendering, these images could be embedded in the html (should compress well), and css could take care of the blur client side. => instant images

Does that sound reasonable? I just don't know what parts of the header need to be retained, and if blurring while resizing is a good idea.

I suppose this is more of a usage question than a feature request, so if I should be asking this elsewhere I apologize.

papandreou:
Looks like the lqip webpack loader does [exactly that using sharp](https://github.com/zouhir/lqip-loader/blob/master/index.js#L65-L76). Showcase here: https://lqip-loader.firebaseapp.com/

lovell:
In sharp, a blur operation will always occur after a resize, if any, a behaviour that matches what would be required here.

The Facebook "trick" is to have the client already hold a common JPEG header, including Huffman tables, for a given JPEG percentage "quality". This fixed header can be removed from the data transferred by the server, then prepended to the incoming compressed pixel data on the client.

First thing I'd do would be to generate a few JPEG images of the same dimensions and quality and see how many bytes at the start are the same.

If you're using base64 then I can't not recommend https://github.com/lovell/64 :)

lovell:
@wmertens Hope these answers helped, please feel free to re-open with more details if not.

tomlokhorst:
I'm also looking into building small preview images.

It seems the Facebook "trick" doesn't work with sharp, because the Huffman tables [are optimised when generating JPEGs](http://sharp.pixelplumbing.com/en/stable/#optimal). Therefor the headers for different JPEGs are different from each other.

Am I understanding this correctly?

PR #1265, opened 2018-06-20
Expose the optimize_coding flag libvips as an option

tomlokhorst:
This exposes the `optimize_coding` flag from libvips, which was previously hardcoded to always be `TRUE`. By default this still remains enabled, but now it's possible to disable optimisation.

My use case is https://github.com/lovell/sharp/issues/846#issuecomment-398883207, I want to generate JPEG files with a constant header.

Admittedly this is a very small and particular use case, and I would understand if you don't want to expose this option in the library, but it would really help me.

### DIFF
--- a/lib/output.js
+++ b/lib/output.js
@@ -148,6 +148,8 @@ function withMetadata (withMetadata) {
  * @param {Boolean} [options.overshootDeringing=false] - apply overshoot deringing, requires mozjpeg
  * @param {Boolean} [options.optimiseScans=false] - optimise progressive scans, forces progressive, requires mozjpeg
  * @param {Boolean} [options.optimizeScans=false] - alternative spelling of optimiseScans
+ * @param {Boolean} [options.optimiseCoding=true] - optimise Huffman coding tables
+ * @param {Boolean} [options.optimizeCoding=true] - alternative spelling of optimiseCoding
  * @param {Boolean} [options.force=true] - force JPEG output, otherwise attempt to use input format
  * @returns {Sharp}
  * @throws {Error} Invalid options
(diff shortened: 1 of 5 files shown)


lib/output.js
```diff
@@ -148,11 +148,14 @@ function withMetadata (withMetadata) {
...
  * @param {Boolean} [options.optimiseScans=false] - optimise progressive scans, forces progressive, requires mozjpeg
  * @param {Boolean} [options.optimizeScans=false] - alternative spelling of optimiseScans
+ * @param {Boolean} [options.optimiseCoding=true] - optimise Huffman coding tables
+ * @param {Boolean} [options.optimizeCoding=true] - alternative spelling of optimiseCoding
  * @param {Boolean} [options.force=true] - force JPEG output, otherwise attempt to use input format
  * @returns {Sharp}
  * @throws {Error} Invalid options
  */
 function jpeg (options) {
+  this._setBooleanOption('jpegOptimiseCoding', true);
```
lovell: Please can this move to the default options in [constructor.js](https://github.com/lovell/sharp/blob/master/lib/constructor.js#L188)

lib/output.js
```diff
@@ -185,6 +188,10 @@ function jpeg (options) {
         this.options.jpegProgressive = true;
       }
     }
+    options.optimiseCoding = is.bool(options.optimizeCoding) ? options.optimizeCoding : options.optimiseCoding;
```
lovell: Thank you for considering the two alternative spellings here :+1:

test/unit/io.js
```diff
@@ -826,6 +826,36 @@ describe('Input/output', function () {
       });
   });
 
+  it('Optimise coding generates smaller output length', function (done) {
```
lovell: Always a pleasure to see a PR with unit tests, thank you.

tomlokhorst:
I've updated the PR with the requested change.

Btw, this is a very nice project! Well documented and easy to contribute to.

lovell:
Thanks again, this will be released as part of either v0.20.5 or v0.21.0, whichever comes first.

lovell:
v0.20.5 now available.

PR #1285, opened 2018-07-03
Expose mozjpeg quant_table flag

rexxars:
This exposes the quant_table flag from mozjpeg, which previously had the libjpeg/mozjpeg default of `0`. Will use `0` as the default, but now it's possible to choose a specific quantisation table when paired with the `optimize_coding` flag that was recently added.

This can be very useful in a certain cases where some tables provide a better quality/compression ratio for certain images, which opens up a few cool tricks.

I unfortunately could not figure out a way to run the tests in a reliable way since mozjpeg is usually not installed, but it seems other tests using mozjpeg features use a similar "same or less" approach - let me know if you have any pointers on how this can be improved. Perhaps if there was some way to check if mozjpeg is available, it could be exposed by sharp somehow and we could do a `skip` on the specific tests if not available?

### DIFF
--- a/test/unit/io.js
+++ b/test/unit/io.js
@@ -389,6 +389,16 @@ describe('Input/output', function () {
     });
   });
 
+  describe('Invalid JPEG quantisation table', function () {
+    [-1, 88.2, 'test'].forEach(function (table) {
+      it(table.toString(), function () {
+        assert.throws(function () {
+          sharp().jpeg({ quantisationTable: table });
+        });
+      });
+    });
+  });
+
   it('Progressive JPEG image', function (done) {
     sharp(fixtures.inputJpg)
       .resize(320, 240)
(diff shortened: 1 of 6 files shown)


lovell:
Hei Espen, thanks for this PR, I should be able to take a closer look in a couple of days.

I think it might reveal a small bug in the way libvips currently only sets the table based on `quant_table` being non-zero. However mozjpeg uses table 3 by default, which means you can never "reset" mozjpeg to use table 0 via libvips. It also means that you might think you're using table 0 "by default" when you're actually getting table 3.

rexxars:
I've compiled against libvips 8.6.4 and mozjpeg 3.3.1, and I think you're onto something, but I'm not seeing _quite_ the behavior you are suggesting. It seems to me that (with this PR, at least) I am able to freely change between the different quantization tables:

```js
onst sharp = require('./')

async function main() {
  for (let i = 8; i >= 0; i--) {
    await sharp('img.jpg')
      .resize(1024, 1024)
      .max()
      .jpeg({
        optimizeCoding: false,
        quantisationTable: i
      })
      .toBuffer()
      .then(res => console.log('Table %d, %d bytes', i, res.length))
      .catch(err => console.error(err))
  }
}

main()
```

Yields the following output:
```
Table 8, 165658 bytes
Table 7, 186603 bytes
Table 6, 146376 bytes
Table 5, 117315 bytes
Table 4, 126824 bytes
Table 3, 115800 bytes
Table 2, 119055 bytes
Table 1, 202427 bytes
Table 0, 126018 bytes
```

As you pointed out however, libjpegs default quantization table is 0, while mozjpeg is 3, and this PR sets it to 0 by default.

lovell:
Thanks for checking. I'd forgotten about the libvips logic, which I added, to reset mozjpeg to use libjpeg defaults and therefore table 0 - see https://github.com/jcupitt/libvips/commit/00e27def8ea841db6b3ec9da00bbdc976241c4ae#diff-6e6f82dc292e7f7f8d223ff3dff4c7faR913

test/unit/io.js
```diff
@@ -856,6 +866,37 @@ describe('Input/output', function () {
...
+        assert.strictEqual(true, withDefaultQuantisationTable.length > 0);
+        assert.strictEqual(withDefaultQuantisationTable.length, withInfo.size);
+        assert.strictEqual('jpeg', withInfo.format);
+        assert.strictEqual(320, withInfo.width);
+        assert.strictEqual(240, withInfo.height);
+        // Then generate with different quantisation table
+        sharp(fixtures.inputJpg)
+          .resize(320, 240)
+          .jpeg({ optimiseCoding: false, quantisationTable: 3 })
+          .toBuffer(function (err, withKodakTable, withoutInfo) {
```
lovell: Table 3 was hand-rolled by image processing expert and libvips contributor Nicolas Robidoux rather than tuned via the Kodak set - see http://www.imagemagick.org/discourse-server/viewtopic.php?f=22&t=20333&p=98008#p98008

rexxars: Oops. I guess the quantisation tables might benefit from having constants defined, but I have no idea how those would be named. Anyway, should I just rename it to `withQuantTable3`, or `withImageMagickTable`? Any preferences?

lovell: Let's go for your `withQuantTable3` option, thanks. Hopefully this will prevent someone (me) in the future (next month) getting confused that there might be an off-by-one error here.

rexxars: I've amended my previous commit to change the variable name, thanks for noticing!

lovell:
Thank you, this will be in v0.20.6.

rexxars:
Thanks a bunch, Lovell! sharp is fantastic ❤️

PR #1290, opened 2018-07-09
docs: fix webp alphaQuality range

sylvaindumont:
As specified [in code](https://github.com/lovell/sharp/blob/master/lib/output.js#L267) webp alphaQuality range is 1-100, not 0-100

### DIFF
--- a/lib/output.js
+++ b/lib/output.js
@@ -261,10 +261,10 @@ function webp (options) {
     }
   }
   if (is.object(options) && is.defined(options.alphaQuality)) {
-    if (is.integer(options.alphaQuality) && is.inRange(options.alphaQuality, 1, 100)) {
+    if (is.integer(options.alphaQuality) && is.inRange(options.alphaQuality, 0, 100)) {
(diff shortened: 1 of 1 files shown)


lovell:
Bonjour Sylvain, thanks for spotting this inconsistency and providing a PR.

I think the value checking code is wrong here rather than the docs as the range is 0-100.

> "The quality factor quality_factor ranges from 0 to 100 and controls the loss and quality during compression. The value 0 corresponds to low quality and small output sizes, whereas 100 is the highest quality and largest output size." https://developers.google.com/speed/webp/docs/api#simple_encoding_api

sylvaindumont:
Bonjour Lovell, thanks for the quick response.
I updated the pr to check the 0-100 range and reverted the docs change.
I was converting some jpeg to webp and get an error when setting `alphaQuality: 0`, effectively the 0-100 range makes more sense.

lovell:
Thank you, this fix will be in v0.20.6

PR #1316, opened 2018-07-31
Updated the "make a transparent image" example

thegareth:
Alpha for the JS color module is between 0-1, not 0-255.

### DIFF
--- a/docs/api-constructor.md
+++ b/docs/api-constructor.md
@@ -55,7 +55,7 @@ sharp({
     width: 300,
     height: 200,
     channels: 4,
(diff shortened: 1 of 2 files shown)


lovell:
Hello, thank you for this PR.

The `docs/api-*` files are automagically generated from source code. Please can you make this change to the example [here](https://github.com/lovell/sharp/blob/master/lib/constructor.js#L84).

thegareth:
my apologies, done.

lovell:
No need to apologise! Thank you very much for spotting and fixing this mistake.

PR #1325, opened 2018-08-07
Fixing image quality instructions at composite lib

rodrigoalviani:


lovell:
Good spot, thank you!

PR #1340, opened 2018-08-16
Update install.md to include AWS Lambda example

dioptre:


lovell:
Hello, thanks for this. The example you link to seems to be mostly S3 configuration rather than Lambda. Have I misunderstood?

lovell:
Closing for now but please re-open with more information if this is still relevant.

Issue #1281, opened 2018-06-29
Enhancement: pass tile depth option to dzsave


dlong500:
Is there a way to pass other options to the `dzsave` command when writing an image to the DZI format?  I would like to pass the `--depth onetile` option to prevent an excessive number of small (and unused) files from being generated.

lovell:
Hello, the `depth` flag would need to be exposed. Happy to accept and/or help with a PR if you're able. The existing `layout` property of `tile()` can be used as a guide for this.

lovell:
v0.20.6 now available with this feature thanks to @alundavies

PR #1342, opened 2018-08-16
Pass tile depth parameter down to libvips

alundavies:
Hi Lovell,

Stumbled across this issue while using sharp https://github.com/lovell/sharp/issues/1281 and thought I'd have a bash at it...

I don't see 'teeth' in the list of branches as a target for this pull request, am I missing something or am I right to use master (it should be non-breaking but it is a change/addition to the api)

Since libvips chooses a default depth based on the layout detected, I thought it best that sharp didn't choose one itself and instead not provide anything unless explicitly set in the javascript.

Alun.


lovell:
Marvellous, thank you very much Alun, this will be included in v0.20.6.

PR #1355, opened 2018-08-24
Avoid race conditions when creating cache directories

ajhool:


lovell:
Thanks for this. Is it worth having this logic in a dedicated "ensureDir" function in `lib/libvips.js` that can be called where needed?

ajhool:
Looks like there are five mkdir calls throughout the whole project.

install/dll-copy.js (2)
lib/libvips.js (2)
install/libvips.js (1)

install/libvips.js currently depends on lib/libvips.js

Maybe one function in lib/libvip.js that is also used in install/libvips.js, and a duplicate function in install/dll-copy.js?

Issue #1357, opened 2018-08-27
Allow floating-point density for arbitrary SVG render size


aeirola:
First, thanks for an awesome library! I'm using it to create a tool for rendering PNG icons for React Native Android and iOS mobile applications from SVG source files. In general, everything goes smoothly until the SVG source file happens to be the wrong size relative to the desired output size. This of course feels a bit silly for *Scalable* Vector Graphics images.

The proper way to resize vector images with the sharp library seems to be to set the `density` constructor parameter when reading a vector image. Unfortunately the value is an integer, which means that it isn't possible to produce arbitrary sized SVG files. For example, it is not possible to render a 100x100 sized SVG image as a 72px x 72px PNG image. The target density would have to be

desiredOutputSize / inputSize * inputDensity = 100 / 72 * 72 = 51.84

which is an invalid value for the `density` parameter. It produces the error `Error: Invalid density (1 to 2400) 51.84`

The only workarounds I can think of is to either load the SVG initially at a higher resolution and the scale down the resulting raster to the desired size with the `resize` function, or alternatively edit the SVG source `width` and `height` attributes on the fly before passing the file to sharp. Both solutions have their issues, with the first possibly causing resizing artefacts in the resulting image, and the second requiring additional logic to parse XML and handle absolute SVG element dimensions with correct `viewBox` attribute.

Thus, I would like to ask if there is a way to support more flexible SVG resizing in sharp. I guess just passing floating point numbers as the `density` parameter probably won't be a good idea, since it has the possibility of rounding errors. Maybe it would be more appropriate to be able to pass a fixed width or height of the canvas the SVG image should be rendered on. I suppose passing both would be superfluous, since there shouldn't be a need to change the image aspect ratio.

Unfortunately though, it seems that libvips [svgload](http://jcupitt.github.io/libvips/API/current/VipsForeignSave.html#vips-svgload) only seems to support defining the rendered image size using density and scale values. Seems to support floating point numbers though, so maybe sharp could expose that API in some way then to allow a bit more control of the resulting image size.

lovell:
Hello, the `density` parameter in sharp originates from *magick support where only integer values were allowed. As you point out, librsvg rendering can handle floating point values for this parameter so we should at least remove the integer-only limitation.

Happy to accept and/or help with a PR if you're able.

aeirola:
[...]

PR #1362, opened 2018-08-31
Allow floating point density input

aeirola:
As discussed in https://github.com/lovell/sharp/issues/1357

Metadata output will still remain integer, since it contains inherent inaccuracy for most file formats which store the density as some integer value, e.g. pixels per meter in PNGs.


test/unit/io.js
```diff
@@ -939,6 +939,21 @@ describe('Input/output', function () {
...
+    sharp(fixtures.inputSvg, { density: 14.4 })
+      .toFormat('png')
+      .toBuffer(function (err, data, info) {
+        if (err) throw err;
+        assert.strictEqual('png', info.format);
+        assert.strictEqual(20, info.width);
+        assert.strictEqual(20, info.height);
+        fixtures.assertSimilar(fixtures.expected('svg14.4.png'), data, function (err) {
+          if (err) throw err;
+          done();
```
aeirola: Not sure if this should also check the metadata of the resulting image. For some reason the density seemed to be `undefined` for this particular image.

lovell:
Thank you Axel, this is great, I expect there to be a v0.20.8 out with this in a few days.

PR #1383, opened 2018-09-23
refactor: Move background extraction into separate method.

freezy:
While working on the rotation feature I came across a 25-line segment that was duplicated and that I probably will need to duplicate again, so I thought I might as well move it into a reusable function.

It's neither a bugfix nor a new feature, so tests stay the same.

If you prefer me including this in the rotation PR, let me know, but I thought it was more proper to do it separately.

### DIFF
--- a/src/common.cc
+++ b/src/common.cc
@@ -602,4 +602,40 @@ namespace sharp {
     }
   }
 
+  /*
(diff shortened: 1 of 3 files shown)


lovell:
Brilliant, thank you!

PR #1385, opened 2018-09-24
Add support for vips_rotate.

### LINKED ISSUE #998: Expose vips_rotate/vips_affine for arbitrary tilt (and/or skew and scale)
Hi There, 
Sharp is very very good,  chainable api,  ease to use.   Really appreciated for the this great project. 
While is it possible to support rotate angles other than 90/180/270/360 ?  Cuz i tested that node-easyimage has supported this feature already. 

Thanks!

freezy:
This PR adds support for arbitrary rotation angles.

Instead of only being able to rotate multiples of 90 degrees, any angle can now be applied, using libvips' `vips_rotate` API, introduced in the recently released v8.7.0.

The JavaScript API stays the same, however a new optional `background` parameter can be provided to set the background color of the rotated image.

The new code is covered in tests, let me know if that's enough or more test cases are needed. For the documentation I'm not sure why it seems to have generated additional documentation for `extend`, `extract` and `trim`, maybe those weren't generated before?

This PR closes #998.


lovell:
Thank you - I should have some time tomorrow to take a proper look at this.

test/unit/rotate.js
```diff
@@ -45,6 +83,17 @@ describe('Rotation', function () {
     });
   });
 
+  [-3750, -510, -150, 30, 390, 3630].forEach(function (angle) {
```
lovell: This is a great test scenario, thank you.

lib/operation.js
```diff
@@ -29,16 +33,29 @@ const is = require('./is');
...
  */
-function rotate (angle) {
+function rotate (angle, options) {
   if (!is.defined(angle)) {
     this.options.useExifOrientation = true;
   } else if (is.integer(angle) && !(angle % 90)) {
     this.options.angle = angle;
+  } else if (is.number(angle)) {
+    this.options.rotationAngle = angle;
+    if (options && options.background) {
```
lovell: There's an `is.object` check available that could be used here.

lib/operation.js
```diff
@@ -1,14 +1,18 @@
...
 
 /**
  * Rotate the output image by either an explicit angle
  * or auto-orient based on the EXIF `Orientation` tag.
  *
- * If an angle is provided, it is converted to a valid 90/180/270deg rotation.
+ * If an angle is provided, it is converted to a valid positive degree rotation.
  * For example, `-450` will produce a 270deg rotation.
  *
+ * If a non-rectangular angle is provided, the color of the background color
```
lovell: I think the mathematical term would be "non-quadrantal angle" but maybe "angles that are not multiple of 90" would be better understood.

lib/operation.js
```diff
@@ -29,16 +33,29 @@ const is = require('./is');
...
+      this.options.rotationBackground = [
+        backgroundColour.red(),
+        backgroundColour.green(),
+        backgroundColour.blue(),
+        Math.round(backgroundColour.alpha() * 255)
+      ];
+    }
   } else {
-    throw new Error('Unsupported angle: angle must be a positive/negative multiple of 90 ' + angle);
+    throw new Error('Unsupported angle: angle must be a positive or negative number.');
```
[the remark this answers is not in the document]
freezy: Yup!

lib/constructor.js
```diff
@@ -108,6 +108,8 @@ const Sharp = function (input, options) {
     embed: 0,
     useExifOrientation: false,
     angle: 0,
+    rotationAngle: 0,
```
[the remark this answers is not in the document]
freezy: Actually no, that wasn't intentional :)

But sounds good to me. Keep it?

freezy:
Thanks for the review, pushed an update with your suggestions.

lovell:
Marvellous, thank you very much, this will be included in sharp v0.21.0.

freezy:
Wonderful, cheers!

benoj:
> Marvellous, thank you very much, this will be included in sharp v0.21.0.

When is `v0.21.0` scheduled to be released?

lovell:
@benoj You can track the progress of everything planned for inclusion in v0.21.0 at https://github.com/lovell/sharp/milestone/40

PR #1394, opened 2018-09-28
Prefer universal English in README ("Licence" => "Licensing")

MiguelAraCo:


MiguelAraCo:
It is but in software development is much more common to see "license". Actually in the README file, just below the title, "license" is used over and over whereas "licence" isn't being used at all (except for the title I'm suggesting to change).

lovell:
Hello, sharp's external API strives to support as many variants of English as possible, for example both [greyscale](http://sharp.pixelplumbing.com/en/stable/api-colour/#greyscale) and [grayscale](http://sharp.pixelplumbing.com/en/stable/api-colour/#grayscale) are provided.

The documentation uses British English, with an exception where legal boilerplate has been copy/pasted, e.g. "Licensed under the Apache License".

In all English-speaking countries apart from the US, "licence" is the noun and "license" is the verb.

To avoid confusion, how about we converge on the use of the verb-form "Licensing" for this header? If so, please can you update the similar heading in `docs/index.md` also.

MiguelAraCo:
That explains the original writing. Thank you for the explanation.

"Licensing" sounds like a good compromise. I'll update the PR with the requested changes.

lovell:
Thanks for updating, I'll merge this after the v0.21.x `teeth` branch has landed in the `master` branch.

lovell:
I've merged this manually to the `teeth` branch via commit 3c54eed, muchas gracias.

PR #1412, opened 2018-10-18
fixed a mistake in api-resize.md

pjkui:
fixed a mistake in api-resize.md


lovell:
I updated the corresponding JSDocs in commit 9dc6492, thanks again for the PR.

Issue #1418, opened 2018-10-19
add_npmrc_download_url


SethWen:
suggestion: add `process.env.npm_config_sharp_dist_base_url` in  libvips-*.tar.gz download base url, then we can config it in `npmrc` file.
That means in `./install/livavips.js` file:
`const distBaseUrl = process.env.npm_config_sharp_dist_base_url || process.env.SHARP_DIST_BASE_URL || `https://github.com/lovell/sharp-libvips/releases/download/v${minimumLibvipsVersion}/`;`

lovell:
Hello, this seems like a sensible idea, happy to accept a PR with this change if you're able.

PR #1422, opened 2018-10-22
support npm_config_sharp_dist_base_url

SethWen:
I have commit a pr for https://github.com/lovell/sharp/issues/1418.


lovell:
谢谢

Issue #1432, opened 2018-10-24
Add Mitchell kernel for resizing


Daiz:
libvips [added a Mitchell kernel for resizing in 8.7.0.](https://github.com/libvips/libvips/releases/tag/v8.7.0) ([related libvips issue](https://github.com/libvips/libvips/issues/966)), so it would be great to see support for this in sharp too now that it's using 8.7.0.

PR #1438, opened 2018-10-26
Add mitchell kernel

Daiz:
Pull request for #1432. Adds newly available (as of libvips 8.7.0) Mitchell-Netravali spline as a resize kernel option.


lovell:
Thank you!

PR #1439, opened 2018-10-27
Add second parameter to gamma operation for different output gamma

Daiz:
I initially opened [libvips#1144](https://github.com/libvips/libvips/issues/1144) for this but it turned out to be a matter in sharp instead. I have a use case where I want to use different gamma exponents for resizing and final output. Sharp currently enforces the same values both ways (well, 1/X and X to be specific, while what I want is 1/X and Y). This pull request fixes this by adding a second optional parameter to the `.gamma()` operation that controls the final output gamma. By default, it uses the same value as the first parameter so this change should be fully backwards-compatible while providing minor (but personally useful) new functionality.

Another approach that could be taken on gamma management would be to change the way gamma is handled to match libvips' underlying model (so you'd do eg. `image.gamma(1/X).resize(W, H).gamma(Y)`) but this would be a lot more complicated to implement and wouldn't be backwards-compatible, thus the current approach which should hopefully fulfill most use cases one could have for this.

So far this PR adds the feature, though I haven't added any tests yet.


lovell:
Hello, thanks for this. Are you able to add tests to cover the new logic?

Daiz:
Rebased to master, added the tests and squashed the results to one commit. Should hopefully be good to go now.

lovell:
Brilliant, thank you very much for this.

Issue #728, opened 2017-03-03
Enhancement: support multiple overlayWith in single pipeline (including blend modes) via vips_composite


ghost:
I'm looking for option to combine multiple images, side by side or one under another. 
I looked the `overlayWith` function but seems it can't do what is required. I also looked in
**libvips** options and maybe `.reduce` is what we need. Please share with me any idea of realization if there is nothing implemented yet.

lovell:
Hello, if you've not already seen them, a few similar questions are https://github.com/lovell/sharp/issues/405 https://github.com/lovell/sharp/issues/597 and https://github.com/lovell/sharp/issues/699

Using the current API, I'd probably suggest looking at https://github.com/lovell/sharp/issues/470#issuecomment-227456844 for how to start with a blank canvas. This can then be processed multiple times, each using `overlayWith` with the relevant values for `top` and `left`.

Etiene:
I'm trying to do the same thing. So far the example code on #405 looks the cleanest to me. 

However, it creates a new sharp object of the whole thing on each small image that is overlaid over a bigger one. In terms of performance this is impractical for me. Are there any other solutions? 

Thanks

lovell:
Hi @Etiene, your performance concerns may be covered by #573.

Etiene:
Hello @lovell, I installed the branch version, performance is still undoable. Are there any modifications to the code example I need to make? 

Is there no way to do something like
```
sharp('base.jpg')
  .overlayWith('mask1.png',{top:10, left:10})
  .overlayWith('mask2.png',{top:80, left:100})
  .overlayWith(.....)
  .....
  .toBuffer(....)
```
?

lovell:
@Etiene The branch referred to in #573 is a WIP and does not yet improve things. The API you propose, although unsupported at present, would make a great future possible enhancement, thank you.

wmertens:
How about, in the meantime, printing a warning when using a second overlayWith in the same pipeline? I was pretty confused when it wasn't working…

Lasto13:
This would be great

lovell:
There's a discussion about adding an image composition operation to the underlying libvips library that will help with this feature request - see https://github.com/jcupitt/libvips/issues/657

CWSpear:
The upstream issue was closed a few weeks ago. Is this unblocked now?

lovell:
[...]

PR #1440, opened 2018-10-30
Support for blend modes

tcz:
Partial fix for https://github.com/lovell/sharp/issues/728, adding blend modes.
It also makes single-pipeline multiple `overlayWith` one step closer, but it's not implemented.

image

For the order of the blend modes I used Photoshop's ordering (for those available):

image


tcz:
@lovell can you help me with the CI failures? Seems like it's not picking up the blend modes at all, maybe different libvips version?

I'm also not 100% sure about the premultiplication, some blend modes look different from when done in photoshop, not sure if it's just because it's a different implementation or I'm doing something wrong.

lovell:
Hello, thanks for this, I should be able to take a closer look at this at some point over the next few days.

Based on a very quick scan of the diff, using named blend modes that match the nicknames from libvips' [enumtypes.c](https://github.com/libvips/libvips/blob/7da31386c73cacc17116529463529014f8525cc1/libvips/iofuncs/enumtypes.c#L385-L409) would be more consistent and allow the use of `vips_enum_from_nick`. See the `kernel` option in `resize()` for how this could be simplified.

docs/api-composite.md
```diff
@@ -28,6 +28,7 @@ If the overlay image contains an alpha channel then composition with premultipli
         -   `options.create.height` **[Number][4]?** 
         -   `options.create.channels` **[Number][4]?** 3-4
         -   `options.create.background` **([String][2] \| [Object][3])?** parsed by the [color][6] module to extract values for red, green, blue and alpha.
+    -   `options.blendMode` **[String][2]** [blend mode][8] to composite images with (optional, default `'over'`). Possible values: `over`, `darken`, `multiply`, `color burn`, `lighten`, `screen`, `color dodge`, `overlay`, `soft light`,  `hard light`, `difference`, `exclusion`
```
lovell: The `docs/api-*.md` files are auto-generated from JSDocs in the source. Please can you add this content to the JSDocs for the `overlayWith()` method then use `npm run docs` to refresh the markdown file.

docs/api-composite.md
```diff
@@ -66,3 +67,5 @@ Returns **Sharp**
 [6]: https://www.npmjs.org/package/color
 
 [7]: https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Error
+
+[8]: https://jcupitt.github.io/libvips/API/current/libvips-conversion.html#VipsBlendMode
```
lovell: libvips has moved to its own GitHub org so these links now start `https://libvips.github.io/libvips/...`

test/unit/overlay.js
```diff
@@ -595,4 +595,17 @@ describe('Overlays', function () {
         fixtures.assertSimilar(fixtures.expected('overlay-jpeg-with-jpeg.jpg'), data, done);
       });
   });
+
+  it('Overlays PNG on a solid background using difference blend mode', function (done) {
```
lovell: It's always a pleasure to see tests in a PR, thank you. Please can you also add tests for the error handling that's been introduced around invalid `blendMode` values.

lib/composite.js
```diff
@@ -2,6 +2,26 @@
 
 const is = require('./is');
 
+/**
+ * Blend modes when compositing images
+ * @member
+ * @private
+ */
+const blendMode = {
```
lovell: Please can these match libvips' VipsBlendMode [nicknames](https://github.com/libvips/libvips/blob/7da31386c73cacc17116529463529014f8525cc1/libvips/iofuncs/enumtypes.c#L385-L409) (see another of my comments below about how this might be achieved).

alewin:
@tcz this is a great feature +1

lovell:
Thank you for submitting this PR @tcz. You'll be pleased to know the `composite` operation, including multiple images and blend modes, will be in v0.22.0. See https://github.com/lovell/sharp/issues/728#issuecomment-471229437 for more details.

PR #1463, opened 2018-11-18
Docs: add installation details for Claudia.js

aliustaoglu:
I could not find any documents for [Claudia JS](https://claudiajs.com). It's very similar to AWS Lambda deployment since it's actually also a Lambda deployment. Took me a while to set it up and do the deployment inside the docker.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -172,6 +172,44 @@ docker run -v "$PWD":/var/task lambci/lambda:build-nodejs8.10 npm install sharp
 To get the best performance select the largest memory available.
 A 1536 MB function provides ~12x more CPU time than a 128 MB function.
 
+### Claudia.js
+
+Similar to AWS Lambda but Claudia deployment has to be done inside the docker.
+Therefore, AWS credentials also need to be passed into docker container.
+This assumes your credentials are stored in standard `$HOME/.aws/credentials` file.
+
+Create a bash script in the same folder as package.json. eg. claudia.sh. 
+Run `chmod +x` for this file and update the content as below:
+
+```sh
+docker run -it -v $PWD:/claudia -v $HOME/.aws:/root/.aws --rm lambci/lambda:build-nodejs8.10 /bin/bash -c "\
+cd /claudia
+rm -rf node_modules
+npm install -g claudia
+npm run create-function
+"
+```
+
+Add "create-function" and "deploy-claudia" scripts inside package.json.
+
+```json
+{
+  "scripts" : {
+    "create-function": "claudia create --no-optional-dependencies --name my-claudia --handler index.handler",
+    "deploy-claudia": "./claudia.sh"
+  }
+}
+```
+
+Do NOT run "create-function" inside your local machine, just run "deploy-claudia" inside the local machine.
+
+```sh
+npm run deploy-claudia
+```
+
+This will create all the artifacts inside the container. Zip file will also be created and then deployed from the container.
+You will need to create another sh file and script for the "claudia update" command in a similar way.
+
 ### NW.js
 
 Run the `nw-gyp` tool after installation.


aliustaoglu:
This initialisation basically prevents any issues like `module initialization error` when working with Claudia.js. With Claudia, you cannot create your own zip file and upload it. Everything is done by Claudia.JS therefore all has to be done inside the docker.

lovell:
Hello, thanks for this PR for the docs.

I haven't used claudia but notice it can control `npm install` via its `---npm-options` flag. Do you think the following might work instead?

```sh
claudia create --npm-options="--arch=x64 --platform=linux --target=8.10.0" ...
```

aliustaoglu:
Hi, I did not know about this npm-options parameter. Not sure if works but will give it a try when I get home. Docker method definitely works but have to admit that it's pretty low level and could be hard to setup. And it's slower.

I'll let you know about the results later.

aliustaoglu:
I get an error `'linux-x64' binaries cannot be used on the 'darwin-x64'` and it seems that is the main reason for Docker.

Full error message is as below:

```
$ claudia update --no-optional-dependencies --profile claudia --npm-options="--arch=x64 --platform=linux --target=8.10.0"
packaging files	npm install -q --no-audit --production --arch=x64 --platform=linux --target=8.10.0

> sharp@0.21.0 install /private/var/folders/x_/18qnbsm16r94s68crds61jxm0000gp/T/cmkeZ5/s3resizer-1.0.0-BPs4dR/package/node_modules/sharp
> (node install/libvips && node install/dll-copy && prebuild-install) || (node-gyp rebuild && node install/dll-copy)

info sharp Downloading https://github.com/lovell/sharp-libvips/releases/download/v8.7.0/libvips-8.7.0-linux-x64.tar.gz
validating package
Error: 'linux-x64' binaries cannot be used on the 'darwin-x64' platform. Please remove the 'node_modules/sharp/vendor' directory and run 'npm install'.
    at Object.hasVendoredLibvips (/private/var/folders/x_/18qnbsm16r94s68crds61jxm0000gp/T/cmkeZ5/s3resizer-1.0.0-BPs4dR/package/node_modules/sharp/lib/libvips.js:61:13)
    at Object.<anonymous> (/private/var/folders/x_/18qnbsm16r94s68crds61jxm0000gp/T/cmkeZ5/s3resizer-1.0.0-BPs4dR/package/node_modules/sharp/lib/constructor.js:9:22)
    at Module._compile (module.js:635:30)
    at Object.Module._extensions..js (module.js:646:10)
    at Module.load (module.js:554:32)
    at tryModuleLoad (module.js:497:12)
    at Function.Module._load (module.js:489:3)
    at Module.require (module.js:579:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (/private/var/folders/x_/18qnbsm16r94s68crds61jxm0000gp/T/cmkeZ5/s3resizer-1.0.0-BPs4dR/package/node_modules/sharp/lib/index.js:3:15)
cannot require ./index after clean installation. Check your dependencies.
```

lovell:
Thanks for checking. The "validating package" message and subsequent error is from [validatePackage](https://github.com/claudiajs/claudia/blob/38fb86a8d2e72a50121a3130917043a03576a68f/src/tasks/validate-package.js#L15-L20) inside `claudia`, which appears to always `require` your code and its dependencies using the build platform rather than the target platform.

The advice in the Claudia docs at https://claudiajs.com/tutorials/packaging.html#local-modules is to  lazy-load native modules such as sharp:

> "For example, if you use binaries for a different operating system, lazy-load those modules when you need them, so they do not prevent package validation."

Perhaps `claudia` could be modified to reduce its `validatePackage` check from an error to a warning when `--npm-options` includes a `--platform=linux` override? That feels like a more user-friendly approach than pushing documentation onto all the native modules that might be used with it.

PR #1475, opened 2018-11-23
Do not rely on function names to populate Sharp's prototype so that the library can be minified. Fixes #1474.

### LINKED ISSUE #1474: The way sharp attaches methods to its prototype forbids minification
Hi there, @lovell :)

I'll take the example of `lib/input.js`:

```js
  [
    // Private
    _createInputDescriptor,
    _write,
    _flattenBufferIn,
    _isStreamInput,
    // Public
    clone,
    metadata,
    stats,
    limitInputPixels,
    sequentialRead
  ].forEach(function (f) {
    Sharp.prototype[f.name] = f;
  });
```

If the code is minified, all those functions have mangled names and those mangled names are used to attach the methods onto the prototype (`f.name`).

AFAIK, Node 4.5 (the minimal requirement for sharp) supports shorthand properties so something like the following should work while being minification proof:

```js
Object.assign( Sharp.prototype,  {
    // Private
    _createInputDescriptor,
    _write,
    _flattenBufferIn,
    _isStreamInput,
    // Public
    clone,
    metadata,
    stats,
    limitInputPixels,
    sequentialRead
  });
```

For the context: I bundle `sharp` and a lot more modules using webpack and I'd like to reduce the size of the file which the above-mentioned pattern makes impossible.

jaubourg:
That should solve the issue. All unit tests are passing on my machine. I wasn't able to control if there were no other issues in the code regarding minification compatibility since `msbuild` is being particularly difficult when I reference `sharp` as a `file:` dependency.

### DIFF
--- a/lib/channel.js
+++ b/lib/channel.js
@@ -117,14 +117,12 @@ function bandbool (boolOp) {
  * @private
  */
 module.exports = function (Sharp) {
-  // Public instance functions
-  [
+  Object.assign(Sharp.prototype, {
+    // Public instance functions
     removeAlpha,
     extractChannel,
     joinChannel,
     bandbool
-  ].forEach(function (f) {
-    Sharp.prototype[f.name] = f;
   });
   // Class attributes
   Sharp.bool = bool;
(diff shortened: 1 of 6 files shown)


jaubourg:
Coverage decreasing... that's what happens when you remove lines of code :/

lovell:
Fantastique, merci beaucoup.

Issue #164, opened 2015-02-14
Recomb would be cool


richardkazuomiller:
Thanks for making this!

I'd like to use a recombination matrix (`vips_recomb`) to do some color space conversions (sepia and all that fun stuff). It goes beyond resizing, but would that be a good addition to this project? I haven't tried implementing it yet but it looks straightforward enough.

lovell:
Thanks for the suggestion. Taking advantage of more features of libvips will be significantly easier after #152.

richardkazuomiller:
That sounds pretty good to me. In the meantime would you take a PR implementing it using the old methods?

lovell:
Of course, that would be great, thank you. If you've not already seen it, there's a [contribution](https://github.com/lovell/sharp/blob/master/CONTRIBUTING.md) guide with a few more details.

Off the top of my head, things to consider include ensuring the matrix is freed from memory after use. The existing [flattening](https://github.com/lovell/sharp/blob/master/src/resize.cc#L363) logic provides an example of this. You may also need to `vips_cast` the resultant image back to integer from float/double.

You've probably already found a matrix suitable for a sepia effect, but if not, the "open source Instagram" that is GPUImage library provides [SepiaFilter](https://github.com/BradLarson/GPUImage/blob/master/framework/Source/GPUImageSepiaFilter.m).

Good luck!

lovell:
The following (hard-coded) matrix seems to work for a sepia-like effect:
```cc
VImage sepia = VImage::new_matrixv(
  3, 3,
  0.3588, 0.7044, 0.1368,
  0.2990, 0.5870, 0.1140,
  0.2392, 0.4696, 0.0912
);
```
I need to think more about how the JavaScript API to expose/support recombination matrices might look.

rnanwani:
Hi. This issue seems an old one. I can work on this if you can suggest a relevant JS API.
I was thinking of exposing this as something like

```
sharp().recombine("sepia").toBuffer()
```

The `recombine` can also be termed as `effect` and we can support only specific named values instead of complete recombination matrices.

lovell:
Thanks Rahul, mirroring libvips' naming would suggest `recomb` is best for this and the API should at least support 2d matrices via an Array of Arrays, e.g.
```javascript
.recomb([[0.3588, 0.7044, 0.1368], [0.2990, 0.5870, 0.1140], [0.2392, 0.4696, 0.0912]])
```

[...]

PR #1477, opened 2018-11-26
Add in the Recomb function.

fromkeith:
Adding in a basci "Recomb" #164


lovell:
Hello, thanks for this, I'll be able to take a closer look in the next day or so.

In the meantime it might be worth adding a test for an PNG image with an alpha transparency e.g. `inputPngWithTransparency` to ensure this works also.

fromkeith:
Sure!

I think the Alpha channel may cause some issues, since the current code is hardcoded for a 3x3 matrix. It doesn't look like 'channels' is necessarily set until pipeline.cc. So I can either:

* Set the identity for Alpha, and force a 3x3 matrix input.
* Or Allow 4x4 matrices, but throw away the extra col/row if no alpha channel is found in pipeline.cc

I'll start with the first, but it shouldn't be too much work to go the route of the second if you prefer.

lovell:
Thanks, option 1 sounds like the right approach.

lovell:
I've merged this in commit 541e710 - the RGBA example wasn't quite right so I took the liberty of simplifying that logic. Thanks again for this PR - it will be included in the next release, probably v0.21.1.

PR #1483, opened 2018-11-30
Add pyramid and tiling output options for TIFF images

mbklein:
This change adds the libvips `tiffsave` `pyramid`, `tile`, `tile_width`, and `tile_height` options to sharp's `tiff()` output format function to allow for the creation of multires/deep-zoom TIFFs.

I wrestled with the idea of trying to overload the `tile()` function to handle `tiff` as another container type, but I stuck with the straightforward “do it more or less the way libvips does it” approach.


lovell:
Fantastic, thank you Michael, this looks like the right approach and with lots of tests too. I should have some time to review this properly early next week.

mbklein:
Thanks! I just realized that I left the trailing `>` off my email address, so I force-pushed it in. All ready to go now.

lib/output.js
```diff
@@ -325,6 +329,36 @@ function tiff (options) {
       throw new Error('Invalid Value for squash ' + options.squash + ' Only Boolean Values allowed for options.squash.');
     }
   }
+  // tiling
+  if (is.object(options) && is.defined(options.tile)) {
```
lovell: This might be the right time to merge all the logic that depends on the `if (is.object(options)) { ... }` statement being true. If nothing else it should reduce the complexity of this function.

mbklein: Merged all that logic together, squashed my commits, and rebased on master.

lovell:
Marvellous, thank you for this Michael, it will be included in the next release.

PR #1526, opened 2018-12-29
update deprecated examples

PascalTemel:


lovell:
Good spot, thank you for submitting this Pascal!

PascalTemel:
Np, thanks for the great library.

PR #1547, opened 2019-01-15
Update `overlay` as optionnal parameter

Maxou44:
Fix the documentation to inform `overlay` parameter is optionnal.


lovell:
I added the corresponding JSDoc change in commit 98e9078 - thanks again for spotting this.

PR #1588, opened 2019-02-22
Add missing comma to `extend` example

crosscompile:
Came across this while copy-pasting the extend example.

### DIFF
--- a/docs/api-resize.md
+++ b/docs/api-resize.md
@@ -143,7 +143,7 @@ sharp(input)
     top: 10,
     bottom: 20,
     left: 10,
-    right: 10
+    right: 10,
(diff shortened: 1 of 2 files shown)


lovell:
Hello, thanks for spotting this.

The `docs/api-*.md` files are auto-generated from JSDoc in the source code. There's a message about this at the top of each auto-generated file.
```
```
Please can you update the relevant source code file, in this case https://github.com/lovell/sharp/blob/master/lib/resize.js#L278 then use `npm run docs` to update the documentation.

crosscompile:
Sorry about that, should have read more closely, updated!

lovell:
No worries, thanks for fixing this.

Issue #1451, opened 2018-11-08
De-float version selection in docs


Prinzhorn:
As mentioned in #1448 I'd like to de-float the version selector (it's in the bottom right).

However, I don't quite understand how the hosted readthedocs works. By default the version selector is indeed on the left, not on the right. See the official docs as an example: https://docs.readthedocs.io/en/latest/

![selection_371](https://user-images.githubusercontent.com/679144/48211531-f7a2f500-e379-11e8-851d-04bdf256d30d.png)
--
![selection_372](https://user-images.githubusercontent.com/679144/48211553-012c5d00-e37a-11e8-8c4e-2b6ceceb1ad2.png)

However, on the sharp docs there are several CSS files that don't exist on the official website which override the CSS and among other things set `left: auto` and `right: 20px`.

![selection_373](https://user-images.githubusercontent.com/679144/48211641-3173fb80-e37a-11e8-9a78-6c4d5e0abb95.png)

sharp.pixelplumbing.com is CNAMEd to readthedocs, so there's no custom reverse proxy injecting those things. Could you check the settings on readthedocs? Maybe it can be reconfigured? I don't think the correct solution is to override it once again with more [custom CSS](https://github.com/lovell/sharp/blob/1fa388370ecef47025c44ec19108f6445773933c/docs/css/extra.css).

lovell:
Thanks for digging into this. Beyond the [subnav improvement](https://github.com/lovell/sharp/blob/master/docs/css/extra.css) any extra CSS is defined somewhere inside the readthedocs theme processing. As you point out, their own site doesn't appear to have this so I'm pretty confused.

It might be time to reassess if readthedocs is the best place to host these given the lack of control over layout. I've briefly looked at a few other solutions and a combination of GitHub pages and [docusaurus](https://docusaurus.io/) (for versioning) might be the way to go.

Any other suggestions are very welcome.

lovell:
The replacement of readthedocs is being tracked at #1974.

PR #1595, opened 2019-03-01
Set page option for pdf

ramiel:
The page parameter is considered also when loading a pdf

### DIFF
--- a/docs/api-constructor.md
+++ b/docs/api-constructor.md
@@ -12,7 +12,7 @@
     -   `options.failOnError` **[Boolean][4]** by default halt processing and raise an error when loading invalid images.
          Set this flag to `false` if you'd rather apply a "best effort" to decode images, even if the data is corrupt or invalid. (optional, default `true`)
     -   `options.density` **[Number][5]** number representing the DPI for vector images. (optional, default `72`)
-    -   `options.page` **[Number][5]** page number to extract for multi-page input (GIF, TIFF) (optional, default `0`)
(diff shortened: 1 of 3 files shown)


docs/api-constructor.md
```diff
@@ -12,7 +12,7 @@
     -   `options.failOnError` **[Boolean][4]** by default halt processing and raise an error when loading invalid images.
          Set this flag to `false` if you'd rather apply a "best effort" to decode images, even if the data is corrupt or invalid. (optional, default `true`)
     -   `options.density` **[Number][5]** number representing the DPI for vector images. (optional, default `72`)
-    -   `options.page` **[Number][5]** page number to extract for multi-page input (GIF, TIFF) (optional, default `0`)
+    -   `options.page` **[Number][5]** page number to extract for multi-page input (GIF, TIFF, PDF) (optional, default `0`)
```
lovell: These are autogenerated from the source code by running `npm run docs`. Please can you update https://github.com/lovell/sharp/blob/master/lib/constructor.js#L67 too so we don't lose this change next time the docs are refreshed.

ramiel: Done. Maybe there's no reason to keep the generated code in the repository if it is generated...

lovell: Thanks for updating. The docs are currently hosted on readthedocs, which requires the markdown to be present in the repo, so maintaining JSDocs and automagically generating markdown is currently the lowest-maintenance approach. There's some discussion in #1451 to migrate away from this approach and/or improve things.

Issue #609, opened 2016-10-20
Brightness, Saturation and Hue


diegocsandrim:
Hi @lovell,
Does sharp support brightness, saturation and hue?
I was looking for something similar to `modulate` of [imagemagick](http://www.imagemagick.org/Usage/color_mods/#modulate) or [graphicsmagick](http://www.graphicsmagick.org/GraphicsMagick.html#details-modulate).
I took a look at the docs but could not find.

lovell:
Hello, *magick's "modulate" operator is approximately equivalent to a colourspace conversion to LCH (Luminance, Chroma, Hue) then a multiplication of pixels in the L and C channels. libvips supports these operations, but this is not currently exposed in sharp.

As always, I'm happy to accept a PR and help with discussions about how the API for this might work.

adityapatadia:
I would like to work on this. Can you point me to some resource which teaches about how can this be done?

lovell:
@adityapatadia Thanks for your interest.

This feature will involve a combination of converting to the LCH colourspace via [colourspace](http://www.vips.ecs.soton.ac.uk/supported/current/doc/html/libvips/libvips-colour.html#vips-colourspace), splitting out any channels to be "modulated", using [multiply](http://www.vips.ecs.soton.ac.uk/supported/current/doc/html/libvips/libvips-arithmetic.html#vips-multiply) to modify them with the relevant value before joining the channels back up again.

You probably won't need to worry about the colourspace conversion away from LCH as there's already logic to clean up at the end of the pipeline.

* L(ightness) ~= brightness
* C(hroma) ~= saturation
* H(ue)

The API could look something like:

```javascript
// identity transform, 100 means 100% of the original value so no change
.modulate({ brightness: 100, saturation: 100, hue: 100 }) 
```

```javascript
// reduce saturation by ~25%
.modulate({ saturation: 75 }) 
```

```javascript
// convert red to green etc.
.modulate({ hue: 166 }) 
```

adityapatadia:
I would be asking much more questions.
Apart form API, how do I implement in the C code? Which files should I add the operations? Can you show me sample from which I can learn?

lovell:
@adityapatadia Although it's marked `patch-welcome`, this is a relatively complex task. Perhaps look at implementing something like #577 first?

lovell:
@jonathan-kosgei Thank you for the offer of help.

sharp uses libvips C++ API
[...]

PR #1601, opened 2019-03-05
Implement brightness, saturation and hue modulation #609

Goues:
Hello, I've read about feature request #609 because this is what I would like to be able to do too. I have no experience in C/C++ but I followed all advice I could find to stitch those changes.

Can you please look at it and point me to all the missing pieces I should do as well?

Thank you very much!
Goues


Goues:
p.s. Unit tests are not passing, because it claims that output is not similar enough to expected output, should I just set a higher threshold?

lovell:
Thank you very much for this PR Jakub. It looks good from a quick scan and I expect to have some time to take a proper look at it next week.

It's always pleasure to see tests - perhaps try `assertMaxColourDistance` instead of `assertSimilar` as this feature is all about colours, plus that might avoid some of those threshold problems you mentioned.

There should probably also be some verification in the JavaScript to throw Errors for things like invalid property values (there are a few examples of this in other operations).

Goues:
I've updated to include input validation and error throwing. Also, I made it work with PNG with alpha channel, which wasn't the case in the first implementation.

lib/operation.js
```diff
@@ -415,6 +415,59 @@ function recomb (inputMatrix) {
...
+ *     hue: 90
+ *   });
+ *
+ * @param {Object} [options]
+ * @param {Number} [options.brightness] Brightness multiplier
+ * @param {Number} [options.saturation] Saturation multiplier
+ * @param {Number} [options.hue] Degrees for hue rotation
+ * @returns {Sharp}
+ */
+function modulate (options) {
```
lovell: We should probably verify `options` is an object.

lib/operation.js
```diff
@@ -415,6 +415,59 @@ function recomb (inputMatrix) {
...
+      this.options.brightness = options.brightness;
+    } else {
+      throw new Error('Invalid brightness ' + options.brightness);
+    }
+  }
+  if ('saturation' in options) {
+    if (is.number(options.saturation) && options.saturation >= 0) {
+      this.options.saturation = options.saturation;
+    } else {
+      throw new Error('Invalid saturation ' + options.brightness);
```
lovell: I think this error message should refer to `options.saturation` (ditto for the hue message below). There's an `is.invalidParameterError` helper function that might be useful here too.

src/operations.cc
```diff
@@ -309,6 +309,20 @@ namespace sharp {
           0.0, 0.0, 0.0, 1.0));
   }
 
+  VImage Modulate(VImage image, double const brightness, double const saturation, double const hue) {
+    image = image.colourspace(VIPS_INTERPRETATION_LCH);
```
lovell: It might be slightly more efficient to perform the colourspace conversion after the removal of alpha channel.

src/operations.cc
```diff
@@ -309,6 +309,20 @@ namespace sharp {
...
+
+    if (HasAlpha(image)) {
+      // Separate alpha channel
+      VImage alpha = image[image.bands() - 1];
+      return RemoveAlpha(image)
+        .linear({brightness, saturation, 1}, {0, 0, hue})
```
lovell: I've just realised the hue addition could result in angle value greater than 360 but I think we're OK as libvips will always convert away from LCH via LAB using sin/cos so this shouldn't be a problem, phew.

Goues: Testing with values like 350 shows that indeed, libvips handles overflow well. Otherwise I would have no idea how to do that manually.

Goues:
Thank you for the review, I've updated the PR and rebased over master.

lovell:
Thanks for all the updates, this is looking great. Were you able to experiment with using PNG input (and output) for the tests to avoid having to set a threshold?

Goues:
Yes, I added this file for testing hue rotation https://github.com/lovell/sharp/blob/bed50294bd59fb5f1a22c1edbd98414c85f733c6/test/fixtures/test-pattern.png with hue from 30 to 360 by steps of 30degs. It came exactly as css `filter: hue-rotate(30deg)` does, because I was testing all those three transforms against css filters.

Also I had this branch in my app for about a week already and QA didn't find anything wrong with uploaded images.

lib/operation.js
```diff
@@ -415,6 +415,62 @@ function recomb (inputMatrix) {
...
+      this.options.saturation = options.saturation;
+    } else {
+      throw is.invalidParameterError('saturation', 'number above zero', options.saturation);
+    }
+  }
+  if ('hue' in options) {
+    if (is.number(options.hue)) {
+      this.options.hue = options.hue % 360;
+    } else {
+      throw is.invalidParameterError('saturation', 'number', options.saturation);
```
lovell: The validation error for `hue` incorrectly refers to `saturation`.

lib/operation.js
```diff
@@ -415,6 +415,62 @@ function recomb (inputMatrix) {
...
+  }
+  if ('saturation' in options) {
+    if (is.number(options.saturation) && options.saturation >= 0) {
+      this.options.saturation = options.saturation;
+    } else {
+      throw is.invalidParameterError('saturation', 'number above zero', options.saturation);
+    }
+  }
+  if ('hue' in options) {
+    if (is.number(options.hue)) {
```
lovell: I think `hue` needs to be an integer value (can use `is.integer` check for this) as it is additive.

src/pipeline.h
```diff
@@ -87,6 +87,9 @@ struct PipelineBaton {
   std::vector<double> flattenBackground;
   bool negate;
   double blurSigma;
+  double brightness;
+  double saturation;
+  double hue;
```
lovell: I think `hue` can be a `int` (signed, as negative values are acceptable).

Goues: I actally started out with hue as an integer, but I don't know how to make it work in C type system. `linear` expects vectors with doubles, so it fails with: `error: non-constant-expression cannot be narrowed from type 'int' to 'double' in initializer list [-Wc++11-narrowing]`

Should I somehow convert int back to double in operations.cc?

lovell: Try something like `{0.0, 0.0, static_cast<double>(hue)}` in the call to `linear()`, which the compiler should be able to use to initialise a `std::vector<double>`.

lovell:
Fantastic, thanks again for this Jakub, it will be in v0.22.1.

PR #1623, opened 2019-03-24
Change docs for `composite` to reflect how `create` works

Goues:
I found an error in docs for `composite` method. The `create` object should be documented under `images[].input` not inside `images[]`. No changes to the code itself.


lovell:
Good spot and thank you for fixing it. This was missed in the migration from the now-deprecated `overlayWith`.

Issue #1624, opened 2019-03-25
Bailing early if bindings are broken


sidharthachatterjee:
Hello!

We've seen some instances of `sharp` failing because of a broken installation or env issue (attempt at documenting them is in https://github.com/gatsbyjs/gatsby/issues/10841). 

To make debugging easier for the user, we added a try catch and friendly error in `gatsby-plugin-sharp` (https://github.com/gatsbyjs/gatsby/pull/10677)

What do you think about adding this around the require here?
https://github.com/lovell/sharp/blob/fc178de309d8e4055e86ffe1bbcfcf40754c885b/lib/constructor.js#L10

In case it helps, the error message we show if requiring `sharp` fails is 

```
The dependency "sharp" does not seem to have been built or installed correctly.
      - Try to reinstall packages and look for errors during installation
      - Consult "sharp" installation page at http://sharp.pixelplumbing.com/en/stable/install/
```

Happy to open a PR if you think this is useful. Thank you for all the great work!

lovell:
Hi Sidhartha, this is a great idea, thank you. I'm always happy to accept PRs that make error messages easier to deal with.

sidharthachatterjee:
Thank you so much, Lovell! Opened a PR adding this 🙂

PR #1638, opened 2019-04-02
Add warning if sharp bindings aren't built correctly

sidharthachatterjee:
Added a friendly error message in case the require to the bindings throws for any reason. The error message looks like this:

```
"sharp" does not seem to have been built or installed correctly.

- Try to reinstall packages and look for errors during installation
- Consult "sharp" installation page at http://sharp.pixelplumbing.com/en/stable/install/

If neither of the above work, please open an issue in https://github.com/lovell/sharp/issues
```

Fixes https://github.com/lovell/sharp/issues/1624

### DIFF
--- a/lib/constructor.js
+++ b/lib/constructor.js
@@ -7,7 +7,26 @@ const events = require('events');
 const is = require('./is');
 
 require('./libvips').hasVendoredLibvips();
-const sharp = require('bindings')('sharp.node');
+
+let sharp;
+
+try {
+  sharp = require('bindings')('sharp.node');
+} catch (error) {
+  // Bail early if bindings aren't available
+  console.error(
+    `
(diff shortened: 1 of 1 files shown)


sidharthachatterjee:
Tests passed but it seems like the Travis job on Linux (musl) - Node 8 failed with 
```
Segmentation fault (core dumped)
npm ERR! Test failed.  See above for more details.
The command "sudo docker exec sharp sh -c "npm test"" exited with 1.

```

lovell:
Thanks Sidhartha, this is a great improvement.

lovell:
I've added some more specific help in commit 687795c based on common error messages. Further additions/improvements are always welcome.

(I've also added commit 90a0382 to hopefully prevent/reduce the occurrence of the unrelated test failure you experienced.)

PR #1687, opened 2019-05-03
add skipBlanks support for dz and zoomify layout

RaboliotTheGrey:
Hi!

Here it is ! 

* I didn't upgrade the libvips in the package.json version
* It seems that upgrade libvips to 8.8.0 broke 2 tests (`Image metadata ->  JPEG with IPTC/XMP` and `Attention strategy -> JPEG`) It also appears with master when libvips 8.8.0 is activated. I didn't fix them because i don't underestand their purposes ...
* I had to force the the skipBlanks in output.js because `AttrTo<int32_t>(options, "skipBlanks")` return 0 if it's NULL/undefined

Let me know if it needs some modifications.

### DIFF
--- a/lib/output.js
+++ b/lib/output.js
@@ -539,6 +539,7 @@ function toFormat (format, options) {
  * @param {Number} [tile.overlap=0] tile overlap in pixels, a value between 0 and 8192.
  * @param {Number} [tile.angle=0] tile angle of rotation, must be a multiple of 90.
  * @param {String} [tile.depth] how deep to make the pyramid, possible values are `onepixel`, `onetile` or `one`, default based on layout.
+ * @param {Number} [tile.skipBlanks=-1] threshold to skip tile generation, a value 0 - 255 for 8-bit images or 0 - 65535 for 16-bit images
  * @param {String} [tile.container='fs'] tile container, with value `fs` (filesystem) or `zip` (compressed file).
  * @param {String} [tile.layout='dz'] filesystem layout, possible values are `dz`, `zoomify` or `google`.
  * @returns {Sharp}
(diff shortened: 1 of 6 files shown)


lovell:
Bonjour Jordan, merci beaucoup for this PR, which from a quick scan looks good - thank you for adding tests!

I notice there's some logic in libvips that defaults `skip_blanks` to a value of 5 when using the "google" layout, which this change should probably mirror.

https://github.com/libvips/libvips/blob/bcdaeca57877ba1d8c3cc7659708e4ef6f396a0b/libvips/foreign/dzsave.c#L1832-L1836

This PR will have to wait until after libvips v8.8.0 is released, which sharp will probably pick up in v0.23.0. I'll fix up the tests at that point to reflect the new/changed features that are making them fail.

RaboliotTheGrey:
Hi, I've fixes this part and added a test to the google layout. 

let me know if you see any improvement :)

lovell:
@RaboliotTheGrey I've changed the destination branch for this PR to the `vision` branch, which will become v0.23.0. Are you able to rebase against the `vision` branch and push to allow the CI tests to pass?

RaboliotTheGrey:
Thanks!
I wont be able to rebase it until thuesday. :)
Have a nice week-end

-------- Original Message --------

RaboliotTheGrey:
Hi,

I've rebase the branch on the lastest version of vision and test agains the version 8.8.1 of libvips.
All the tests are green on local.

lib/output.js
```diff
@@ -539,6 +539,7 @@ function toFormat (format, options) {
  * @param {Number} [tile.overlap=0] tile overlap in pixels, a value between 0 and 8192.
  * @param {Number} [tile.angle=0] tile angle of rotation, must be a multiple of 90.
  * @param {String} [tile.depth] how deep to make the pyramid, possible values are `onepixel`, `onetile` or `one`, default based on layout.
+ * @param {Number} [tile.skipBlank=-1] threshold to skip tile generation, a value 0 - 255 for 8-bit images or 0 - 65535 for 16-bit images
```
lovell: I think `tile.skipBlank` should be the plural form `tile.skipBlanks`.

Please can you run `npm run docs` after fixing this, which should automagically update `docs/api-output.md`.

RaboliotTheGrey: Nice catch :)
Sorry I took a bit long to fix this.

lovell:
Merveilleux, merci beaucoup.

Issue #903, opened 2017-08-04
Unexpected image rotation during resize


mcurcio:
My generated images are always "horizontal" (larger in width) even when the input was "vertical" -- for instance, an upright phone picture.

My resize code is basically:
```js
sharp(originalImagePath)
    .resize(200, 200) // to make a small thumbnail
    .min()
    .toFile(outputImagePath);
```

mcurcio:
My mistake! I just found the documentation for `rotate` and I now understand that my source phone image was actually wider than it was tall but has an EXIF tag to rotate.

For anyone else who may have this issue, the correct code is:
```js
sharp(originalImagePath)
    .rotate()
    .resize(200, 200) // to make a small thumbnail
    .min()
    .toFile(outputImagePath);
```

alxvallejo:
How can you tell if the image has "an EXIF tag to rotate"? Your solution suggests you just rotate everything ... ?

lovell:
@alxvallejo If you've not already done so, please see http://sharp.dimens.io/en/stable/api-operation/#rotate

> "If no angle is provided, it is determined from the EXIF data."

If there's no EXIF data, or if the EXIF data states there should be no rotation, then none will occur for a parameter-less call to `rotate()`.

MentalGear:
@lovell Tightly & very well explained. You might want to add this phrase to the docs, because to be honest, I didn't get this just from the doc text.
"If there's no EXIF data, or if the EXIF data states there should be no rotation, then none will occur for a parameter-less call to rotate()."

lovell:
@MentalGear Please feel free to submit a PR for the [JSDocs](https://github.com/lovell/sharp/blob/master/lib/operation.js#L12) with this improvement.

demisx:
I believe the latest doc link is: https://sharp.pixelplumbing.com/en/stable/api-operation/#rotate

avichal-neweradeveloper:
@lovell rotate() works but it is increasing my buffer size a few times

PR #1748, opened 2019-06-10
Docs: Revise resize API

keksipurkki:
Add the rotate() call to the code examples, see https://github.com/lovell/sharp/issues/903. 

Disclaimer: I bumped into the aforementioned issue. This PR is meant to prevent this from happening in the future.

### DIFF
--- a/docs/api-resize.md
+++ b/docs/api-resize.md
@@ -53,18 +53,10 @@ Possible interpolation kernels are:
 ```javascript
 sharp(input)
   .resize({ width: 100 })
+  .rotate() // Ensure orientation is preserved
   .toBuffer()
   .then(data => {
     // 100 pixels wide, auto-scaled height
-  });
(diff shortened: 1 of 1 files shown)


lovell:
Hello, thanks for the PR. The `docs/api-*.md` files are autogenerated from source files in the `lib/` directory. You'll see there's a message about this at the top of the file you edited:
```
```

https://github.com/lovell/sharp/blob/master/.github/CONTRIBUTING.md#documentation

Did you see the existing docs for rotate? https://sharp.pixelplumbing.com/en/stable/api-operation/#rotate

I'm not sure that including rotate is required for all the resize examples, but can see the value in including it in one of them as it's quite a common use case.

lovell:
Closing due to inactivity but please do feel free to reopen with these additions to the jsdocs in the source.

PR #1832, opened 2019-08-11
change reguler function expression to arrow function

nocord:


lovell:
Hi, thank you for this PR. There's no description - are you able to explain the benefit of using arrow functions in this specific file/context? There are parts of sharp that use arrow functions where there is a need to refer to `this` from the outer scope, but I don't see such a benefit with this change.

nocord:
it can reduce bundle size

lovell:
This is a server-side only module, so "bundle size" is of a much lower concern than the provision of named functions/variables that aid debugging (where the use of anonymous arrow functions can make debugging harder).

nocord:
I didn't changed any named function

Issue #1824, opened 2019-08-05
"heif" is missing from the list checked in toFormat


jaubourg:
https://github.com/lovell/sharp/blob/master/lib/output.js#L509

I can workaround by using `.heif()` directly (what `.toFormat()` actually does).

lovell:
A PR to add this would be very welcome if you're able.

jaubourg:
@lovell https://github.com/lovell/sharp/pull/1834

lovell:
v0.23.1 now available, merci beaucoup for the PR.

PR #1834, opened 2019-08-12
Have toFormat accept heic/heif (#1824)

jaubourg:
I used a Map instance to store format correspondances. Seemed to make eveything a bit clearer.


lovell:
Magnifique, merci beaucoup.

Issue #1830, opened 2019-08-09
SVG feBlend or mix-blend-mode support


Andargor:
**What are you trying to achieve?**

sharp 0.23.0

Render an SVG file that either uses feBlend for two images in color-burn mode or the mix-blend-mode style with patterns. It renders fine in Chrome but sharp seems to either not render the second image (with the filter) or with no blending (with the patterns and CSS).

**Have you searched for similar questions?**

Extensively, I even looked at the libvips issues, I thought #804 would be related, but that was just a syntax error. Or I may be just blind.

**Are you able to provide a standalone code sample that demonstrates this question?**

Using this simple code:

```js
sharp("test.svg")
	.jpeg({
		quality: 80
	})
	.toFile("output/test.jpg")
	.catch(err => {
		console.error(err);
	});
```

There are two versions of the SVG file, the first with filters:

```xml
<?xml version="1.0" encoding="utf-8"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" preserveAspectRatio="xMidYMid" width="822" height="1122" viewBox="0 0 822 1122" color-interpolation-filters="sRGB">
	<defs>
		<filter id="image-front-logo-burn">
			<feImage xlink:href="img/logo-RUN.png" x="0" y="0" width="822" height="1122" result="logo"/>
			<feImage xlink:href="img/old-paper-dark.png" x="0" y="0" width="822" height="1122" result="paper"/>
			<feBlend mode="color-burn" in="logo" in2="paper"/>
		</filter>
	</defs>
	<g id="paper">
		<rect x="0" y="0" width="822" height="1122" filter="url(#image-front-logo-burn)"/>
	</g>
</svg>
```

The second version involves the use of patterns and CSS:

```xml
<?xml version="1.0" encoding="utf-8"?>
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" preserveAspectRatio="xMidYMid" width="822" height="1122" viewBox="0 0 822 1122" color-interpolation-filters="sRGB">
    <style>
		.color-burn {mix-blend-mode: color-burn;}
    </style>
	<defs>
		<pattern id="image-front-paper" patternUnits="userSpaceOnUse" x="0" y="0" width="822" height="1122">
			<image xlink:href="img/old-paper-dark.png" x="0" y="0" width="822" height="1122"/>
		</pattern>
[...]

PR #1835, opened 2019-08-13
Add premultiplied option to composite

Andargor:
Currently, sharp automatically premultiplies during composite. However, this may cause issues if the image below is already premultiplied. This additional `premultiplied` option for composite emulates the VIPS behavior of the `--premultiplied` flag.

If true, premultiplication is not performed by sharp. If false or absent, the current existing behavior is performed.

See #1830 and @jcupitt comments (specifically https://github.com/lovell/sharp/issues/1830#issuecomment-519985352 )


lovell:
Thanks for the PR. Are you able to add a test for this? Perhaps perform the composite with and without this setting and verify via the `assertMaxColourDistance` test utility?

Andargor:
> Thanks for the PR. Are you able to add a test for this? Perhaps perform the composite with and without this setting and verify via the `assertMaxColourDistance` test utility?

Hopefully there's an example in one of your existing test cases I can base myself on to respect your guidelines, if any?

lovell:
The existing composite "blend" test provides an example of using `assertMaxColourDistance`.

https://github.com/lovell/sharp/blob/3fa91bb4ce2ee561fd33b74666c3870ce24e7d97/test/unit/composite.js#L48-L63

Andargor:
@lovell Hopefully I didn't mess up too badly...

lib/composite.js
```diff
@@ -147,6 +148,14 @@ function composite (images) {
         throw is.invalidParameterError('gravity', 'valid gravity', image.gravity);
       }
     }
+    if (is.defined(image.premultiplied)) {
+      if (is.bool(image.premultiplied)) {
+        composite.premultiplied = image.premultiplied;
+      } else {
+        throw is.invalidParameterError('premultiplied', 'boolean', image.premultiplied);
```
lovell: Please can you add a unit test to cover the line that throws an Error should `image.premultiplied` be defined but not a boolean.

Andargor: Hopefully this is consistent with the test for "invalid tile"

lovell:
Fantastic, thank you very much for this, it will be in v0.23.1.

Issue #1855, opened 2019-08-31
toBuffer() without parameter returns ambiguous result (buffer/object) if toBuffer() is used previously


MihalySzabo:
If I use `toBuffer()` without any parameters, it returns a promise with a buffer parameter. That is the expected behaviour. If I use it with `{ resolveWithObject: true }` parameter, then it returns a promise with an object parameter, again, that is the expected behaviour. But if I use it with the parameter and then use it without any parameter, the second call also returns a promise with an object, instead of a plain buffer.

What is the output of running `npx envinfo --binaries --languages --system --utilities`?

  ```
System:
    OS: macOS 10.14.6
    CPU: (4) x64 Intel(R) Core(TM) i5-4278U CPU @ 2.60GHz
    Memory: 94.21 MB / 8.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 8.10.0 - ~/.nvm/versions/node/v8.10.0/bin/node
    Yarn: 1.16.0 - /usr/local/bin/yarn
    npm: 5.6.0 - ~/.nvm/versions/node/v8.10.0/bin/npm
  Utilities:
    Make: 3.81 - /usr/bin/make
    GCC: 10.14. - /usr/bin/gcc
    Git: 2.20.1 - /usr/bin/git
  Languages:
    Bash: 3.2.57 - /bin/bash
    Perl: 5.18.4 - /usr/bin/perl
    PHP: 7.1.23 - /usr/bin/php
    Python: 2.7.10 - /usr/bin/python
    Ruby: 2.3.7 - /usr/bin/ruby
```
What are the steps to reproduce?

Load an image
Use `image.toBuffer({resolveWithObject: true}).then(function (result) {...})`
Then use `image.toBuffer().then(function (result) {...})` (without parameter) again.

What is the expected behaviour?

Second result should be a buffer instead of an object, because second `toBuffer()` was called without any parameter.

Are you able to provide a standalone code sample, without other dependencies, that demonstrates this problem?

To get it work, a JPEG image called `image.jpg` have to be in the same directory.

```
const Sharp = require('sharp');
const fs = require('fs');

var image = Sharp(fs.readFileSync('./image.jpg'));
return image.png().toBuffer({
	resolveWithObject: true
}).then(function () {
	return image.png().toBuffer().then(function (output) {
		console.log(output);
	});
});
```

lovell:
[...]

PR #1860, opened 2019-09-05
reusing same sharp instance and differente options to toBuffer do not…

RaboliotTheGrey:
Reusing same sharp instance and differente options to toBuffer do not produce the same output.

Resolves this issue : https://github.com/lovell/sharp/issues/1855

### DIFF
--- a/lib/output.js
+++ b/lib/output.js
@@ -92,6 +92,8 @@ function toFile (fileOut, callback) {
 function toBuffer (options, callback) {
   if (is.object(options)) {
     this._setBooleanOption('resolveWithObject', options.resolveWithObject);
+  } else if (this.options.resolveWithObject) {
+    this.options.resolveWithObject = false;
   }
   return this._pipeline(is.fn(options) ? options : callback);
 }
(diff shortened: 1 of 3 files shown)


test/unit/toBuffer.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+// const fs = require('fs');
```
lovell: We can remove this commented-out import.

lib/output.js
```diff
@@ -92,6 +92,8 @@ function toFile (fileOut, callback) {
 function toBuffer (options, callback) {
   if (is.object(options)) {
     this._setBooleanOption('resolveWithObject', options.resolveWithObject);
+  } else if (this.options.resolveWithObject) {
+    this._setBooleanOption('resolveWithObject', false);
```
lovell: Can we get away with just setting `this.options.resolveWithObject = false` in the else clause?

test/unit/toBuffer.js
RaboliotTheGrey: Yup, it's done. A leftover of the duplication of the file :)

lib/output.js
RaboliotTheGrey: I kept the save structure than the previous statement but it's more to check the input than a setter. Also change to a simple assignment.

lovell:
Merveilleux!

Issue #1597, opened 2019-03-01
Trim functionality fails with error: Unexpected error while trimming. Try to lower the tolerance


Granga:
In the latest version 0.21.3, trimming a png image fails.
In a previous version 0.20.8, the code completes successfully.

Here's my function in typescript:
```typescript
async trim(dataUrl: string) {
    dataUrl = dataUrl.replace(/^data:image\/png;base64,/, "");
    dataUrl += dataUrl.replace('+', ' ');

    let buffer = new Buffer(dataUrl, "base64");

    return sharp(buffer)
        .png()
        .trim(1) //also tried default parameter
        .toBuffer();
}
```
This is the output in console: 
`Error: Unexpected error while trimming. Try to lower the tolerance`.

lovell:
Hello, the following entry for v0.21.0 in the changelog is relevant here:

> Switch from custom trim operation to vips_find_trim. #914
> https://sharp.pixelplumbing.com/en/stable/changelog/#v0210-4th-october-2018

Please can you provide a sample image that behaves in this manner.

Granga:
Thank you for your response @lovell.
I generate the image's `dataUrl` with fabricjs inside puppeteer, so it's basically `canvas.toDataUrl()`.

Both 'jpeg' and 'png' produce this error. Here's the png that won't trim:
![5c79acd0d45450aa28dd8c8d](https://user-images.githubusercontent.com/6062311/53673223-8b306680-3c86-11e9-8afe-2127df688f97.png)

Granga:
My node version is v10.15.1 on Windows10.

I did another test with the uploaded png:
Opened it with mspaint and just saved it in a new file. 
It lost the transparent background, but sharp v0.21.3 trimmed it **successfully**.

papandreou:
Then we just need a way to automate mspaint :)

lovell:
@Granga Thank you for the example PNG. I think this has something to do with the input being 2 channel greyscale+alpha but I'll need to take a closer look.

lovell:
All of the pixels in this image are black and the text is contained only in the transparency layer.

By default sharp flattens to black hence the entire image is background.

We need to retrieve the background colour from the bKGD chunk in the PNG, if present, and flatten to that instead.

kapouer:
Hi,
using sharp 0.22.1 i get the same error when trying to trim() an empty image. Since resize doesn't crash on those, it would be nice if trim did not crash too.

stefanak-michal:
[...]

PR #1879, opened 2019-09-20
Improve trimming of images with transparency

muse-monotype:
Retrieve the background colour alpha value and pass image alpha band to find_trim() if background colour is not fully opaque. Attempt to fix issue #1597


lovell:
Thanks for the PR. Am I correct in thinking that if the top left pixel has any transparency then only the alpha channel is used for the trim? I'm not sure this is the desired behaviour, or at least could break backwards compatibility.

Perhaps a two-step process could be used instead, where if the initial call to `find_trim` returns zero offsets and the image has an alpha channel, then perform a second call on just the alpha channel.

lovell:
I've addressed this in commit c41b873, which adds support for trimming images where the alpha channel contains the images (and non-alpha channels are a single colour) and maintains backwards compatibility with existing behaviour.

@muse-monotype Thank you very much for taking the time to submit the original PR Martin, these are always highly appreciated.

PR #1891, opened 2019-09-27
add installing with permissions tip to install doc

estevezluis1:
Users trying to install sharp globally or installing parent packages globally with sharp as a dependency typically runs into permissions error.

Therefore, I'm requesting permission (ah more permission blockers haha) to add this note to install.md.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -4,6 +4,9 @@
 npm install sharp
 ```
 
+If you are having permissions error while attempting to install globally,
+try `npm install -g --unsafe-perm sharp` or `npm config set user root` then `npm install -g sharp`.
+
 ```sh
 yarn add sharp
 ```


lovell:
Thanks for the PR.

Do people still install modules globally?

I'm a bit reluctant to "officially" recommend that people do this, and usually advise against it, especially as tools such as `npx` and `yarn dlx` allow pseudo-global behaviour for command line utils without the permissions problem.

Are you aware of a particular module/utility that depends on sharp that currently recommends global installation?

estevezluis1:
Sadly yes, ionic-team/cordova-res recommends it's package to install globally and has sharp as dependent. I don't like it especially since it's not a CLI tool.

On second thought, this is not sharp's business. I'll remove this PR.
Thanks for replying!

lovell:
No worries, perhaps the cordova-res repo would accept a PR to update its docs to recommend the use of `npx cordova-res` etc. rather than global installation.

Issue #1918, opened 2019-10-17
Tile overlap regression in v0.23


rustyguts:
I think I've found a regression with the `overlap` attribute in `tile()`.

Steps to reproduce

- yarn add sharp@0.23.1 or 0.23.0
- run script with an image

```json
width: 256,
height: 256,
```

- yarn add sharp@0.22.1
- run script with an image

```json
width: 257,
height: 257,
```

Script to reproduce

```js
(async () => {
  const sharp = require('sharp');
  const fs = require('fs-extra');

  const testImageName = 'test.jpg';

  fs.emptyDirSync(`${testImageName}_files`);
  fs.removeSync(`${testImageName}_files`);
  fs.removeSync(`${testImageName}.dzi`);

  await sharp(fs.readFileSync(testImageName))
    .tile({ size: 256, overlap: 1 })
    .toFile(`${testImageName}.dz`);

  const { width, height } = await sharp(
    `${testImageName}_files/13/0_0.jpeg`,
  ).metadata();

  console.log(`tile is ${width}x${height}`);
})();
```

lovell:
Thanks for reporting this, I can reproduce it too.

It looks like there's a typo at the following line:

https://github.com/lovell/sharp/blob/ba46ad1fd9f9bea37fdc2146989e17ea55074ac1/lib/output.js#L577

...which should be `options.overlap`.

Happy to accept a PR if you're able, even better if it adds your sample code as a new unit test to help prevent future regression.

rustyguts:
I'll make a PR and add some more unit tests, thanks!

lovell:
v0.23.3 now available, thank you for fixing this.

rustyguts:
Thanks for maintaining! 👍

PR #1921, opened 2019-10-17
Ensure tiles have overlap applied

rustyguts:
This merge request addresses #1918, a tile overlap regression spotted in `v0.23.0` and higher.

I've made the change needed to fix the regression, I've also added a test to prevent it from happening again.

Let me know if there are any improvements that I can make, this is my first PR for Sharp!

### DIFF
--- a/test/unit/tile.js
+++ b/test/unit/tile.js
@@ -118,7 +118,6 @@ const assertTileOverlap = function (directory, tileSize) {
           assert.strictEqual(true, metadata.width > tileSize);
           assert.strictEqual(true, metadata.height > tileSize);
         }
-
         done();
       }
     });
@@ -331,7 +330,7 @@ describe('Tile', function () {
           assert.strictEqual(2225, info.height);
           assert.strictEqual(3, info.channels);
           assert.strictEqual('undefined', typeof info.size);
-          assertTileOverlap(directory, 512, 16);
+          assertTileOverlap(directory, 512);
           assertDeepZoomTiles(directory, 512 + (2 * 16), 13, done);
         });
     });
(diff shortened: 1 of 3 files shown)


test/unit/tile.js
```diff
@@ -91,6 +91,39 @@ const assertGoogleTiles = function (directory, expectedTileSize, expectedLevels,
...
+    });
+  });
+
+  eachLimit(tiles, 8, function (tile, done) {
+    sharp(tile).metadata(function (err, metadata) {
+      if (err) {
+        done(err);
+      } else {
+        // Only checks square tiles
+        if (metadata.width >= tileSize && metadata.height >= tileSize) {
```
lovell: Is there a way to verify the presence of overlap without introducing a conditional statement into the unit tests? Perhaps we could check only the penultimate level, the images in which should all be square and should all meet this condition.

rustyguts: I've adjusted the test case in question. Now the test will choose a square tile from the approximate center of the image in the outermost tile level. Thanks for the feedback and let me know what you think.

test/unit/tile.js
```diff
@@ -297,6 +319,7 @@ describe('Tile', function () {
           assert.strictEqual(2225, info.height);
           assert.strictEqual(3, info.channels);
           assert.strictEqual('undefined', typeof info.size);
+          assertTileOverlap(directory, 512);
```
lovell: The assertions within `assertTileOverlap()` are async so that function should probably accept a callback parameter to  allow the call to `assertDeepZoomTiles()` be be deferred until after it has completed.

rustyguts: Makes total sense. To remedy this, I've added the callback parameter and placed `assertTileOverlap()` in the callback for `assertDeepZoomTiles()`. I thought it made the most sense to check the tile overlap after `assertDeepZoomTiles()`.

lovell:
Marvellous, thank you very much Brendan for both fixing this bug and adding new tests to help prevent it occurring again.

Issue #1923, opened 2019-10-18
tile() with background color


neave:
[libvips dszave](https://libvips.github.io/libvips/API/current/VipsForeignSave.html#vips-dzsave) provides a `background` option to set the default background color of the tiles, but this option does not seem to work in sharp. Am I doing this incorrectly or is this option not implemented?

Example:

	sharp('file.png')
		.tile({
			size: 512,
			overlap: 0,
			layout: 'google',
			background: '#000'
		})
		.toFile(tilesDirectory, (error, info) => {
			done();
		});

lovell:
Hi, as you've seen, the `background` option of libvips' `dzsave` operation is not currently exposed in sharp.

https://sharp.pixelplumbing.com/en/stable/api-output/#tile

Happy to accept a PR to add this if you're able.

neave:
OK thanks @lovell. I've had a go but C++ isn't my strong suit. See PR #1924.

lovell:
v0.23.2 now available, thanks for the PR!

michael7213:
hello paul i m with problems with star atlas

michael7213:
> ```
https://staratlas.com/ 
![image](https://user-images.githubusercontent.com/57768203/69837947-03d14d00-1230-11ea-8095-1cce631d98e9.png) belt of venus and earth shadow 
the star atlas want the new atmosphere i want a new

michael7213:
Precomputed Atmospheric Scattering Figure 7: Validation. The sky luminance relatively to the zenith luminance for several sun zenith and view zenith angles (and null azimuth between view and sun directions). Comparison between our model (with α¯ = 0.1, β
s
M = 2.210−5 m−1
, β
s
M/β
e
M = 0.9, g = 0.73 and HM = 1.2 km)
and the CIE sky model 12 (based on actual measurements). We note an overestimation near the horizon (view angles near 90 and -90), which
is also visible in Figure 6. As shown in [ZWP07] the Preetham model [PSS99] also suffers from this problem, which probably comes from the
physical models currently used in CG.
Figure 8: Results. (a), from top to bottom: [SFE07], single scattering, multiple scattering and photo. With [SFE07] the shadow does not
appear due to the missing ν parameter. It is too dark with single sattering only. (b) sunset viewed from space. (c) the view used for our
performance measurements.

c 2008 The Author(s)
Journal compilation
c 2008 The Eurographics Association and Blackwell Publishing Ltd.
[...]

PR #1924, opened 2019-10-18
Add background colour option to tiles operation (#1923)

neave:
I've done the best I can to add background colour as an option for the tile() method. I may have missed something here as I'm not a C++ programmer. Let me know your thoughts. Thanks.

### DIFF
--- a/lib/output.js
+++ b/lib/output.js
@@ -605,9 +605,7 @@ function tile (options) {
       }
     }
     // Background colour
-    if (is.defined(options.background)) {
-      this._setBackgroundColourOption('tileBackground', options.background);
-    }
+    this._setBackgroundColourOption('tileBackground', options.background);
     // Depth of tiles
     if (is.defined(options.depth)) {
       if (is.string(options.depth) && is.inArray(options.depth, ['onepixel', 'onetile', 'one'])) {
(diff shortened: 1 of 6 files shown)


neave:
Hm it seems the tests fail in regards to skipBlanks, but the PR doesn't touch that code. Any idea what's going on here? Thanks!

lovell:
Thanks for the PR, I think the test failures are due to the need to match the number of channels in the "background" with the number of channels in the image. Try conditionally using `pop_back` to remove the last element when there is no alpha channel.

neave:
OK that seemed to do the trick. It can now set the tiles background colour and passes the tests. I've committed the changes but it's now reporting coverage failure?

lib/output.js
```diff
@@ -603,6 +604,14 @@ function tile (options) {
         throw is.invalidParameterError('angle', 'positive/negative multiple of 90', options.angle);
       }
     }
+    // Background colour
+    if (is.defined(options.background)) {
```
lovell: The `_setBackgroundColourOption` function already contains validation logic so we should be able to use a single, conditionless call to `this._setBackgroundColourOption(options.background)` here.

neave: OK I've tried that but test coverage still seems to fail 🤷‍♂️

lovell: It should be possible to remove the `if (is.defined(...` check too. (I've just spotted a call to `_setBackgroundColourOption` in `lib/resize.js` also contains this unnecessarily-repeated logic, which I'll remove separately.)

neave: OK great, all checks now pass.

lovell:
Brilliant, thank you Paul.

caioaugusto1998:
we report the guy that send emailing to we finally we report the guy

ferreiradantas7:
@lovell hello lovell i need of help with the applications you can say to paul neave

PR #1952, opened 2019-11-06
FreeBSD uses stat instead of stat64

pouya-eghbali:
`stat64` does not exist on FreeBSD.

### DIFF
--- a/src/pipeline.cc
+++ b/src/pipeline.cc
@@ -38,7 +38,7 @@
 #elif defined(__APPLE__)
 #define STAT64_STRUCT stat
 #define STAT64_FUNCTION stat
-#elif defined(__FreeBSD__)
+#elif defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__NetBSD__) || defined(__DragonFly__)
 #define STAT64_STRUCT stat
 #define STAT64_FUNCTION stat
 #else
--- a/src/pipeline.cc
+++ b/src/pipeline.cc
@@ -38,6 +38,9 @@
 #elif defined(__APPLE__)
 #define STAT64_STRUCT stat
 #define STAT64_FUNCTION stat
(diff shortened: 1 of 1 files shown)


src/pipeline.cc
```diff
@@ -38,6 +38,9 @@
 #elif defined(__APPLE__)
 #define STAT64_STRUCT stat
 #define STAT64_FUNCTION stat
+#elif defined(__FreeBSD__)
```
lovell: Is it worth expanding this to include variants, e.g.:
```c
#elif defined(__FreeBSD__) || defined(__OpenBSD__) || defined(__NetBSD__) || defined(__DragonFly__)
```

pouya-eghbali: Ok I checked `sys/stat.h` on OpenBSD, NetBSD and DragonFly, they're all stat instead of stat64. Wanted to include all of them but I wasn't sure if they're all stat.

pouya-eghbali: I made the change.

lovell:
Brilliant, thank you, I'll add FreeBSD to the CI matrix to help prevent this occurring again.

pouya-eghbali:
@lovell it may need extra steps to build on FreeBSD. Latest vips in FreeBSD ports is 8.7.4. I had to fork the port and update it to 8.8.3 to make sharp work.

lovell:
@pouya-eghbali Thanks, are you able to submit a PR/patch to ports itself to upgrade it to the latest vips?

/cc @daniloegea in case this is already happening

lovell:
@pouya-eghbali I've created #1953 to track the addition CI for FreeBSD.

pouya-eghbali:
I sent an email to @daniloegea .
I'll prepare diffs and make a PR if necessary.

Issue #2020, opened 2020-01-02
resize: withoutEnlargement: Height disregarded when cover && width > original width


BrychanOdlum:
What are the steps to reproduce?
1. Take an image of dimensions 1500 x 1000.

2.
```
await Sharp.file(imageData)
  .resize({ width: 1750, height: 250, fit: 'cover', withoutEnlargement: true });
```

What is the expected behaviour?
An image of 1500 x 250.

Instead we get an image of 1500 x 1000.

Example code:
https://github.com/BrychanOdlum/Sharp-BugExample/blob/master/main.js
https://github.com/BrychanOdlum/Sharp-BugExample (has input & output image in repo)

What is the output of running `npx envinfo --binaries --languages --system --utilities`?
  System:
    OS: macOS 10.15
    CPU: (8) x64 Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
    Memory: 232.27 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 12.11.0 - ~/.nvm/versions/node/v12.11.0/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.11.3 - ~/.nvm/versions/node/v12.11.0/bin/npm
  Utilities:
    CMake: 3.15.4 - /usr/local/bin/cmake
    Make: 3.81 - /usr/bin/make
    GCC: 4.2.1 - /usr/bin/gcc
    Git: 2.21.0 - /usr/bin/git
    Clang: 1100.0.33.8 - /usr/bin/clang
    Subversion: 1.12.2 - /usr/local/bin/svn
  Languages:
    Bash: 3.2.57 - /bin/bash
    Go: 1.12.6 - /usr/local/go/bin/go
    Java: 1.8.0_161 - /usr/bin/javac
    Perl: 5.30.0 - /usr/local/bin/perl
    PHP: 7.3.8 - /usr/bin/php
    Python: 2.7.16 - /usr/local/bin/python
    Python3: 3.7.5 - /usr/local/bin/python3
    Ruby: 2.6.3 - /usr/bin/ruby

BrychanOdlum:
At the surface it looks as if this here is the culprit:
https://github.com/lovell/sharp/blob/755a0caf3d495b57529221504bece16930766b7e/src/pipeline.cc#L226-L237

Might changing that to something like this instead do the trick?
```
      if (baton->withoutEnlargement) {
[...]

PR #2024, opened 2020-01-06
Correctly bind max width and height values when using withoutEnlargement

BrychanOdlum:
Fix re #2020

When the `withoutEnlargement` option is set to true the pipeline *should* now bind the input image's width and height as maximum.

Previously if one value was in bounds and one was out of bounds, then the output image would keep the dimensions of the input image.

Also added two unit tests to back this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,8 @@
     "Ilya Ovdin <iovdin@gmail.com>",
     "Andargor <andargor@yahoo.com>",
     "Paul Neave <paul.neave@gmail.com>",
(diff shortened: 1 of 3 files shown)


BrychanOdlum:
Not sure what's up with Travis here, but it's showing:
```
fatal: unable to access 'https://github.com/lovell/sharp.git/': Could not resolve host: github.com
```

😢

lovell:
Marvellous, an even simpler approach, thank you very much Brychan.

Issue #1160, opened 2018-03-16
Enhancement: composite with negative values for top and/or left


seripap:
Hello! 

Is there a reason why negative values are not supported with `top` and `left` from within overlayWith? I am not familiar with how libvips work as well, so feel free to close this if this is something that is specifically not supported.

A use case for this is composing coordinate-based map images (such as a radar image on top of a static base map).

For example, a base map could be cropped like this:

<img width="746" alt="screen shot 2018-03-16 at 9 27 24 am" src="https://user-images.githubusercontent.com/683200/37523097-4d801162-28fc-11e8-9804-70a7cc519c3c.png">

The square outlines for this base map indicates the position of a 256x256 tile image. From this information, we can calculate the pixel position of where tile should go and place them where they belong. So the first tile could be at x:-44,y:-102, second would be x:212,y:-102, and so on.

Please let me know if I missed something from the docs or this isn't something that is supported.

zekth:
I have managed to do something this but needs to calculate top / left from metadata of base image and overlay image. But with this approach i don't use gravity because i'm forced to calculate the coordinates of the overlay.
For example setting an overlay in SouthEast with a margin of 20px. I can share my code if you want.

christianflintrup:
Now that the new composite operation is implemented, is this feature request being considered for implementation? Or is it tricky due to how libvips works?

lovell:
@christianflintrup Always happy to accept a PR, if you're able.

ariklevin:
Hi,

Is there a workaround, until the PR is completed?

10x

lovell:
v0.27.0 is now available, thank you for the PRs to improve this feature!

PR #2071, opened 2020-02-05
Allow for negative top/left offsets in composite overlays

christianflintrup:
Composite overlays can be negatively offset, as requested in #1160.

Potentially breaking changes:

- A top or left offset value of -1 will no longer mean that the value is not set, but will now be an actual offset of -1.
- Only either of top and left offsets are now required; not both.


lovell:
Thank you very much for this PR Christian, I'll try to make time to review it properly next week.

Does this approach still support composition at the top-left corner, i.e. top=0 and left=0? It might be worth adding this as a test case either way.

lovell:
Thanks to your PR, I noticed the thresholds of some of the existing composite offset unit tests were not protecting us well enough against regression. 

Please can you rebase this PR against the latest upstream master to pick up commit 1717173 to check the updated "zero offset" test case.

CurosMJ:
👍

ariklevin:
Any plans to approve this 1?

lovell:
@christianflintrup Please can you rebase this PR to pick up the newer tests mentioned in https://github.com/lovell/sharp/pull/2071#issuecomment-586578782

CurosMJ:
@christianflintrup @lovell I'll be happy to fix this, waiting for a few days for the original author to rebase, if not I will open up another PR. 😄

Thanks!

lovell:
Closing in favour of #2391 but thank you very much @christianflintrup for starting things - you will get credit when it's merged.

PR #2094, opened 2020-02-21
Clarify stats.isOpaque

joeflateau:
It's hard to tell from the docs whether `stats.isOpaque` means that all pixels are 100% opaque, or if it means at least one pixel is not 100% transparent.


joeflateau:
Ahh, okay. Sure thing, I'll update this later today. Thank you

lovell:
Brilliant, thank you, the docs will be updated as part of the next release.

Issue #1335, opened 2018-08-14
Support for Level 0 IIIF Image API?


edsilv:
Hi,

I'm working on a nodejs library for generating IIIF manifests/collections: 

https://github.com/edsilv/biiif

Right now you need to generate image tiles using a python script: 

https://github.com/edsilv/biiif#image-services

It would be nice to be able to use nodejs (specifically Sharp) to generate the tiles, but it looks like only the dzi format is currently supported. It's possible to use dzi in an image service like this: 

https://gist.github.com/tomcrane/34180dc6c5c1caf7e244ffd12e64c389#what-about-the-infojson-what-about-an-av-api

But I expect that the IIIF community will prefer to use the format that they are accustomed to.

I was wondering if you had any plans to support the [Level 0 IIIF Image API](https://iiif.io/api/image/3.0/compliance/#5-level-0-compliance) in the [tile](http://sharp.pixelplumbing.com/en/stable/api-output/#tile) method?

lovell:
Hello, `tile` is a wrapper around `vips_dzsave`.

Please see http://jcupitt.github.io/libvips/API/current/Making-image-pyramids.md.html and http://jcupitt.github.io/libvips/API/current/VipsForeignSave.html#vips-dzsave for details of what it currently supports.

It would make most sense to add support for the IIIF directory layout in libvips itself. https://github.com/aaronland/go-iiif/issues/20#issuecomment-409675639 is relevant here.

lovell:
libvips v8.9.0 will have this - see https://github.com/libvips/libvips/issues/1465

The `layout` parameter of `tile()` will need extending to also accept `iiif` as a possible value.

https://sharp.pixelplumbing.com/en/stable/api-output/#tile

mbklein:
FWIW – and I'm not sure this is the right place to mention it, but here goes – I've written a fast, lightweight HTTP wrapper for sharp that implements the IIIF 2.1 spec. Realtime extraction of tiles and regions of interest along with IIIF's custom cropping, rotation, and color options. Works best with pyramid TIFF source images, but it's compatible with any image format libvips can read. https://github.com/nulib/node-iiif

lovell:
v0.25.2 is now available with IIIF support, thanks again for the PR.

PR #2098, opened 2020-02-23
IIIF Support

edsilv:
Starting a PR to include libvips' new IIIF support as per: https://github.com/libvips/libvips/issues/1465#issuecomment-552487666

However, with the changes included in this PR as per https://github.com/lovell/sharp/issues/1335#issuecomment-552613752 it doesn't generate the tiles and info.json...

I thought passing the correct layout should now work? e.g.

```
    sharp("test.jpg")
    .jpeg()
    .tile({
      size: 512,
      layout: "iiif"
    })
    .toFile("info.json");
```

As far as I understand it, the layout should be getting passed here? 

https://github.com/lovell/sharp/blob/master/src/pipeline.cc#L966

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@edsilv/sharp",
+  "name": "sharp",
   "description": "High performance Node.js image processing, the fastest module to resize JPEG, PNG, WebP and TIFF images",
   "version": "0.25.1",
   "author": "Lovell Fuller <npm@lovell.info>",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "sharp",
   "description": "High performance Node.js image processing, the fastest module to resize JPEG, PNG, WebP and TIFF images",
-  "version": "0.25.0",
+  "version": "0.25.1",
   "author": "Lovell Fuller <npm@lovell.info>",
   "homepage": "https://github.com/lovell/sharp",
   "contributors": [
--- a/package.json
+++ b/package.json
@@ -135,7 +135,7 @@
   },
   "license": "Apache-2.0",
   "config": {
-    "libvips": "8.9.1"
+    "libvips": "8.9.0"
   },
   "engines": {
     "node": ">=10"
--- a/package.json
+++ b/package.json
@@ -135,7 +135,7 @@
   },
   "license": "Apache-2.0",
   "config": {
-    "libvips": "8.9.0"
+    "libvips": "8.9.1"
   },
   "engines": {
     "node": ">=10"
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "sharp",
+  "name": "@edsilv/sharp",
   "description": "High performance Node.js image processing, the fastest module to resize JPEG, PNG, WebP and TIFF images",
   "version": "0.25.1",
   "author": "Lovell Fuller <npm@lovell.info>",
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "sharp",
   "description": "High performance Node.js image processing, the fastest module to resize JPEG, PNG, WebP and TIFF images",
-  "version": "0.25.1-beta",
+  "version": "0.25.1",
   "author": "Lovell Fuller <npm@lovell.info>",
   "homepage": "https://github.com/lovell/sharp",
   "contributors": [
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "sharp",
   "description": "High performance Node.js image processing, the fastest module to resize JPEG, PNG, WebP and TIFF images",
-  "version": "0.25.1",
+  "version": "0.25.1-beta",
   "author": "Lovell Fuller <npm@lovell.info>",
   "homepage": "https://github.com/lovell/sharp",
   "contributors": [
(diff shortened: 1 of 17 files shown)


lovell:
Hi Edward, thank you for the PR, I think we'll need to change this logic too:

https://github.com/lovell/sharp/blob/d5ecc537afa9b5329e79d4a6e50b945dfcf2fcc6/src/pipeline.cc#L1380-L1387

Rather than if-else-if-else... we should probably change this whole block to something simpler like the following (untested):

```cc
baton->tileLayout = static_cast<VipsForeignDzLayout>(
    vips_enum_from_nick(nullptr, VIPS_TYPE_FOREIGN_DZ_LAYOUT,
    sharp::AttrAsStr(options, "tileLayout").data()));
```
If you're able to add a test or two as well, that would be perfect.

lovell:
Please can you target the `yield` branch to ensure your changes are compatible with the forthcoming N-API support.

lovell:
I simplified the tile option parsing in commit 51b1432 on the `yield` branch, so if you rebase against upstream/yield you should hopefully start to see this new option work.

lovell:
v0.25.0 is now available so I've moved this PR back to the master branch.

edsilv:
Hi, nice! Do you need anything further from me? I was a bit out of my depth on the c++ stuff and winded up needing dynamic tiling using [serverless-iiif](https://github.com/nulib/serverless-iiif) (also libvips-based). Looking forward to getting IIIF tiling in Sharp and updating [biiif](https://github.com/edsilv/biiif)!

lovell:
Thanks for the update. Please can you rebase against the latest upstream/master branch and you should the proposed changes in the PR start to work. If you're able to add a unit test or two as well, that would be great, then we can get this merged and out in the next release.

edsilv:
I've pushed the beginnings of a unit test and tried the outputted tiles here: https://repl.it/@edsilv/osd-test

You can see that some tiles are missing when fully zoomed out.

I've also had to add `"preferredFormats": [ "jpeg" ]` to the generated `info.json` as otherwise OSD will look for files ending `.jpg` by default. To get around _that_ you probably would have to do some gymnastics with `IIIFTileSource`, but really it would be better if it just worked out of the box with no customisation. 

Any chance we could change the default outputted tile format to `.jpg` not `.jpeg`?

lovell:
The JPEG suffix logic at https://github.com/lovell/sharp/blob/master/src/pipeline.cc#L936 will need updating to also include `VIPS_FOREIGN_DZ_LAYOUT_IIIF` in the list of layouts for which `.jpg` is used.

lovell:
I had a quick look at this and inverted the JPEG extension logic via commit cf39fc4 to make this a bit simpler, which should hopefully allow you to continue with this PR.

edsilv:
Have tested that locally and the tiles are now .jpg and the generated info.json works without any modifications. However, I'm getting a timeout error when running `npm test`, even if I set the timeout to 5 mins: ` IIIF layout:Error: Timeout of 300000ms exceeded.`
Could this be why I'm seeing missing tiles when fully zoomed out?

lovell:
The "IIIF layout" test is failing as the call to `done()` is commented out (as are the assertions).

https://travis-ci.org/github/lovell/sharp/jobs/662551580#L1596

edsilv:
d'oh - sorry, it was a late night!
That's fixed, but still seeing the bug in OSD: https://youtu.be/69oYSDI22ik

lovell:
Could https://github.com/libvips/libvips/issues/1530 and/or https://github.com/libvips/libvips/issues/1472 relate to this problem?

edsilv:
Some tiles are not available: 
image

edsilv:
(changing `scalefactors` to `scaleFactors` didn't make a difference)

edsilv:
The input image `2569067123_aca715a2ee_o.jpg` is `2725px` in width. The first failing tile in the image above has an `xywh` region of `2048,0,678,2048`. `x:2048 + w:678 = 2726px`, 1px more than the full width of the image. Could this be the problem? Libvips is failing to cut that tile because its bounds exceed the width of the image?

lovell:
Yes, I think this might be a libvips thing. Are you able to reproduce it via the command line?

```sh
vips dzsave input output --layout iiif
```

edsilv:
I downloaded v8.9.1 and am getting the same issue via the command line. I suppose we should flag this with libvips? The explanation above still applies.

lovell:
Yes please, if you could open an upstream issue in libvips that would be great.

It doesn't stop this PR so when you find a moment, please can you uncomment that lovely unit test code, get the assertions passing then we're good to merge, thank you.

edsilv:
Have fixed the tests and left a comment here: https://github.com/libvips/libvips/issues/1472#issuecomment-600270954

package.json
```diff
@@ -135,7 +135,7 @@
   },
   "license": "Apache-2.0",
   "config": {
-    "libvips": "8.9.1"
+    "libvips": "8.9.0"
```
lovell: Please can you remove this change.

test/unit/tile.js
```diff
@@ -91,6 +91,40 @@ const assertGoogleTiles = function (directory, expectedTileSize, expectedLevels,
   });
 };
 
+// const assertIIIFTiles = function (directory, expectedSize, expectedLevels, done) {
```
lovell: Please can you remove this commented-out test code if no longer required.

package.json
edsilv: done

test/unit/tile.js
edsilv: done

lovell:
Marvellous, thank you Edward, this will be in v0.25.2.

Issue #2105, opened 2020-02-27
libvips binary download does not allow proxying of GitHub releases


rbywater:
Did you see the [documentation relating to installation](https://sharp.pixelplumbing.com/install)?
Yes

Have you ensured the platform and version of Node.js used for `npm install` is the same as the platform and version of Node.js used at runtime?
N/A

If you are installing as a `root` or `sudo` user, have you tried with the `npm install --unsafe-perm` flag?
N/A

If you are using the `ignore-scripts` feature of `npm`, have you tried with the `npm install --ignore-scripts=false` flag?
N/A

What is the complete output of running `npm install --verbose sharp`? Have you checked this output for useful error messages?
Checked and problem highlighted when it tries to download binary. Available on request.

What is the output of running `npx envinfo --binaries --languages --system --utilities`?
Available on request.

---

As we don't want to constantly redownload the binaries required by the sharp node module, we attempted to setup a proxy in our on-premise artifact server to the Github release URL for sharp-libvips. However we have found that this does not work as, if you override the Sharp Distribution Base URL, the version portion of the directory is not passed and therefore a 404 is received.

e.g. If you set SHARP_DIST_BASE_URL="https://artifacts.example.com/repository/node-sharp-binaries/, instead of looking for https://artifacts.example.com/repository/node-sharp-binaries/v8.9.0/libvips-8.9.0-darwin-x64.tar.gz, it tries to find https://artifacts.example.com/repository/node-sharp-binaries/libvips-8.9.0-darwin-x64.tar.gz.

This appears to be because at https://github.com/lovell/sharp/blob/master/install/libvips.js#L18 the code adds the version number to the base URL for the default GitHub URL but not for the overridden base URL. One could argue that you could add the required part of the directory to the base URL provided, but this isn't really easy to do when the sharp module is being used as a dependency on another module you are trying to install.

(Note that to make sure it isn't a proxy repo issue, I've manually curl'd the correct path from the proxy repo and I do successfully get the file back.)

lovell:
Happy to accept a PR that also allows the setting of a new environment variable that will insert the version number dynamically, if you're able. Please use `npm_config_libvips_binary_host` for this (so it matches the `npm_config_sharp_binary_host` pattern used for prebuilt sharp binaries).

This change will help people in China as I believe many are using the Taobao/Alibaba mirror at https://npm.taobao.org/mirrors/sharp-libvips

rbywater:
👍Thanks - will look at putting something together.

[...]

PR #2110, opened 2020-03-02
Add ability to fetch libvips from Github Release mirror

rbywater:
Fixes https://github.com/lovell/sharp/issues/2105

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -86,7 +86,23 @@ Building from source requires:
 
 This is an advanced approach that most people will not require.
 
-To install the prebuilt libvips binaries from a custom URL,
+To install the prebuilt libvips binaries from a GitHub releases mirror,
+set the `sharp_libvips_binary_host` npm config option
+or the `SHARP_LIBVIPS_BINARY_HOST` environment variable.
+
+For example, both of the following will result in an attempt to download the file located at
+`https://hostname/path/vx.y.z/libvips-x.y.z-platform.tar.gz`
+
+```sh
+npm config set sharp_libvips_binary_host "https://hostname/path"
+npm install sharp
+```
+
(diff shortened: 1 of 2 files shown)


docs/install.md
```diff
@@ -86,7 +86,23 @@ Building from source requires:
 
 This is an advanced approach that most people will not require.
 
-To install the prebuilt libvips binaries from a custom URL,
+To install the prebuilt libvips binaries from a GitHub releases mirror,
+set the `sharp_libvips_binary_host` npm config option
+or the `SHARP_LIBVIPS_BINARY_HOST` environment variable.
```
lovell: What does the `SHARP_LIBVIPS_BINARY_HOST` environment variable provide beyond the logic being introduced by the (implicit) `npm_config_sharp_libvips_binary_host` environment variable?

rbywater: I guess the main rationale was just following along with the npm config / environment variable combo pattern that was used for the Sharp Dist Base URL. I've found personally that environment variables are slightly easier to work with in some build systems than npm config changes for some work I've done in the past so I guess that was also in the back of my mind..

Happy to go either way however so will await your decision and take it from there :)

lovell: Thanks, let's keep things consistent with `npm_config_sharp_binary_host` and expose only `npm_config_sharp_libvips_binary_host`. After this is merged, I'd like to deprecate `SHARP_DIST_BASE_URL` at some point (and eventually remove it) as I think this new dynamic option you're proposing will help/allow people to upgrade more easily.

lovell:
Thanks again for the PR. I think this has now been superseded by #2217 but please feel free to re-open if not, or if further changes are required to provide the feature you need.

PR #2120, opened 2020-03-09
Added N-API badge

NickNaso:
I provided to add the N-API badge to the readme. This could help to identify your native addon as supporting N-API.
See: https://github.com/nodejs/abi-stable-node#badges

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,7 @@ readableStream
 ```
 
 [![Test Coverage](https://coveralls.io/repos/lovell/sharp/badge.svg?branch=master)](https://coveralls.io/r/lovell/sharp?branch=master)
+![N-API v3 Badge](https://img.shields.io/badge/N--API-v3-green.svg)
 
 ### Documentation
 


lovell:
Hi, thanks for the PR, are you able to make this badge link directly to the N-API support matrix so it provides greater utility?

https://nodejs.org/dist/latest/docs/api/n-api.html#n_api_n_api_version_matrix

NickNaso:
Hi @lovell,
for now we have the badges only on [abi-stable-node](https://github.com/nodejs/abi-stable-node#badges) repo. I will propose your idea at next N-API meeting.

lovell:
Thanks Nick. As an aside, the support matrix in the Node.js docs is a little confusing at first glance as it could suggest that e.g. by using N-API v3, there might be a lack of support for Node.js 12 and 13. We know this is not the case, but someone less acquainted with N-API could easily draw the wrong conclusion.

lovell:
Landed via commit 40c0003 - grazie!

PR #2123, opened 2020-03-12
Adding a scale-down resize option

soundsgoodsofar:
I have a use case where I need to scale down the image when it is larger, and keep it the same when it is smaller than the target dimensions.  This is an offered scaling type in the CSS object-fit api that is loosely followed (https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit).

Seems like a pretty simple addition, let me know what you think.

### DIFF
--- a/lib/resize.js
+++ b/lib/resize.js
@@ -69,7 +69,8 @@ const fit = {
   cover: 'cover',
   fill: 'fill',
   inside: 'inside',
-  outside: 'outside'
+  outside: 'outside',
+  scaledown: 'scaledown'
 };
 
 /**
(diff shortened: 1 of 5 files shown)


lovell:
Hi, thanks for the PR, did you see the `withoutEnlargement` option, which might provide what you're looking for?

https://sharp.pixelplumbing.com/api-resize#parameters

soundsgoodsofar:
@lovell hey, yep tested it and it appears that does what I'm looking for, thanks!

While I have you, can I suggest making the language of the "cover" and "contain" descriptions a little more precise?  I never would have guessed that "contain" letterboxes the image from the current description.  Probably not worth a PR, but just 2 cents.

Best

lovell:
Thanks for confirming, and thanks for the nudge to improve the docs as these (and PRs) are always very welcome. Commit c9bff94 should make things a little clearer.

Issue #524, opened 2016-07-21
Example for getting a "finish" event on multiple writable streams


dsine-de:
Hi,

I want to listen for a "finish" event of two writable streams when using the ```.clone()``` function, to do something after multiple files have been written. Is there an example on how to do this?

Here's an example code using an express route:

````js 
var stream = require('stream');
var sharp = require('sharp');

...

var writeStream1 = fs.createWriteStream('test1.jpg').on('finish', () => {
	console.log('Writable Stream 1 finished.');
});
var writeStream2 = fs.createWriteStream('test2.jpg').on('finish', () => {
	console.log('Writable Stream 2 finished.');
});

var pipeline = sharp().sharpen();

pipeline.clone().resize(10, 10).pipe(writeStream1);
pipeline.clone().resize(50, 50).pipe(writeStream2);

var bufferStream = new stream.PassThrough();
bufferStream.end(req.files.file.data); // req.files.file.data is a buffer and needs to be converted to a readable stream)

bufferStream
.on('finish', () => {
	console.log('This is printed before the writable streams are finished.');
})
.pipe(pipeline);
````

Can't figure this out, any help appreciated.
I know there was an example folder earlier, but can't find it anymore.

Thanks!

dsine-de:
I think I solved it like that (using Promise.all) - thanks to myndzi @ #Node.js IRC.

````js
function promiseForStream(stream) {
	return new Promise((resolve, reject) => {
		stream.once('error', reject);
		stream.once('finish', resolve);
	});
}

...

var stream1 = pipeline.clone().resize(10, 10).pipe(writeStream1);
var stream2 = pipeline.clone().resize(50, 50).pipe(writeStream2);

...

Promise.all([promiseForStream(stream1), promiseForStream(stream2)]).then(() => {
	console.log('This prints as soon as both writable streams are finished.');
});

bufferStream.pipe(pipeline);
````

PR #2149, opened 2020-04-02
docs(clone): add promise example

vvo:
It took me some time to figure out how to properly create a pipeline and know
when it's done, while https://github.com/lovell/sharp/issues/524 provides an
example about listening to the finish event, I needed one for toFile.

Now I think we should also document the example from
https://github.com/lovell/sharp/issues/524 but I'd like you the maintainer to
create it because I am not 100% sure this is the most elegant way.

Thanks for the awesome library! It helped me build this: https://coronamaison.net/ (https://github.com/vvo/coronamaison/blob/549a807aea1e2034a0a5821dff6ea89488be46a3/prepareData.js#L153)

### DIFF
--- a/lib/constructor.js
+++ b/lib/constructor.js
@@ -253,6 +253,54 @@ util.inherits(Sharp, stream.Duplex);
  * // firstWritableStream receives auto-rotated, resized readableStream
  * // secondWritableStream receives auto-rotated, extracted region of readableStream
  *
+ * @example
+ * // Create a pipeline that will download an image, resize it and format it to different files
+ * // Using Promises to know when the pipeline is complete
+ * const fs = require("fs");
+ * const got = require("got");
+ * const sharpStream = sharp({
+ *   failOnError: false
+ * });
+ *
+ * const promises = [];
+ *
+ * promises.push(
+ *   sharpStream
+ *     .clone()
+ *     .jpeg({ quality: 100 })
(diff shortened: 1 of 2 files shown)


lib/constructor.js
```diff
@@ -253,6 +253,56 @@ util.inherits(Sharp, stream.Duplex);
...
+ *     .clone()
+ *     .resize({ width: 500 })
+ *     .webp({ quality: 80 })
+ *     .toFile("optimized-500.webp")
+ * );
+ *
+ * // https://github.com/sindresorhus/got#gotstreamurl-options
+ * got.stream("https://www.example.com/some-file.jpg").pipe(sharpStream);
+ *
+ * fs.createReadStream(jpgOriginalImagePath).pipe(sharpStream);
```
lovell, line 293: Might the piping of two streams into the same sharp instance confuse people? I'd prefer to stick to one example, perhaps focus on the use of `got` as the use of HTTP Stream-based input is more common (for filesystem-based input, passing the path to the constructor is preferable).

vvo: ah! that's a mistake of me, in my code those two (got/fs.createReadStream) are in different code paths (if ...), Ill update that! (definitely confusing lol)

vvo: ok just updated the pr now, thx!

lovell:
Landed in commit 1eedb22 - merci beaucoup!

PR #2174, opened 2020-04-20
docs: change 'colourspace' to 'toColourspace'

Fdebijl:
Came across this while updating the typings for DefinitelyTyped, in the example for `.raw()` the method `.colourspace()` is called, which I think should be `.toColourspace()` instead?


lovell:
Good find, thank you for the PR.

Please can you update the source JSDoc from which this is created too: https://github.com/lovell/sharp/blob/master/lib/output.js#L535

lovell:
Wonderbaar, bedankt.

Issue #2183, opened 2020-04-25
Downloading win32 prebuild on linux doesn't copy dlls


dimadeveatii:
Installing using `npm_config_platform=win32` on a linux machine doesn't copy the dlls.

**Steps to reproduce**

- Use a linux machine to install `sharp` using `npm`
- Install with command: `npm_config_platform=win32 npm i sharp`
- Check that dlls are copied to `build/Release` directory

**Actual result**
DLLs are not copied.

**Expected result**
DLLs are copied so that the win32 prebuild is valid when downloaded using a linux machine

I think the problem is in `dll-copy.js` script https://github.com/lovell/sharp/blob/b36ad25030e409d9b8339f5e379346bbe7d764f1/install/dll-copy.js#L9-L12

As in the case of `install/libvips.js` script it should look first at `npm_config_platform` environment variable.

I encountered this issue while trying to build an electron app for windows on linux. The app uses `sharp`.

lovell:
Thanks for reporting this, a PR to look for `npm_config_platform` before `process.platform` would be most welcome.

lovell:
v0.25.3 now available, thank you for fixing this.

PR #2188, opened 2020-04-29
fix: copy dlls for win32 #2183

dimadeveatii:


lovell:
Thank you, this will be included in v0.25.3

PR #2223, opened 2020-05-25
Add support for statically-linked dependencies (lovell/sharp-libvips#39)

kleisauke:
This PR adds supports for a single shared libvips library with its dependencies statically-linked as described in PR https://github.com/lovell/sharp-libvips/pull/44 and issue https://github.com/lovell/sharp-libvips/issues/39.

I've also included several other enhancements:
- Set the minimum macOS deployment target to 10.9 (Mavericks) which is a good balance between features and compatibility. Starting with 10.9, Apple has made libc++ the default, so it is no longer necessary to set `CLANG_CXX_LIBRARY`. Note that [Node v12 has set this variable to 10.10](https://github.com/nodejs/node/commit/a7d7d4dfb73233b23e478c73d6fff784a0738b9b).
- Link the dependencies with a fully-qualified soname (e.g. `libvips.so.42`) as the symbolic links in the tarball are dereferenced now.
- Unpack the prebuilt libvips binaries into a versioned directory.
- Move `library_dirs` and `libraries` to the `link_settings` section within `binding.gyp`, see: https://gyp.gsrc.io/docs/InputFormatReference.md#dependent-settings

Any help with testing would be much appreciated, testing can be done with:
```bash
npm install kleisauke/sharp#statically-linked
```

Note that the last commit must be reverted before merging. This commit is intended for CI testing purposes as the new tarballs have not yet been uploaded to https://github.com/lovell/sharp-libvips/releases.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -1,6 +1,6 @@
 {
   'variables': {
-    'vips_version': '8.9.2',
+    'vips_version': '<!(node -p "require(\'./lib/libvips\').minimumLibvipsVersion")',
     'sharp_vendor_dir': '<(module_root_dir)/vendor/<(vips_version)'
   },
   'targets': [{
@@ -77,8 +77,8 @@
       'runtime_link%': 'shared',
       'conditions': [
         ['OS != "win"', {
-          'pkg_config_path': '<!(node -e "console.log(require(\'./lib/libvips\').pkgConfigPath())")',
-          'use_global_libvips': '<!(node -e "console.log(Boolean(require(\'./lib/libvips\').useGlobalLibvips()).toString())")'
+          'pkg_config_path': '<!(node -p "require(\'./lib/libvips\').pkgConfigPath()")',
+          'use_global_libvips': '<!(node -p "Boolean(require(\'./lib/libvips\').useGlobalLibvips()).toString()")'
         }, {
           'pkg_config_path': '',
           'use_global_libvips': ''
--- a/binding.gyp
+++ b/binding.gyp
@@ -1,4 +1,8 @@
 {
+  'variables': {
+    'vips_version': '8.9.2',
+    'sharp_vendor_dir': '<(module_root_dir)/vendor/<(vips_version)'
+  },
   'targets': [{
     'target_name': 'libvips-cpp',
     'conditions': [
@@ -16,15 +20,18 @@
           'src/libvips/cplusplus/VImage.cpp'
         ],
         'include_dirs': [
-          'vendor/include',
-          'vendor/include/glib-2.0',
-          'vendor/lib/glib-2.0/include'
+          '<(sharp_vendor_dir)/include',
+          '<(sharp_vendor_dir)/include/glib-2.0',
+          '<(sharp_vendor_dir)/lib/glib-2.0/include'
         ],
-        'libraries': [
-          '../vendor/lib/libvips.lib',
-          '../vendor/lib/libglib-2.0.lib',
-          '../vendor/lib/libgobject-2.0.lib'
-        ],
+        'link_settings': {
+          'library_dirs': ['<(sharp_vendor_dir)/lib'],
+          'libraries': [
+            'libvips.lib',
+            'libglib-2.0.lib',
+            'libgobject-2.0.lib'
+          ],
+        },
         'configurations': {
           'Release': {
             'msvs_settings': {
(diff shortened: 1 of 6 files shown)


lovell:
Thanks for the update, we should look at why the "main" FreeBSD Cirrus CI task is failing to build as it usually at least compiles (even if the tests don't always pass). Perhaps use `npm install --unsafe-perm --verbose` to expand the compilation stage?

The Appveyor Windows x86 environment has been a bit wonky for a few weeks. Let me experiment with Travis' new-ish Windows environment to see if that's any less flaky.

lovell:
FreeBSD appears to compile if we revert this change (as you were intending to do).

```diff
-    "install": "node install/libvips && node-gyp rebuild && node install/dll-copy",
+    "install": "(node install/libvips && node install/dll-copy && prebuild-install --runtime=napi) || (node-gyp rebuild && node install/dll-copy)",
```

This is because the presence of a globally-installed libvips produces a non-zero exit code from the `node install/libvips` command.

Commit 13bdb68 should make the Windows x86 CI environment a bit less flaky.

Please can you rebase/squash then set the destination branch to `zoom` for inclusion in the v0.26.0 release and we can merge this in. I expect we'll upgrade to libvips v8.10.0 as part of that release too. Thank you!

kleisauke:
Ah, I assumed that the `node install/libvips` command always returned a zero exit code. The reason for that change was that users could test more easily without having to pass the `--build-from-source` flag during `npm install`.

Rebased and squashed with commit https://github.com/lovell/sharp/pull/2223/commits/fbdac804a0104f09a713ce5e59ae0bd95d67f1b6.

binding.gyp
```diff
@@ -1,4 +1,8 @@
 {
+  'variables': {
+    'vips_version': '8.9.2',
```
kleisauke, line 3: Note that this variable must be in-sync with `minimumLibvipsVersion` from `lib/libvips.js`.

https://github.com/lovell/sharp/blob/85459e0ec6d17db55cd4491e64483cfc5e13d843/lib/libvips.js#L13

Perhaps there's a better alternative?

kleisauke:
There still seem to be some Travis CI flakiness for some jobs:
https://travis-ci.org/github/lovell/sharp/jobs/702629233
https://travis-ci.org/github/lovell/sharp/jobs/702624013

fwiw, here's a successful build for commit https://github.com/lovell/sharp/pull/2223/commits/fbdac804a0104f09a713ce5e59ae0bd95d67f1b6:
https://travis-ci.org/github/kleisauke/sharp/builds/702629210

binding.gyp
lovell: The `npm_package_config_libvips` environment variable should contain the value from `package.json` or we could invoke `<!(node -p "require('./package.json').config.libvips")`, which might be a bit more explicit/safer.

kleisauke: The problem with `<!(node -p "require('./package.json').config.libvips")` is that it contains the labelled version. https://github.com/lovell/sharp/blob/fbdac804a0104f09a713ce5e59ae0bd95d67f1b6/package.json#L140

We need the unlabelled version (i.e. 8.9.2) for the versioned subdirectory because we can't detect that `-alpha2` at runtime (as far as I know).
https://github.com/lovell/sharp/blob/fbdac804a0104f09a713ce5e59ae0bd95d67f1b6/lib/utility.js#L22-L27


Perhaps we should ensure that the version of `package.json` is always unlabelled?

lovell: Something like `node -p "require('semver').coerce(require('./package.json').config.libvips).version"` might do the job (and would still not be the most horrible thing in the binding.gyp file).

kleisauke: I see. Perhaps we could also just use `minimumLibvipsVersion` from `lib/libvips.js`? Something like this:
```suggestion
    'vips_version': '<!(node -p "require(\'./lib/libvips\').minimumLibvipsVersion")',
```

As an aside,
Is the `console.log` required here?:
https://github.com/lovell/sharp/blob/85459e0ec6d17db55cd4491e64483cfc5e13d843/binding.gyp#L73-L74

I think `node -p` could work here, too.

lovell: Yes please to both of these, thank you.

kleisauke: Fixed with https://github.com/lovell/sharp/pull/2223/commits/3e56eedc42ee06b360f12d7669221001af9a343c.

Issue #2224, opened 2020-05-26
Weak documentation for PNG quality and palette options


romaleev:
Hi,
I have PNG and would like to have it minified:
```
sharp(image).png({
    quality: 50
  }).toFile(destination)
```

However I noticed that some of my .png got increased size after conversion.
Secondly I found out that <b>quality</b> option doesn't work at all - always the same output size disregarding 1 or 100.
Thirdly I started looking int docs and found:
> PNG
> options.quality number  use the lowest number of colours needed to achieve given quality, <b>requires libvips compiled with support for libimagequant</b> (optional, default 100)

I spent **hours** for installing macos xcode CLI, libvips, libimagequant, checking if libvips during compilation picked up libimagequant - neither helped.

And finally I checked sharp source and found that `quality` option is only used together with `palette` option:
```
if (is.defined(options.palette)) {
      this._setBooleanOption('pngPalette', options.palette);
      if (this.options.pngPalette) {
        if (is.defined(options.quality)) {
          if (is.integer(options.quality) && is.inRange(options.quality, 0, 100)) {
            this.options.pngQuality = options.quality;
```

So finally I had `quality` option working and image minified with:
```
sharp(image).png({
    quality: 50,
    palette: true
  }).toFile(destination)
```

I find this behavior/documentation very annoying, saw multiple complaints from others, who likely switched to different lib to minify PNG.
Nobody wants to dig that deep, learn PNG specification to simply compress the PNG image.

I propose:
- either to automatically enable PNG `palette` option when `quality` option is set.
- or explicitly specify in documentation that `quality` option is used **only** when `palette` is enabled

lovell:
Happy to accept a PR that does the former, namely set `palette` to `true` when at least one of `quality`, `colours`, `colors` or `dither` is set.

lovell:
v0.25.4 now available, thank you for the PR!

PR #2226, opened 2020-05-27
PNG palette is enabled when at least one of quality, colours, colors …

romaleev:
…or dither is set

[Weak documentation for PNG quality and palette options #2224](https://github.com/lovell/sharp/issues/2224#issuecomment-634291571)


lovell:
Bedankt Roman!

romaleev:
You are welcome! Awesome job with this library!

PR #2244, opened 2020-06-04
Simplify helpers

Connormiha:
I just simplified a couple helpers


lovell:
Thank you for modernising these Mikhail, the previous logic will have been from the days of supporting io.js, Node.js 0.12 and older!

Issue #1315, opened 2018-07-30
Request: allow TypedArray/ArrayBuffer as input


IvanSanchez:
Small feature request.

When providing raw image data, sharp currently accepts a `Buffer` as an input. However, I'm most used to work with `Uint8Array`s (and other `TypedArray`s) and `ArrayBuffer`s when dealing with binary data. Having some APIs that return `Uint8Array` as output and some other APIs that only accept `Buffer`s as input is slightly cumbersome.

I would like the constructor to accept any `TypedArray` or `ArrayBuffer` as input, and let sharp perform any conversion into `Buffer` internally.

lovell:
Hello, did you see #1289?

IvanSanchez:
@lovell Yeah, but #1289 that's a different issue - a new option for the constructor. My request is about the constructor allowing a different type for the image data, for [robustness](https://en.wikipedia.org/wiki/Robustness_principle)/convenience.

In other words, my code right now looks like:

```
    var bytes = new Uint8Array(/* stuff */);
    const image = new sharp(Buffer.from(bytes), {
        raw: { width: w, height: h, channels: 4 }
    });
```

And I would like it to look like 

```
    var bytes = new Uint8Array(/* stuff */);
    const image = new sharp(bytes, {
        raw: { width: w, height: h, channels: 4 }
    });
```

Whereas #1289 is asking for something along the lines of

```
    var bytes = new Float32Array(/* stuff */);
    const image = new sharp(Buffer.from(bytes), {
        raw: { width: w, height: h, channels: 4, bitsPerChannel: 32}
    });
```

lovell:
#1289 will probably involve adding `TypedArray` support anyway so the `bitsPerChannel` option you've suggested can instead be implicitly determined from the input type. I'll add a comment to this effect.

lovell:
Tracking this at #1289

Issue #1289, opened 2018-07-09
Enhancement: Include channel bit depth in raw constructor options.


andrew-higgins-4d:
Sharp appears to store data internally with varying bit depths; i.e. it is possible to load a tiff file saved with 32 bits per channel and the resulting metadata shows a pixel depth format of 'float' instead of 'uchar'.

Currently, it doesn't seem to be possible to create a new sharp instance that will use anything other than 8 bits per channel. It would be useful to be able to set this via the constructor's 'options.raw' object. For example, setting 'options.raw.depth' to '32' or 'float' could then accept a Buffer containing a Float32Array.

The image data could then be saved to file without loss of precision for those formats that support it, or converted to the maximum supported bit depth before saving.

lovell:
Hello, yes, it should be possible to add this feature. Is there a specific use case or integration you're thinking of that might help focus the initial approach? Happy to help with a PR if you're able.

andrew-higgins-4d:
Hi, the use case is reading frame buffer data from a ray tracer which is stored as 32 bit floating point values per channel. The goal would be to maintain precision throughout, as the various render elements are often used during post processing later on.

I'm not sure if I'll have time to work on a PR at the moment, but I may be able to look into it at some point.

lovell:
As discussed in #1315, the channel bit depth can be implicity determined from the type of the `TypedArray`. This means sharp would accept both `Buffer` and concrete `TypedArray` types for raw input.

lu4:
Any progress on this issue? With little help on setting up the C++ debugging environment (standard approach through launch.json in Visual Studio Code wouldn't work) I think I would be able to come out with PR for this feature, anyone?

Thank you in advance!

nnmrts:
I would also like to know the current state of this. Being able to use a Uint8Array or a Uint8ClampedArray would be a game-changer for me, and honestly, although it could be in the scope of this issue, I don't think it should be.

The input could just be "coerced" to the node.js Buffer by using Buffer.from. Or is that not a viable option?

lovell:
@nnmrts 8 bit per channel raw data held in a `Uint8Array` should already "just work".

```javascript
sharp(Buffer.from(uint8Array), {
  raw: { width, height, channels }
})...
```
This future possible enhancement relates to other bit depths.

nnmrts:
[...]

PR #2256, opened 2020-06-12
added uint8array as input and output option

nnmrts:
maybe dirty, but also simple and does the job (AFAIK)

should "fix" parts of #1315, discussion also in #1289

### DIFF
--- a/lib/input.js
+++ b/lib/input.js
@@ -28,9 +28,9 @@ function _createInputDescriptor (input, inputOptions, containerOptions) {
   if (is.string(input)) {
     // filesystem
     inputDescriptor.file = input;
-  } else if (is.buffer(input)) {
-    // Buffer
-    inputDescriptor.buffer = input;
+  } else if (is.uint8Array(input)) {
+    // Buffer or Uint8Array
+    inputDescriptor.buffer = Buffer.from(input);
   } else if (is.plainObject(input) && !is.defined(inputOptions)) {
     // Plain Object descriptor, e.g. create
     inputOptions = input;
--- a/lib/output.js
+++ b/lib/output.js
@@ -144,7 +144,7 @@ function toUint8Array (options, callback) {
   return new Promise((resolve, reject) => {
     this.toBuffer(options, callback).then((buff) => {
       resolve(Uint8Array.from(buff));
-    })
+    });
   });
 }
 
--- a/lib/output.js
+++ b/lib/output.js
@@ -117,6 +117,38 @@ function toBuffer (options, callback) {
 }
 
 /**
+ * Same as toBuffer, but writes output to a Uint8Array.
+ *
+ * @example
+ * sharp(input)
+ *   .toBuffer((err, data, info) => { ... });
+ *
+ * @example
+ * sharp(input)
+ *   .toUint8Array()
+ *   .then(data => { ... })
+ *   .catch(err => { ... });
+ *
+ * @example
+ * sharp(input)
+ *   .toUint8Array({ resolveWithObject: true })
+ *   .then(({ data, info }) => { ... })
+ *   .catch(err => { ... });
+ *
+ * @param {Object} [options]
+ * @param {boolean} [options.resolveWithObject] Resolve the Promise with an Object containing `data` and `info` properties instead of resolving only with `data`.
+ * @param {Function} [callback]
+ * @returns {Promise<Buffer>} - when no callback is provided
+ */
+function toUint8Array (options, callback) {
+  return new Promise((resolve, reject) => {
+    this.toBuffer(options, callback).then((buff) => {
+      resolve(Uint8Array.from(buff));
+    })
+  });
+}
+
+/**
  * Include all metadata (EXIF, XMP, IPTC) from the input image in the output image.
  * This will also convert to and add a web-friendly sRGB ICC profile.
  *
@@ -797,6 +829,7 @@ module.exports = function (Sharp) {
     // Public
     toFile,
     toBuffer,
+    toUint8Array,
     withMetadata,
     toFormat,
     jpeg,
--- a/test/unit/toUint8Array.js
+++ b/test/unit/toUint8Array.js
@@ -1,0 +1,19 @@
+'use strict';
+
+const assert = require('assert');
+
+const sharp = require('../../lib');
+const fixtures = require('../fixtures');
+
+describe('toUint8Array', () => {
+  it('reusing same sharp object does not reset previously passed parameters to toUint8Array', (done) => {
+    const image = sharp(fixtures.inputJpg);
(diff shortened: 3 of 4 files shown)


nnmrts:
dunno why those checks are failing, did i do something wrong?

lib/input.js
```diff
@@ -28,9 +28,9 @@ function _createInputDescriptor (input, inputOptions, containerOptions) {
   if (is.string(input)) {
     // filesystem
     inputDescriptor.file = input;
-  } else if (is.buffer(input)) {
-    // Buffer
-    inputDescriptor.buffer = input;
+  } else if (is.uint8Array(input)) {
+    // Buffer or Uint8Array
+    inputDescriptor.buffer = Buffer.from(input);
```
lovell, line 33: The Node.js docs suggest this will copy data rather than create a reference/view.

https://nodejs.org/dist/latest-v14.x/docs/api/buffer.html#buffer_class_method_buffer_from_buffer

Perhaps use `Buffer.from(input.buffer)` instead?

Perhaps also have this as an additional `else if (is.uint8Array(input))` condition, keeping the original `else if (is.buffer(input))` intact.

test/unit/toUint8Array.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+const assert = require('assert');
+
+const sharp = require('../../lib');
+const fixtures = require('../fixtures');
+
+describe('toUint8Array', () => {
+  it('reusing same sharp object does not reset previously passed parameters to toUint8Array', (done) => {
```
lovell, line 9: Thanks for adding tests to cover the output. Please can you add more to cover both callback and Promise based outputs as well as tests for Uint8Array as input, including Stream-based input.

There should be plenty of examples you can use in https://github.com/lovell/sharp/blob/master/test/unit/io.js

lib/output.js
```diff
@@ -116,6 +116,38 @@ function toBuffer (options, callback) {
...
+ *
+ * @param {Object} [options]
+ * @param {boolean} [options.resolveWithObject] Resolve the Promise with an Object containing `data` and `info` properties instead of resolving only with `data`.
+ * @param {Function} [callback]
+ * @returns {Promise<Buffer>} - when no callback is provided
+ */
+function toUint8Array (options, callback) {
+  return new Promise((resolve, reject) => {
+    this.toBuffer(options, callback).then((buff) => {
+      resolve(Uint8Array.from(buff));
```
lovell, line 146: I think this will copy the data. Would something like `new Uint8Array(buffer, 0, buffer.length)` to create a view be better here?

lib/output.js
```diff
@@ -116,6 +116,38 @@ function toBuffer (options, callback) {
...
+ * @example
+ * sharp(input)
+ *   .toUint8Array({ resolveWithObject: true })
+ *   .then(({ data, info }) => { ... })
+ *   .catch(err => { ... });
+ *
+ * @param {Object} [options]
+ * @param {boolean} [options.resolveWithObject] Resolve the Promise with an Object containing `data` and `info` properties instead of resolving only with `data`.
+ * @param {Function} [callback]
+ * @returns {Promise<Buffer>} - when no callback is provided
```
lovell, line 141: Reference to `Buffer`

lib/output.js
```diff
@@ -116,6 +116,38 @@ function toBuffer (options, callback) {
   return this._pipeline(is.fn(options) ? options : callback);
 }
 
+/**
+ * Same as toBuffer, but writes output to a Uint8Array.
+ *
+ * @example
+ * sharp(input)
+ *   .toBuffer((err, data, info) => { ... });
```
lovell, line 124: Reference to `toBuffer`

lib/output.js
lovell: I think there might be a lack of error handling here - what happens when `toBuffer` rejects?

lib/input.js
nnmrts: About the copy/mutable thing:
Yes, that's true, and I have to confess that I don't know a lot about "buffers" generally speaking, because I grew up with Javascript and not with languages where you actually **have** to deal with memory and buffers directly all the time. So the first time I came across the "Arraybuffer is just a reference" and "whatever you do to this buffer, also changes things in every reference buffer" kind of thing, it weirded me out, because most of the times things in Javascript don't behave like this. I get that sharing the same allocated memory is intended behaviour, but I still feel like it confuses more than it makes things faster/better. Look at [this small example](https://runkit.com/nnmrts/buffers).

Yada yada yada, I think for sharp, and especially in this scenario, where sharp later passes that buffer on to libvips and then gives us back a new one, the whole copy/reference stuff is irrelevant. We are copying it/creating a new one anyway later on, afaik.

But, A, of course, you could still argue that using the underlying ArrayBuffer is more explicit and B, I often read that this has a lot to do with performance, so I guess using ArrayBuffer is faster? I don't know. But considering all of that, I'll look into how to use ArrayBuffer correctly here, so we don't make useless copies.

About the condition thing:
Sure, I can do that if you really want, but as I said before, node's Buffer is just a subclass of Uint8Array, so this:

```js
const input = Buffer.from([1,2,3,4]); // can be Buffer or uint8array

let result; // should be Buffer

if (input instanceof Buffer) {
	result = buffer;
}
else if (input instanceof Uint8Array) {
	result = Buffer.from(input.buffer);
}
```

is exactly the same as:

```js
const input = Buffer.from([1,2,3,4]); // can be Buffer or uint8array

let result; // should be Buffer

if (input instanceof Uint8Array) {
	result = Buffer.from(input.buffer);
}
```

Using `else if (is.uint8Array(input))` is probably easier to read/follow though, true.

test/unit/toUint8Array.js
nnmrts: Alright, I'll look into that, although this test file is just a copy of the toBuffer.js test file. I don't fully understand the project and testing structure of sharp, so this will probably take me way longer than it would take anyone else with experience of sharp and/or unit tests of a big project, though.

lib/output.js
nnmrts: Same as what I said about the input, but I'll look into it and how to useArrayBuffer here.



> I think there might be a lack of error handling here - what happens when `toBuffer` rejects?

Oh snap! Yep, I forgot to add that in my last commit, working on it.

test/unit/toUint8Array.js
nnmrts: Hm, so I probably should just search for every test that uses `toBuffer` and copy it but with `Uint8Array`, right?

nnmrts:
oof, my approach actually doesn't work at all, because toBuffer can accept and return so different things...

I pretty much have to rewrite the _pipeline function from output.js to make this work.

lovell:
Closing in favour of #2511, thank you for taking the time to submit this PR.

PR #2258, opened 2020-06-15
Add support for ImageMagick version 7

vouillon:
`libvips` uses specific loaders for ImageMagick 7.

### DIFF
--- a/src/common.cc
+++ b/src/common.cc
@@ -197,7 +197,8 @@ namespace sharp {
         imageType = ImageType::HEIF;
       } else if (EndsWith(loader, "PdfBuffer")) {
         imageType = ImageType::PDF;
-      } else if (EndsWith(loader, "MagickBuffer")) {
+      } else if (EndsWith(loader, "MagickBuffer") ||
+                 EndsWith(loader, "Magick7Buffer")) {
         imageType = ImageType::MAGICK;
       }
     }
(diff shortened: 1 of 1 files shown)


lovell:
La magie, merci/thanks!

jcupitt:
This is not very important, but checking for the exact class names feels a little ugly, and could break.

How about checking the nickname instead? The libvips base object has a `nickname` class member:

https://github.com/libvips/libvips/blob/master/libvips/include/vips/object.h#L525

It's also exposed as an object member and glib property on vips objects. The nickname is the thing that's used to look up operations by name, eg.:

```C
GType vips_type_find( const char *basename, const char *nickname );
const char *vips_nickname_find( GType type );
```

`vips_type_find` uses a lazy hash to go from a nickname to a `GType`efficiently.

In this case you could use:

```C
char const *load = file ? vips_foreign_find_load(file) : NULL;
GType gtype = load ? g_type_from_name(load) : 0;
const char *nickname = gtype ? vips_nickname_find(gtype) : NULL;
```

And `nickname` will be something like `magickload` or NULL. libvips guarantees that nicknames are unique and unvarying, ie. they are a supported part of the API (unlike class names, which are not). You could perhaps use a hash to map nicknames to file types and save the switch.

jcupitt:
... I could try a PR, if that would help.

lovell:
@jcupitt Great idea, thank you, I've added this via commit 8da30f0

jcupitt:
Ah, much prettier! Nice!

PR #2259, opened 2020-06-15
Allow page selection when loading a file through ImageMagick

vouillon:
ICO files, for instance, may have multiple pages, which can be selected when loading a file through ImageMagick.


lovell:
Un autre commit magique, merci beaucoup.

PR #2280, opened 2020-07-04
Add correlate operation based on libvips spcor and fastcor.

apodolny:


lovell:
Thank you for the PR, it'll be a little while before I can get around to reviewing this properly, but in the interim please can you remove the yarn lockfile and unpin semistandard (I don't think the newline-related failure affects npm).

apodolny:
Spent a bit more time with this code and looks like there is another issue I need to address. The output from correlation is VIPS_FORMAT_FLOAT and when it is cast VIPS_FORMAT_UCHAR, it loses the precision necessary to actually contain the results.

Two potential options I've considered and I'd welcome your guidance:
1. The real goal of this PR is to enable support for something like matchTemplate in OpenCV, which is basically correlate + min. I could add a function similar to stats() that runs correlate and returns the min directly.
2. Modify pipeline.cc so that it doesn't cast the output to VIPS_FORMAT_UCHAR in the following code:
```          if (image.format() != VIPS_FORMAT_UCHAR) {
            // Cast pixels to uint8 (unsigned char)
            image = image.cast(VIPS_FORMAT_UCHAR);
          }```
3. Add some output option to enable (2) above.

Separately, semistandard does need to be pinned until the bug I filed is fix. See this TravisCI failure for sharp in a MacOS environment: https://travis-ci.org/github/lovell/sharp/jobs/705021697

Thanks!

PR #2335, opened 2020-08-25
Default extend directions to 0

Pulau-Komodo:
Allow any extend direction to be omitted (falsey) for a default of 0.

### DIFF
--- a/lib/resize.js
+++ b/lib/resize.js
@@ -319,15 +319,15 @@ function extend (extend) {
     this.options.extendRight = extend;
   } else if (
     is.object(extend) &&
-    is.integer(extend.top) && extend.top >= 0 &&
-    is.integer(extend.bottom) && extend.bottom >= 0 &&
-    is.integer(extend.left) && extend.left >= 0 &&
-    is.integer(extend.right) && extend.right >= 0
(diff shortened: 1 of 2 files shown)


lovell (changes requested):
Hello, thanks for the PR. I like the idea of defaulting these to zero. I think it will be better to allow each of the top/bottom/left/right parameters to be set independently of each other (so each parameter gets its own if statement), rather than the current approach of having to set all of them. Please can you add unit test coverage for all changes too.

Pulau-Komodo:
The way I wrote it, every parameter has to be falsey _or_ valid, and then the falsey ones default to 0. So it already achieves independent setting without individual if statements. You may prefer individual if statements stylistically, though.

I changed it, but I may have committed several new style violations in the process.

Also, my new pull request requires parameters to be undefined or valid, and throws more specific errors.

I don't know how to add unit test coverage. This was my first pull request.

lovell:
Have a look in https://github.com/lovell/sharp/blob/master/test/unit/extend.js for existing unit tests that cover the `extend` operation. When you run `npm test` locally you'll see a `coverage` directory appear with full HTML reports on which new code branches require test coverage.

Issue #1421, opened 2018-10-19
SVG to PNG conversion produces pixelated output with height and width values


nikwen:
When resizing svg files and converting them to png, the output is really pixelated if

* the svg file has a height and a width specified in pixels and
* the size it should be resized to is larger than these values.

I would expect svgs to be resized properly, no matter if a height and a width are specified.

## Sample project

https://github.com/nikwen/sharp-demo/tree/svg-resize-height-width (make sure you are on the `svg-resize-height-width` branch)

## Sample image conversion

Source (32x32px): https://github.com/nikwen/sharp-demo/blob/svg-resize-height-width/image.svg
Result (512x512px): https://github.com/nikwen/sharp-demo/blob/svg-resize-height-width/out.png

lovell:
Hello, did you see the [density](http://sharp.pixelplumbing.com/en/stable/api-constructor/#sharp) property to set the DPI at which to render vector images?

```diff
- sharp('./image.svg')
+ sharp('./image.svg', { density: 300 })
```

nikwen:
Thank you, Lovell! Setting a density makes the output file look a lot better. For the demo, however, 300 is by far not enough and I need to increase it to 800 or something in that range to make it somewhat sharp. I attached an image of how it looks like with a density of 300.

So to make sure that my app works fine with all SVGs, I need to set a high enough density. Therefore, I need to set it to the max value: 2400.

Please have a look at the [sample project](https://github.com/nikwen/sharp-demo/tree/svg-resize-height-width) again, where I set things to max density. I am still able to find a sample image that produces [a pixelated result](https://raw.githubusercontent.com/nikwen/sharp-demo/ac3e3a3d84de19f5d41b3779baab6e0344829ec2/out.png).

Now, when I remove the `width` and `height` properties of the SVG, I get a perfectly fine result. With them, stuff is pixelated.

In this case, one can argue that setting such properties (especially to such low values) is bad practice and I'd agree. However, I just recently received an SVG with size 16x16px from a designer who used Adobe Illustrator to export it and I believe the library should do its best to render the SVG at the greates possible (but still reasonable) quality. A simple solution would be to ignore `height` and `width` pixel values and only use them to calculate the image's aspect ratio.

That should produce good results for every image.

If you do not want to implement that or know corner cases where this would break stuff, feel free to close this issue. I just stumbled across it and thought others might run into the same trap as me.

Thank you for all your time!

<br>

**Image at a density of 300:**

[...]

PR #2348, opened 2020-09-01
fix: allow density values above 2400 dpi

stefanprobst:
currently, `density` is capped at 2400dpi. when e.g. converting svg images with a small viewBox or small width/height values  to png images, this can be not enough to produce crisp results.

according to [this comment](https://github.com/lovell/sharp/issues/1421#issuecomment-493950250) `librsvg` supports values up to 100000, which is what this PR allows.

my main motivation here is actually not even image quality but file size, i'm seeing 100243 bytes with the density cap, vs 22262 bytes without.

(not sure though if this has any consequences i'm not aware of, or what the original reason for the 2400dpi cap was)


lovell:
Thank you very much!

PR #2369, opened 2020-09-15
fix: Animated GIF to WEBP using a Sharp stream is not animated anymore

### LINKED ISSUE #2362: Animated GIF to WEBP using a Sharp stream is not animated anymore
**The installation went well, and I managed to make it work with a file or an URL for the input and a file for the output, but i don't manage to find a way to get it animated when using a stream for the output.**

@deftomat Sorry for bothering you, but I think you might have a quick answer 👼 

Are you using the latest version? Is the version currently in use as reported by `npm ls sharp` the same as the latest version as reported by `npm view sharp dist-tags.latest`?
Yes

What are the steps to reproduce?
```javascript
const app = new Koa();
const router = new Router();
const agent = require('superagent');
const sharp = require('sharp');

router.get('/:spec/i.:extension', ctx => {
  let sharpStream = sharp() // WORKING BUT NOT ANIMATED
  let sharpStream = sharp({ animated: true }) // Error [ERR_STREAM_DESTROYED]: Cannot call write after a stream was destroyed
  let sharpStream = sharp({ pages: -1 }) // Error [ERR_STREAM_DESTROYED]: Cannot call write after a stream was destroyed
  let sharpStream = sharp(undefined, { pages: -1 }) // Error: Unsupported input 'undefined' of type undefined when also providing options of type object
  let sharpStream = sharp(undefined, { animated: true }) // Error: Unsupported input 'undefined' of type undefined when also providing options of type object

  sharpStream = sharpStream.toFormat('webp', {
    quality,
    pages: -1, // doesn't change anything
    animated: true, // doesn't change anything
  });

  ctx.set('Content-Type', 'image/webp');
  ctx.body = agent.get(url).pipe(sharpStream);
  ctx.status = 200;
});
```
What is the expected behaviour?
The output stream webp image should be animated.

Are you able to provide a minimal, standalone code sample, without other dependencies, that demonstrates this problem?
Kind of, please see below.

Are you able to provide a sample image that helps explain the problem?
I used [this image](https://giffiles.alphacoders.com/386/3869.gif).

What is the output of running `npx envinfo --binaries --system`?
```shell
  System:
    OS: macOS 10.15.6
    CPU: (16) x64 Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
    Memory: 12.02 GB / 32.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 10.22.0 - ~/.nvm/versions/node/v10.22.0/bin/node
    Yarn: 1.22.5 - /usr/local/bin/yarn
    npm: 6.14.8 - ~/.nvm/versions/node/v10.22.0/bin/npm
```

AcrylicShrimp:
Fixes #2362.

### DIFF
--- a/test/unit/gif.js
+++ b/test/unit/gif.js
@@ -98,4 +98,36 @@ describe('GIF input', () => {
       sharp().gif({ delay: [65536] });
     });
   });
+
+  it('should work with streams when only animated is set', function (done) {
+    if (sharp.format.magick.output.buffer) {
+      fs.createReadStream(fixtures.inputGifAnimated)
+        .pipe(sharp({ animated: true }))
+        .gif()
+        .toBuffer(function (err, data, info) {
+          if (err) throw err;
+          assert.strictEqual(true, data.length > 0);
+          assert.strictEqual('gif', info.format);
+          fixtures.assertSimilar(fixtures.inputGifAnimated, data, done);
+        });
+    } else {
+      done();
+    }
+  });
+
+  it('should work with streams when only pages is set', function (done) {
+    if (sharp.format.magick.output.buffer) {
+      fs.createReadStream(fixtures.inputGifAnimated)
+        .pipe(sharp({ pages: -1 }))
+        .gif()
+        .toBuffer(function (err, data, info) {
+          if (err) throw err;
+          assert.strictEqual(true, data.length > 0);
+          assert.strictEqual('gif', info.format);
+          fixtures.assertSimilar(fixtures.inputGifAnimated, data, done);
+        });
+    } else {
+      done();
+    }
+  });
 });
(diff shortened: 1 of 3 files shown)


test/unit/gif.js
```diff
@@ -98,4 +98,28 @@ describe('GIF input', () => {
       sharp().gif({ delay: [65536] });
     });
   });
+
+  it('should work with streams when only animated is set', function (done) {
```
lovell, line 102: Not all test environments will have *magick available - please can you wrap these two GIF output tests with:
```js
if (sharp.format.magick.output.buffer) { ... }
```

lovell (changes requested):
Thank you very much for this PR. There's one small comment about the tests that will need addressing but otherwise this is good to merge.

AcrylicShrimp:
Thank you for feedback! I applied what you said, could you review it's ok for now?

lovell (approved):
Thanks for the update!

PR #2379, opened 2020-09-22
Avoid install script crashing on Yarn

jalovatt:
The version check added in https://github.com/lovell/sharp/commit/d406cb619c9a9db336f138b58bfde267e573e1ce expects `process.env.npm_package_engines_node` to exist. Yarn 2 (not sure about v1) doesn't expose fields from `package.json` like `npm` does, so the build instantly fails, rendering 0.26.1 uninstallable.

I've added a fallback that reads and parses `package.json` directly if the expected variable isn't found.

### DIFF
--- a/install/libvips.js
+++ b/install/libvips.js
@@ -56,13 +56,6 @@ const extractTarball = function (tarPath) {
   );
 };
 
-const getSupportedNodeVersion = () => {
-  const rawPackage = fs.readFileSync(path.join(__dirname, '..', 'package.json'), { encoding: 'utf8' });
-  const packageObj = JSON.parse(rawPackage);
-
-  return packageObj && packageObj.engines.node;
-};
-
 try {
   const useGlobalLibvips = libvips.useGlobalLibvips();
   if (useGlobalLibvips) {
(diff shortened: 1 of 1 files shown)


lovell:
Thanks for the PR. Could we perhaps simplify this with something like the following?
```diff
- process.env.npm_package_engines_node
+ process.env.npm_package_engines_node || require('../package.json').engines.node
```

jalovatt:
> Thanks for the PR. Could we perhaps simplify this with something like the following?
> 
> ```diff
> - process.env.npm_package_engines_node
> + process.env.npm_package_engines_node || require('../package.json').engines.node
> ```

Done. I've been working with ES modules for so long that I forgot we could pull JSON files in that way; good call. :)

install/libvips.js
```diff
@@ -80,9 +80,14 @@ try {
         throw new Error(`Use with glibc ${detectLibc.version} requires manual installation of libvips >= ${minimumLibvipsVersion}`);
       }
     }
-    if (!semver.satisfies(process.versions.node, process.env.npm_package_engines_node)) {
-      throw new Error(`Expected Node.js version ${process.env.npm_package_engines_node} but found ${process.versions.node}`);
+    const supportedNodeVersion = process.env.npm_package_engines_node || require('../package.json').engines.node;
+    if (!supportedNodeVersion) {
```
lovell, line 84: Can the code inside this `if` block ever be reached given `require` will throw an error if there's a problem?

jalovatt: Now that you mention it, `engines` would have to be defined but not `engines.node`; that probably wouldn't happen by accident. Happy to change it if you don't think it's worth the extra check + error message.

lovell: Yes please to removing this check.

jalovatt: :+1:

lovell:
Thank you!

PR #2391, opened 2020-10-01
Allow for negative top/left offsets in composite overlays, updated

CurosMJ:
Composite overlays can be negatively offset, as requested in #1160.

Potentially breaking changes:

- A top or left offset value of -1 will no longer mean that the value is not set, but will now be an actual offset of -1
- INT_MIN is now used as the special value instead of -1

Co-authored-by: Christian Flintrup <chr@gigahost.dk>

This PR extends the work done in #2071


test/unit/composite.js
```diff
@@ -333,25 +351,13 @@ describe('composite', () => {
...
+      }, /Expected integer for left but received 0.5 of type number/);
     });
 
     it('invalid top', () => {
       assert.throws(() => {
-        sharp().composite([{ input: 'test', top: -1 }]);
-      }, /Expected positive integer for top but received -1 of type number/);
-    });
-
-    it('left but no top', () => {
```
lovell, line 345: The "left but no top" and "top but no left" tests have been removed; is this intentional?

src/pipeline.h
```diff
@@ -47,8 +47,8 @@ struct Composite {
     input(nullptr),
     mode(VIPS_BLEND_MODE_OVER),
     gravity(0),
-    left(-1),
-    top(-1),
+    left(INT_MIN),
```
lovell, line 50: `INT_MIN` is -32767, which could still be a valid offset. Perhaps add a `composite->hasOffset` boolean instead? This could be set in JavaScript, removing the need for the `IsUndefined` check in C++.

lovell (changes requested):
Thank you very much for updating this PR Manan, I've left a couple of questions inline.

src/pipeline.h
CurosMJ: Thanks for the suggestion - just implemented it!

test/unit/composite.js
CurosMJ: Added the test cases back :)

lovell (approved):
Thank you for the updates Manan, this is ready to merge. Given it contains a slightly breaking change it's probably best to wait and include this in v0.27.0, so please bear with me whilst I get v0.26.2 out of the door first.

master117:
Will this also allow top/left that would move the image partially outside on the right or bottom?

lovell:
Landed as https://github.com/lovell/sharp/commit/02676140e8c0c86471548a98c49389876abb9ff2 - thanks all, this will be in v0.27.0

PR #2397, opened 2020-10-04
tile center

beig:
added center/centre image in tile


lib/output.js
```diff
@@ -726,6 +727,14 @@ function tile (options) {
     } else if (is.defined(options.layout) && options.layout === 'google') {
       this.options.tileSkipBlanks = 5;
     }
+    // Center image in tile
+    if (is.defined(options.centre) || is.defined(options.center)) {
+      if (is.bool(options.centre) || is.bool(options.center)) {
+        this.options.centre = options.centre || options.center;
```
lovell, line 733: Please can you use `this._setBooleanOption()` to set this boolean property, preferably with a `tile` prefix for consistency, i.e. `tileCentre`. You may also want to set it to a default value of `false` when `this.options` is created in the constructor.

src/pipeline.cc
```diff
@@ -1404,6 +1405,10 @@ Napi::Value pipeline(const Napi::CallbackInfo& info) {
     vips_enum_from_nick(nullptr, VIPS_TYPE_FOREIGN_DZ_DEPTH,
     sharp::AttrAsStr(options, "tileDepth").data()));
 
+  if (sharp::HasAttr(options, "centre")) {
+    baton->centre = sharp::AttrAsBool(options, "centre");
```
lovell, line 1409: The `if` wrapper can be removed by ensuring `this.options.tileCentre` defaults to `false` in JS land.

lovell (changes requested):
Hi, thank you for this PR, I've left a couple of comments inline that should help you with those failing tests.

beig:
I hopefully correctly applied your suggestions

lovell:
Thank you for the update, it looks like there is a linting error to deal with.

lovell:
Brilliant, thank you very much, this will be in v0.26.2.

PR #2400, opened 2020-10-05
Docs: Clarify response object  description for .trim(). Closes #2396.

### LINKED ISSUE #2396: trim(): how to access offsets?
I am trying to use the `.trim()` method to get starting offsets that I can adjust based on other factors (e.g. adding additional padding, or creating a cumulative bounding box based on how multiple images get trimmed).

[The docs say](https://sharp.pixelplumbing.com/api-resize#trim) that "The `info` response Object will contain `trimOffsetLeft` and `trimOffsetTop` properties.", however I cannot find a way to access this "info response object", as the return value is the Sharp instance and none of its properties seem to contain a field call "info" or a field with those trimOffset names.

I was unable to find an answer to this here or on Stack overflow, though that may be because the keywords involved are all very generic.

adam-coster:


lovell:
Brilliant, thank you!

PR #2412, opened 2020-10-15
Conditionally download Brotli or gzip

### LINKED ISSUE #2410: conditionally download gzip or brotli libvips based on node version
What are you trying to achieve?
Gatsby has set its engine to 10.13.0. We would love to move to 0.26.0, technically we could bump a major to achieve this. Would you be okay to have a node version check that fallback to gz download on < 10.16.0.

ascorbic:
This PR conditionally downloads the libvips tarball in Brotli or gzip format, depending on whether the user's version of Node supports Brotli decompression. This allows the minimum Node version to revert to 10.

Fixes #2410

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Lanczos resampling ensures quality is not sacrificed for speed.
 As well as image resizing, operations such as
 rotation, extraction, compositing and gamma correction are available.
 
-Most modern macOS, Windows and Linux systems running Node.js v10.16.0+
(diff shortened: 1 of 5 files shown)


lovell (approved):
Great, thank you. As this PR updates the supported version in the README, which some people use as documentation, I'm going to wait and merge this just before the next release.

lovell:
Thanks again, this will be in v0.26.3

lovell:
v0.26.3 now available

PR #2418, opened 2020-10-23
Set up GitHub Actions CI

nstepien:
AppVeyor and Travis CI can be really slow, with GitHub Actions CI jobs can run in parallel and without much delay, WDYT?

image

- You'll probably need to set up the `prebuild_upload` token for this in the repo's settings.
- I've separated linting into its own job since `cpplint` was silently failing on Windows, and it doesn't seem like it needs to run on all platforms.
- The Debian builds are failing, not sure how to fix these.
  - You can check builds on my fork: https://github.com/nstepien/sharp/actions/runs/324453211
- One issue is that 32 bit Windows is not available. Maybe AppVeyor should be kept only for this specific platform combination?
- I have not set up code coverage reporting from Actions, maybe we can do this later.
- Edits are allowed on my PRs, so feel free to play around with it, feel free to ask me any questions as well.

https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions
https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners


.github/workflows/ci.yml
```diff
@@ -0,0 +1,115 @@
...
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v1
+      with:
+        node-version: ${{ matrix.node }}
+    - name: Install dependencies
+      run: npm install
+    - name: Unit tests
+      run: npm run test-unit
+    - name: Prebuild
+      if: matrix.node == 10 && env.prebuild_upload != '' && startsWith(github.ref, 'refs/tags/')
```
nstepien, line 22: We probably don't need to check for the `prebuild_upload` token, if we already check for the node version and tag.

lovell:
Hi, thanks for this PR Nicolas. The slowest CI environments are currently Windows (Appveyor) and macOS (Travis), which should probably be the priorities for any speed improvement.

GitHub Actions doesn't support arm64, hence the failed tasks you're seeing. I think I'd prefer to keep all the Linux (amd64 and arm64) CI running on Travis for now as that part of it is relatively fast.

It looks like GitHub Actions may soon have support for 32-bit Node.js on Windows - see https://github.com/actions/setup-node/pull/194 - we should wait for this as I'd prefer to have all the Windows CI tasks running on the same service.

Perhaps we could focus initially on migrating macOS to GitHub Actions?

nstepien:
>The slowest CI environments are currently Windows (Appveyor) and macOS (Travis)
>I think I'd prefer to keep all the Linux (amd64 and arm64) CI running on Travis for now as that part of it is relatively fast.

https://www.traviscistatus.com/#month
I was under the impression that it was the opposite, as Travis CI can have a lot of backlogged builds on Linux.

>Perhaps we could focus initially on migrating macOS to GitHub Actions?

Sure, I'll tweak it.

nstepien:
CI run: https://github.com/nstepien/sharp/actions/runs/327307286

lovell:
Thanks for the updates. I'd like to wait until after v0.27.0 before making this change, probably be mid-November, so will come back to this soon.

aminya:
> It looks like GitHub Actions may soon have support for 32-bit Node.js on Windows - see [actions/setup-node#194](https://github.com/actions/setup-node/pull/194) - we should wait for this as I'd prefer to have all the Windows CI tasks running on the same service.
> 

Because @GitHub has become slow in reviewing that PR, I have tagged it on my fork. You can use it now if you want:
https://github.com/atom-ide-community/fuzzaldrin-plus-fast/blob/a62422a04ee9165763f86510b6728e26f789629a/.github/workflows/CI.yml#L42

lovell:
Rather coincidentally, it looks like we're going to have to say "so long and thanks for all the fish" to Travis CI before the end of the year.

https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing

lovell:
I've migrated another native module, farmhash, away from Travis to now use GitHub Actions.

The 16-way matrix/config for Linux glibc (Centos 7), Linux musl (Alpine 3.11), macOS and Windows x64 can be seen at https://github.com/lovell/farmhash/blob/master/.github/workflows/ci.yml

It still uses Appveyor for Windows x86 - see https://github.com/lovell/farmhash/blob/master/appveyor.yml

The generation of prebuilds are split into their own conditional step, but the use of `cpplint` remains part of the testing step as the latest version now works on Windows (it was a Python 2 vs 3 thing).

The farmhash module doesn't support ARM, so I've not had a chance to test out the QEMU-based https://github.com/uraimo/run-on-arch-action yet.

lovell:
Commit https://github.com/lovell/sharp/commit/fb07c1f4695f6bad40d08cd4dbc10f614c59137c on the `actions` branch (which is branched from the `avif` v0.27.0 branch) takes the excellent work in this PR as a starting point and adds GitHub Actions config for x64 Linux, macOS and Windows - see https://github.com/lovell/sharp/actions/runs/410589336 for a successful run.

Please so let me know if you see any problems/improvements.

It still uses Travis for ARM64 (as it's much faster than QEMU and there's never a queue for this infrastructure) - see https://travis-ci.org/github/lovell/sharp/builds/748541551

It still uses Appveyor for x86 Windows - see https://ci.appveyor.com/project/lovell/sharp/builds/36740068

There's one very minor problem where Coveralls is missing the commit message, but that could well be a feature/bug of the Coveralls Action, and I'm happy to ignore that - see https://coveralls.io/builds/35568878

nstepien:
>Please so let me know if you see any problems/improvements.

LGTM

>It still uses Appveyor for x86 Windows

I wonder if this even worth supporting, it must be a really tiny portion of your users. There are no supported 32 bit windows server versions anymore.
https://en.wikipedia.org/wiki/Comparison_of_Microsoft_Windows_versions#Windows_NT
I guess it doesn't really hurt to have.

>Coveralls is missing the commit message

¯\\\_(ツ)_/¯ Should open an issue https://github.com/coverallsapp/github-action/issues

lovell:
Thanks for taking a look (and thank you for opening this PR in the first place), I plan to make the switch before v0.27.0.

The download stats suggest a 10:1 ratio of 64 to 32-bit Windows users. I think Azure only offers 32-bit Windows VMs on their free/basic tier(s) and I guess there will be people who install the 32-bit `node.exe` on 64-bit machines.

lovell:
Commit https://github.com/lovell/sharp/commit/e59e14688707215233bfb11e46cb9b58b2705d04 on the `avif` branch migrates as many environments as possible to Github Actions, which will be used to create binaries for v0.27.0.

@nstepien Nicolas, thank you for doing _la préparation_ for this. When we are able to, let's meet in London for a glass of wine, _c'est ma tournée_.

nstepien:
🎉 I only drink Orangina tho.

lovell:
Très bon, moins cher.

Issue #2489, opened 2020-12-19
Reducing overall package size


chronoDave:
**What are you trying to achieve?**

Reducing the overall (production) package size

**Have you searched for similar feature requests?**

[Yes, I have](https://github.com/lovell/sharp/issues?q=label%3Aenhancement+package+size)

**What would you expect the API to look like?**

Currently, `sharp` requires **68** external packages (70 excluding `.bin` and `.yarn-integrity`). This can be reproduced using `yarn add sharp --prod`
 
 ![node_modules](https://i.imgur.com/Lft6BNK.png)

Packages that can (probably) be moved and / or replaced

- `array-flatten`: Can be replaced by the native Node [`Array.flat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flat) (requires node version >`11.0.0`)
- `detect-libc`: Not entirely sure if this package is required to _run_ `sharp`, but if this is the case, this could be moved to `devDependencies`
- `node-addon-api`: Not entirely sure if this package is required to _run_ `sharp`, but if this is the case, this could be moved to `devDependencies`. From my understanding, this package is only used to build native addons (such as `sharp`)
- `npmlog`: It's a logger, this probably should be moved to `devDepedencies`.
- `prebuild-install`: From my understanding, this package is used to install prebuild binaries, which probably works together with `node-addon-api`. Probably also should be moved to `devDepedencies`, if possible.
- `semver`: Package version parser, used by `libvips.js`. From my understanding, also used to install `sharp`, so probably should be a `devDepedency`
- `simple-get`: This can be replaced by using Node's built-in `http`:

```JS
libvips.js 104

const http = require('http')

http.get(url, { agent: agent() }, function (err, response) {
  // etc.
}
```

- `tar-fs`: This _can_ be replaced by using bash (`tar xzvf <tar.gz>`), but understandable if it must be a JS script. It is possible to execute this command using `child_process`, however. Even still, this probably also should be moved to `devDepedencies`
- `tunnel-agent`: Not entirely sure what this package does, but seems part of the installation toolchain, so can probably be moved to `devDepedencies`

If my assumptions are correct, this means that `sharp` could possibly be a 0 dependency package, which would be amazing!

**What alternatives have you considered?**

[...]

PR #2490, opened 2020-12-20
Deps: Removed 'array-flatten' dependency

chronoDave:
As described in #2489 , `array-flatten` can simply be replaced by Node's native `Array.flat`


chronoDave:
Cannot be added due to Node 10 support

Issue #2500, opened 2020-12-28
tiff float cast


randyridge:
I noticed that in pipeline.cc when saving tiffs to a buffer the following code is run:

```
// Cast pixel values to float, if required
if (baton->tiffPredictor == VIPS_FOREIGN_TIFF_PREDICTOR_FLOAT) {
  image = image.cast(VIPS_FORMAT_FLOAT);
 }
```

However, saving tiff to a file does not have the same cast applied.  Is this by design?

lovell:
No, this looks like a bug, thanks for reporting this. Happy to accept a PR if you're able.

lovell:
v0.27.1 now available, thanks for the PR.

PR #2502, opened 2020-12-29
add missing float cast check

randyridge:
[Issue 2500](https://github.com/lovell/sharp/issues/2500)


randyridge:
It's unclear to me if the failing test "TIFF deflate compression with float predictor shrinks test file" should in fact be shrinking with this change.  There's not a corresponding test related to the float predictor outputting to a buffer for comparison...

lovell:
Thank you for the PR, I agree that the test that has started failing with this change is wrong.

Given the input is integral, use of the float predictor and therefore a cast to floating point values will probably result in a larger file.

Shall we refactor this test as something more like "TIFF deflate compression of integral input with float predictor increases file size"?

lovell:
Landed via commit 0bb8cb9 - thanks for the PR!

PR #2503, opened 2020-12-29
Rr/tiff bitdepth

randyridge:
It appears per [tiff_save](https://libvips.github.io/libvips/API/current/VipsForeignSave.html#vips-tiffsave) that bitdepth here should be 1, 2, or 4, and not 8.  I receive a warning from libvips as such:

``` sh
(NetVips:12484): VIPS-WARNING **: 13:09:07.670: bitdepth 1, 2 or 4 only -- disabling bitdepth
```

### DIFF
--- a/lib/constructor.js
+++ b/lib/constructor.js
@@ -227,7 +227,7 @@ const Sharp = function (input, options) {
     tiffCompression: 'jpeg',
     tiffPredictor: 'horizontal',
     tiffPyramid: false,
-    tiffBitdepth: 8,
+    tiffBitdepth: 4,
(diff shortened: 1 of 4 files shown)


randyridge:
I'm somewhat confused, some of the docs for libvips say 1,2, and 4, the code does seem to allow 8.  I'm unclear why I'm getting a warning from libvips on 8 though.

lovell:
We should probably prevent a value of 8 being considered valid and instead have 0 as the default for `bitdepth`.

lovell:
(It looks like the libvips docs may need updating too.)

PR #2511, opened 2021-01-03
Added support for Uint8Array, Uint8ClampedArray

leon:
This allows us to work with raw pixels data easier.

It is a continuation of: https://github.com/lovell/sharp/pull/2256

Regarding memory efficiency. I've done research and have come up with

```js
inputDescriptor.buffer = Buffer.from(uint8array.buffer);
```
This will create a `Buffer` which points to the underlying Uint8Arrays buffer.
They will share the same memory.

And for the other way around its:
```js
new Uint8Array(data.buffer)
```

### Example
So the full around circle would be:

```js
const input = Uint8Array.from([1, 1, 1, 2, 2, 2]); // or Uint8ClampedArray
const image = sharp(input, {
  raw: {
    width: 2,
    height: 1,
    channels: 3
  }
});
image.toUint8Array().then((output) => {
  assert.deepStrictEqual(input, output);
});
```

I have added both unit tests and documentation for the new feature.
All tests are passing.

Performance wise, the only changed code path is another couple of `instanceof` checks in the input handler, which should not be a problem.

Let me know if I need to do any changes 👍🏻


test/unit/toUint8Array.js
```diff
@@ -0,0 +1,103 @@
...
+    sharp.cache(true);
+  });
+
+  it('should output data as Uint8Array', (done) => {
+    const image = sharp(fixtures.inputJpg320x240);
+    image.toUint8Array().then((data) => {
+      assert.strictEqual(data instanceof Uint8Array, true);
+      assert.strictEqual(data.length, 26641);
+      done();
+    });
```
lovell, line 22: Tests using the callback style (via `done`) that also use Promise chains will need a `.catch(done)` on the chain to ensure the test runner is made aware of any errors.

An alternative is to remove the `done` callback entirely and return the Promise chain.

yarn.lock
```diff
@@ -0,0 +1,7978 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
lovell, line 1: Please can you remove this lockfile.

test/unit/toUint8Array.js
```diff
@@ -0,0 +1,103 @@
...
+    sharp.cache(false);
+  });
+  afterEach(function () {
+    sharp.cache(true);
+  });
+
+  it('should output data as Uint8Array', (done) => {
+    const image = sharp(fixtures.inputJpg320x240);
+    image.toUint8Array().then((data) => {
+      assert.strictEqual(data instanceof Uint8Array, true);
```
lovell, line 19: Node.js `Buffer` objects will always satisfy the `instanceof Uint8Array` check. Perhaps verify `data.constructor.name` instead?

lib/is.js
```diff
@@ -48,6 +48,15 @@ const buffer = function (val) {
   return val instanceof Buffer;
 };
 
+/**
+ * Is this value a Uint8Array or Uint8ClampedArray object?
+ * @private
+ */
+const uint8Array = function (val) {
+  // allow both since Uint8ClampedArray simply clamps the values between 0-255
+  return val instanceof Uint8Array || val instanceof Uint8ClampedArray;
```
lovell, line 57: Node.js `Buffer` objects will always satisfy the `instanceof Uint8Array` check. Perhaps verify `data.constructor.name` instead?

lib/output.js
```diff
@@ -118,6 +118,48 @@ function toBuffer (options, callback) {
...
+ *
+ * @param {Object|Function} [optionsOrCallback] The resolveWithObject option if you want to the promise to include info.
+ * @param {boolean} [options.resolveWithObject] Resolve the Promise with an Object containing `data` and `info` properties instead of resolving only with `data`.
+ * @returns {Promise<Uint8Array>} - when no callback is provided
+ */
+function toUint8Array (optionsOrCallback) {
+  // callback mode
+  if (is.fn(optionsOrCallback)) {
+    // wrap callback with function that converts to Uint8Array
+    return this.toBuffer(optionsOrCallback, (err, data, info) => optionsOrCallback(err, new Uint8Array(data.buffer), info));
```
lovell, line 148: Are there too many callbacks here?
```diff
- return this.toBuffer(optionsOrCallback, (err, data, info) => optionsOrCallback(err, new Uint8Array(data.buffer), info));
+ return this.toBuffer((err, data, info) => optionsOrCallback(err, new Uint8Array(data.buffer), info));
```

lovell (changes requested):
Hi Leon, thank you for this PR, I've left a few questions/comments inline.

yarn.lock
leon: Removed it.

test/unit/toUint8Array.js
leon: Used your proposed way, worked like a charm

lib/output.js
leon: Don't know how I missed that. good catch.

leon:
# Question, advice?

Should we discard the `toUint8Array()` function?

Having looked at it for a bit, maybe it's not needed.
At first I thought it would be a good idea to match the input and output apis.
But since we won't be able to output to both Uint8Array and Uint8ClampedArray, maybe it's better if I just update the documentation for toBuffer() to show how to easily create a Uint8Array or Uint8ClampedArray from the returned Buffer?

What do you think @lovell ?

lovell:
Great, happy to keep it simple; the addition of real world examples to the documentation is always a good thing, thank you.

There's value in supporting Uint8Array (and Uint8ClampedArray) on the _input_ side so please can we keep this bit.

leon:
Did the changes we discussed.
It was the right call.
It is better to inform the users of how to work with Buffers and Uint8Arrays.

lib/output.js
```diff
@@ -104,6 +104,20 @@ function toFile (fileOut, callback) {
  *   .then(({ data, info }) => { ... })
  *   .catch(err => { ... });
  *
+ * @example
+ * const data = await sharp('my-image.jpg')
+ *   // output the raw pixels
+ *   .raw()
+ *   .toBuffer();
+ *
```
lovell, line 112: /__w/sharp/sharp/lib/output.js:112:3: Trailing spaces not allowed.

lib/output.js
```diff
@@ -104,6 +104,20 @@ function toFile (fileOut, callback) {
...
+ * const data = await sharp('my-image.jpg')
+ *   // output the raw pixels
+ *   .raw()
+ *   .toBuffer();
+ * 
+ * // create a more type safe way to work with the raw pixel data
+ * // this will not copy the data, instead it will change `data`s underlying ArrayBuffer
+ * // so `data` and `pixelArray` point to the same memory location
+ * const pixelArray = new Uint8ClampedArray(data.buffer);
+ *
```
lovell, line 117: /__w/sharp/sharp/lib/output.js:117:3: Trailing spaces not allowed.

lovell (changes requested):
Thanks for the updates, there are a couple of lines failing the linter checks. Please can you fix these.

leon:
Sorry, fixed it now

lovell (approved):
Brilliant, thank you.

nnmrts:
Wow @leon, I'm actually really happy to see this, thank you and @lovell. Feels great to see something I struggled with and couldn't finish, actually being finished. Awesome. This was the last piece for my project to go completely node.js-buffer-free (on my end at least). Really hyped to try this out when the new release is here. :heart:

leon:
@nnmrts I'm happy you liked it 😃

PR #2513, opened 2021-01-04
Convert to jest testing

leon:
While I was developing the Uint8Array feature, I was having troubles with mocha.
It would hang completely it any of the assert statements where wrong, and it would take quite long to run all the tests.

I've been using jest for our server side developments at work and it's speed and ease of use has really helped a lot.

I thought I could contribute by implementing it in sharp.

The initial tests are encouraging. 

## using nyc and mocha
running
```bash
time npm test

npm test  107.23s user 86.01s system 356% cpu 54.139 total
```
It took 54s to run

## using jest
```bash
time npm test

npm test  170.19s user 97.10s system 884% cpu 30.221 total
```
It takes 30s to run
and if I only run `jest` it takes 19s

## Going forward.
This was just a proof of concept, and I had to convert some assert statements to get all tests to pass.
Converting to use jests expect syntax instead of assert would improve readability in some tests but...
We don't have to convert them, jest works just fine with any assertion library.
It also is a larger amount of work to do this, but it could be something for the future.

### another suggested improvement: Prettier
Something we could also do to improve things would be to switch to use prettier and configure automatic code formatting.
The automatic code formatting in VS Code works great and has made it so much easier for our team to keep the same look and feel to the code.

Let me know what you think.
// Leon


lovell:
> It would hang completely it any of the assert statements where wrong

Does this relate to #2511 ? I notice the new tests added in that PR are using callback style but missing error handlers, which would probably cause this rather than the test runner.

> and it would take quite long to run all the tests.

The unit tests are currently not designed for parallel running so won't be able to benefit from the suggested performance benefits of jest right now.

Mocha already allows for possible parallel running, so any effort to improve the tests should probably be spent removing things that prevent parallel running e.g. there are currently tests that output to the same filesystem location, which will trample on each other.

> another suggested improvement: Prettier
> The automatic code formatting in VS Code works great 

sharp uses `semistandard` and there's a vscode plugin for this at https://marketplace.visualstudio.com/items?itemName=chenxsan.vscode-standardjs

leon:
Yes I think that was part of the problem. got it running better now.

Yes it would probably need to be re done a bit to get it working 100%.

I still have a lot to learn regarding testing, but it was a fun exercise to tackle a real life project.
But as you say there are a lot of other things to take into account, so I'll give up for now.

Thanks for the feedback 👍🏻

Issue #2345, opened 2020-08-28
Gaussian Noise in sharp


jbveepee:
I'm trying to add gaussian noise an image but can't find the function in sharp.js.
It exists in libvips, but can it also be added to / accessed from sharp.js?

I'm pretty new here and still finding my way, so please do redirect me if needed. Thanks in advance!

lovell:
Hi, did you see #725 ?

jbveepee:
Hi Mr. Lovell, thanks for your reply. Would it be better to reply on 725?
Again, new to git... Apologies if this is bad form.

I came across 725 Thursday but did not know what to do with the information. At this point I'm only trying to get conclusive information on whether it is currently possible. 725 states a dependency on #470. (Resolved since 01/04/2017) and is marked as an enhancement mentioning it needs exposing from libvips.

If it's not possible yet, I'm happy to try and contribute though I do not know C... Willing to give it a good old try though. :-)

All the best!

lovell:
#725 is a future possible enhancement. PRs are always welcome If you're able to help.

I'll close this issue - let's continue to track this at #725.

PR #2525, opened 2021-01-09
feat: exposing vips_gaussnoise operation

### LINKED ISSUE #725: could  add a new function noise() like imageMagick provided?
could  add a new function noise() like imageMagick provided?
thanks.

alza54:
Fixes #725 and #2345 

Can be composited with an image, but background color needs to be added, because libvips *gaussnoise* operation relies on multiband colorspace with only 1 channel.

### Noise generation:

Noise generation

#### Before

Noise generation

#### After
Noise generation


lovell:
Hello, thank you very for the PR.

Were you aware that vips' `gaussnoise` creates a new image rather than being a pipeline operation that is applied to an existing image?

I imagine the sharp API for noise creation being part of the existing "create" mechanics.

Rather than provide a `background` property for the colour, we'd provide a `noise` property, something like:

```js
// PROPOSED API
sharp({
  create: {
    width: 10,
    height: 20,
    channels: 3,
    noise: {
      type: 'gaussian',
      mean: 128,
      sigma: 30
    }
  }
})...
```

If 3 channels are requested then there will be 3 calls to vips' `gaussnoise`. We'll probably the internals of this to cast from float to uchar.

Does this look like something we could do with this PR?

(This approach should also allow us to easily introduce the cell-based Worley and Perlin noises in the future too.)

alza54:
Hello, this is true. I should be able to pull new PR regarding the proposed API

alza54:
Closed in in favour of #2527.

PR #2527, opened 2021-01-12
feat: exposing vips_gaussnoise operation

alza54:
Gaussian noise feature respecting the [proposed API](https://github.com/lovell/sharp/pull/2525#issuecomment-757448918)

```javascript
sharp({
  create: {
    width: 1024,
    height: 768,
    channels: 1,
    noise: {
      type: 'gaussian',
      mean: 128,
      sigma: 30
    }
  }
});
```

### DIFF
--- a/lib/input.js
+++ b/lib/input.js
@@ -137,22 +137,50 @@ function _createInputDescriptor (input, inputOptions, containerOptions) {
         is.object(inputOptions.create) &&
         is.integer(inputOptions.create.width) && inputOptions.create.width > 0 &&
         is.integer(inputOptions.create.height) && inputOptions.create.height > 0 &&
-        is.integer(inputOptions.create.channels) && is.inRange(inputOptions.create.channels, 3, 4) &&
-        is.defined(inputOptions.create.background)
+        is.integer(inputOptions.create.channels)
       ) {
         inputDescriptor.createWidth = inputOptions.create.width;
         inputDescriptor.createHeight = inputOptions.create.height;
         inputDescriptor.createChannels = inputOptions.create.channels;
-        const background = color(inputOptions.create.background);
-        inputDescriptor.createBackground = [
-          background.red(),
-          background.green(),
-          background.blue(),
-          Math.round(background.alpha() * 255)
-        ];
+        // Noise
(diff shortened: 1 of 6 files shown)


src/common.cc
```diff
@@ -318,15 +324,28 @@ namespace sharp {
...
+          image = VImage::new_matrix(descriptor->createWidth, descriptor->createHeight);
+          std::vector<VImage> bands = {};
+          bands.reserve(descriptor->createChannels);
+          for (int _band = 0; _band < descriptor->createChannels; _band++) {
+            bands.push_back(image.gaussnoise(
+              descriptor->createWidth,
+              descriptor->createHeight,
+              VImage::option()->set("mean", descriptor->createNoiseMean)->set("sigma", descriptor->createNoiseSigma)));
+          }
+          image = image.bandjoin(bands).colourspace(VIPS_INTERPRETATION_sRGB).cast(VipsBandFormat::VIPS_FORMAT_UCHAR);
```
alza54, line 337: I supposed that converting to (s)RGB colour space even in case for only 1 channel is an expected behaviour.

lib/input.js
```diff
@@ -137,22 +137,50 @@ function _createInputDescriptor (input, inputOptions, containerOptions) {
...
+            background.blue(),
+            Math.round(background.alpha() * 255)
+          ];
+        } else {
+          throw new Error('Expected valid noise or background to create a new input image');
+        }
         delete inputDescriptor.buffer;
       } else {
-        throw new Error('Expected valid width, height, channels and background to create a new input image');
+        throw new Error('Expected valid width, height and channels to create a new input image');
```
lovell, line 183: Please can you add a unit test that covers this line.

src/common.cc
lovell, line 337: 1 and 2 channel images should probably be `VIPS_INTERPRETATION_B_W` (3 and 4 can be `VIPS_INTERPRETATION_sRGB`).

lovell (changes requested):
Fantastic, thank you very much for this. I've left a couple of questions/comments inline that will probably need addressing, then we can merge this useful feature.

lovell (approved):
Marvellous, thank you very much.

Issue #2535, opened 2021-01-14
[basic_string::_S_construct null not valid] (Electron, Linux, 0.27.0)


gajop:
Sharp: v0.27.0
Electron: v11.1.1

My code is as simple as:

`await sharp('test.png').toFile('test.jpg');`

I get a `[basic_string::_S_construct null not valid]` error when loading a PNG file as described  in https://github.com/lovell/sharp/issues/2356#issuecomment-744371757. I believe the linked comment was never fixed - only the GLib-GObject:ERROR (0.26.0) one was, so I decided to reopen this issue as a new and separate issue.

To summarize:
- 0.27.0 doesn't work with the basic_string error
- 0.26.0 doesn't work with the GLib-GObject:ERROR error
- 0.25.0 seems to work

lovell:
What is the output of running `npx envinfo --binaries --system`?

Please can you provide a minimal repo (code, image and config) that allows someone else to reproduce this problem, including dependency versions in a `package.json`.

lovell:
I think I've been able to reproduce this on Linux using the prebuilt binaries provided by the latest versions of sharp and electron.

It looks like there's problem where strings can become null and it's occurring somewhere between the N-API layer and electron.

I think this is due to nodejs and sharp depending on libstdc++ (which is the right thing to do) but electron does not. 

```sh
$ ldd $(which node) | grep "c++"
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fefa4f3f000)
$ ldd node_modules/sharp/build/Release/sharp.node | grep "c++"
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007efecbdfa000)
$ ldd node_modules/electron/dist/electron | grep "c++"
... crickets ...
```

From what I can tell, the prebuilt binaries provided by electron statically-link the ABI-incompatible libc++ instead but expose API that relies on it.

There appear to be quite a few open issues against electron relating to libc++ vs libstdc++ conflicts:
https://github.com/electron/electron/search?q=%22libc%2B%2B%22&type=issues

I'm unsure what we can do in sharp; supporting electron and its ABI changes has an ongoing headache for years.

gajop:
> What is the output of running npx envinfo --binaries --system?

Unsure if you still want the answer, but:
`npx envinfo --binaries --system?`

```
npx envinfo --binaries --system
npx: installed 1 in 1.194s

  System:
[...]

PR #2551, opened 2021-01-27
Add support for a single shared library (#2535)

kleisauke:
This PR adds supports for a single shared libvips library (i.e. `libvips.so.42` statically linked into `libvips-cpp.so.42`) as described in issue https://github.com/lovell/sharp/issues/2535 and PR https://github.com/lovell/sharp-libvips/pull/81.

Testing can be done with:
```bash
rm ~/.npm/_libvips/libvips-8.10.5-linux-x64.tar.br
SHARP_DIST_BASE_URL="https://github.com/kleisauke/libvips-packaging/releases/download/v8.10.5-single-shared/" \
  npm install kleisauke/sharp#single-shared
```

Depends on: https://github.com/lovell/sharp-libvips/pull/81.

(the last commit must be reverted before merging)

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -140,8 +140,7 @@
             'link_settings': {
               'library_dirs': ['<(sharp_vendor_dir)/lib'],
               'libraries': [
-                'libvips-cpp.42.dylib',
-                'libvips.42.dylib'
+                'libvips-cpp.42.dylib'
               ]
             },
             'xcode_settings': {
@@ -158,8 +157,7 @@
             'link_settings': {
               'library_dirs': ['<(sharp_vendor_dir)/lib'],
               'libraries': [
-                '-l:libvips-cpp.so.42',
-                '-l:libvips.so.42'
+                '-l:libvips-cpp.so.42'
               ],
               'ldflags': [
                 # Ensure runtime linking is relative to sharp.node
--- a/install/libvips.js
+++ b/install/libvips.js
@@ -23,8 +23,8 @@ const minimumGlibcVersionByArch = {
 };
 
 const { minimumLibvipsVersion, minimumLibvipsVersionLabelled } = libvips;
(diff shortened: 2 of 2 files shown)


kleisauke:
Some unit tests on Linux ARM64v8 seems to be failing. This worked yesterday without any problems, see:
https://travis-ci.org/github/kleisauke/sharp/builds/756278183

lovell:
> Some unit tests on Linux ARM64v8 seems to be failing.

I switched to parallel processes for the unit tests but it looks like the ARM64 VMs provided by Travis use LXD and might over-commit CPU and/or RAM, leading to random segfaults. They've got some new Graviton-based VMs but those are only on the paid-for service. I expect Travis to go away fairly soon so it might be time to invest in a Raspberry Pi 4 as a self-hosted runner.

dennisameling:
@lovell you can [apply for free OSS credits](https://travis-ci.com/plans) on the new travis-ci.com, have you tried that already? Then you'd be able to use the new Graviton-based VMs.

image

You can just email them at support [at] travis-ci [dot] com to apply for credits 👍🏼

lovell:
@dennisameling I can try after my Travis account is switched from their .org to .com site, which I have wisely resisted for as long as possible. I've heard from other OSS maintainers that there's nothing available beyond the initial offering, which I estimate would be enough for fewer than 40 builds in total. :sad_panda:

tomByrer:
> invest in a Raspberry Pi 4 as a self-hosted runner

I thought of that myself, but one can find very powerful Android Phones for under $100 (eg new Samsung S01 for $70 at local Wal*Mart) then root & put Linux on it.

lovell:
Landed via commit https://github.com/lovell/sharp/commit/4fc1d0c9007ba6e38edb57f8a0dc5b22de53f31f

Issue #2567, opened 2021-02-08
JSDoc: wrong parameter type for `speed` in AVIF output


aprat84:
For AVIF output, parameter `speed` is typed as `boolean`, but should be `number`, as is in HEIF output.

https://github.com/lovell/sharp/blob/d7d580ae6f9bada7edea941e8f2c0efdc324d793/lib/output.js#L585

Edit: also at documentation:
https://github.com/lovell/sharp/blob/master/docs/api-output.md

lovell:
Good spot, thanks, PR welcome if you're able - please update the JSDoc in the source then run `npm run docs-build` to regenerate the markdown file.

lovell:
Docs updated as part of v0.27.2, thank you.

https://sharp.pixelplumbing.com/api-output#avif

PR #2568, opened 2021-02-08
Docs: correct type for AVIF speed output option

aprat84:
Should be `number`.

Fixes issue: https://github.com/lovell/sharp/issues/2567


lovell:
Brilliant, thank you very much for fixing this.

PR #2581, opened 2021-02-16
Update output.js

florian-busch:
This is my first pull request (at all!). Please let me know if something needs improvement.

I added a const formatLowerCase at line 190 that saves the lowerCase version of the format parameter. 
At 192 and 194 I replaced "format" with "formatLowerCase.

### DIFF
--- a/lib/output.js
+++ b/lib/output.js
@@ -188,9 +188,10 @@ function withMetadata (options) {
  * @throws {Error} unsupported format or options
  */
 function toFormat (format, options) {
-  const actualFormat = formats.get(is.object(format) && is.string(format.id) ? format.id : format);
+  const formatLowerCase = format.toLowerCase()
+  const actualFormat = formats.get(is.object(formatLowerCase) && is.string(formatLowerCase.id) ? formatLowerCase.id : formatLowerCase);
   if (!actualFormat) {
-    throw is.invalidParameterError('format', `one of: ${[...formats.keys()].join(', ')}`, format);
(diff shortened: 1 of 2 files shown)


lib/output.js
```diff
@@ -188,9 +188,10 @@ function withMetadata (options) {
  * @throws {Error} unsupported format or options
  */
 function toFormat (format, options) {
-  const actualFormat = formats.get(is.object(format) && is.string(format.id) ? format.id : format);
+  const formatLowerCase = format.toLowerCase()
```
lovell, line 191: `format` might be an object so the call to `.toLowerCase()` needs to happen on the next line, something like the following (untested):
```diff
- ... formats.get(is.object(format) && is.string(format.id) ? format.id : format);
+ ... formats.get((is.object(format) && is.string(format.id) ? format.id : format).toLowerCase());
```

lovell (changes requested):
Hi, thank you very much for the PR.

Please can you run `npm test` locally to ensure all the linting rules and unit tests still pass. If you could also add a test to cover this new feature e.g. `.toFormat('JPEG')` that would be even better.

I've also left one comment inline that will need addressing.

florian-busch:
Updated output.js to accept upper case characters.
Created test for string and object input toFormat(). 
All tests work locally. Please let me know if other problems occur

lovell (approved):

lovell:
Brilliant, thank you very much.

lovell:
v0.27.2 now available, thanks for the PR.

Issue #2571, opened 2021-02-09
composite: top/left values can be erroneously clipped


SHG42:
What are you trying to achieve?
Extract image data from a 900x700px KonvaJS stage with layered canvases, send to server and composite the layer data buffers with another image data buffer for a sprite that was created on a separate, single-canvas 500x500px Konva stage. On the backend, Sharp receives the most recent buffers from both stages and "sandwiches" the buffer from the smaller stage in between the two buffers from the larger stage. I am attempting to use the absolute position of a sprite on the larger canvas to set the top/left parameters for the composite function.

Sharp returns a buffer which is then passed to an upload method for Cloudinary, which creates the final png file upon upload.

Extra details: the smaller stage ("creator stage") allows for the creation of a game sprite, and the larger stage ("equip stage") allows the user to equip items in front of and behind the sprite to create a scene. On the equip stage, the base sprite is loaded by itself onto a middle canvas layer, and the user can drag it around to position it relative to its items (for instance, standing on top of a fallen log or next to a lake). On form-submit from the equip page, the absolute position of the base sprite on the equip stage is retained in the sprite's MongoDB document to be used as the top/left parameter in the composite function.

The problem: In cases where the base sprite's bounds went past the equip canvas's bottom bound or right-side bound, the sprite's positioning in the composited png is incorrect. To my eye, it looks like the base sprite in the composite is being pushed up and/or left by whatever amount its bounding box has gone past the canvas's bottom or right bounds.

Have you searched for similar questions?
Yes, here and on StackOverflow. The problem is that I can't tell where the issue is coming from. It could be coming from Sharp, KonvaJS, the Cloudinary upload processing, or my code. There are no errors thrown at any point. Console logs at every step along the way haven't been conclusive enough to narrow down what part of the process is at fault. I've searched for solutions relating to each step and none have helped.

Are you able to provide a minimal, standalone code sample that demonstrates this question?
 [This CodeSandbox replicates the functionality of the main project as closely as possible but without database functionality](https://codesandbox.io/s/composite-png-placement-error-4yx3n). If the database could be a factor in the problem, a link can be provided upon request to view the full source code.

Are you able to provide a sample image that helps explain the question?
[Here is a Drive folder containing screenshots](https://drive.google.com/drive/folders/14QyUldaeTvmm_o8eUgL6YrQ4s1hHDl3k?usp=sharing) for all cases.

lovell:
[...]

PR #2594, opened 2021-02-24
Fix erroneous top/left clipping in composite #2571

SHG42:
#2571 Fixes bug where certain input values for top/left parameters
in composite can conflict with clipping logic, resulting in
inaccurate alignment in output png. Adds logic to run check
which determines if clipping needs to be applied.
Issue #2571

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ vendor
 package-lock.json
 .idea
 .firebase
+/.vs
--- a/src/pipeline.cc
+++ b/src/pipeline.cc
@@ -594,8 +594,13 @@ class PipelineWorker : public Napi::AsyncWorker {
           int top;
           if (composite->hasOffset) {
             // Composite image at given offsets
-            std::tie(left, top) = sharp::CalculateCrop(image.width(), image.height(),
-              compositeImage.width(), compositeImage.height(), composite->left, composite->top);
+            if (composite->tile) {
+              std::tie(left, top) = sharp::CalculateCrop(image.width(), image.height(),
+                compositeImage.width(), compositeImage.height(), composite->left, composite->top);
+            } else {
+              left = composite->left;
+              top = composite->top;
+            }
           } else {
             // Composite image with given gravity
             std::tie(left, top) = sharp::CalculateCrop(image.width(), image.height(),
--- a/test/unit/composite.js
+++ b/test/unit/composite.js
@@ -390,4 +390,20 @@ describe('composite', () => {
       }, /Expected valid gravity for gravity but received invalid of type string/);
     });
   });
+
+  it('Allow offset beyond bottom/right edge', async () => {
+    const red = { r: 255, g: 0, b: 0 };
+    const blue = { r: 0, g: 0, b: 255 };
+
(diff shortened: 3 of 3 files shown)


lovell:
Thank you for the PR. If you're able to add a unit test to [test/unit/composite.js](https://github.com/lovell/sharp/blob/master/test/unit/composite.js) with an example scenario this change fixes, that would be amazing.

SHG42:
Hi, sure! I'm not familiar with unit tests yet, so I'll do some reading on it and get back to you when I've taken a crack at it.

SHG42:
Hi, I'm not entirely sure how to tackle this unit test, if you have a moment could you clarify the process for me? I've reviewed the unit test file you linked, and I've done some reading on Javascript unit testing as a concept, but I'm not sure how to apply that to this specific issue, since I only discovered the bug due to a very particular set of circumstances in my project. How would I write a test that demonstrates this fix?

lovell:
Here's a possible test case that checks colour of the top-left pixel after compositing a 2x2 blue over a 2x2 red image at an offset of 1+1.

The current version of sharp will incorrectly clip the offset and produce a top left pixel that is blue, but with the change in this PR it should be corrected to red (as only the bottom right pixel should be blue).

```js
   it('Allow offset beyond bottom/right edge', async () => {
    const red = { r: 255, g: 0, b: 0 };
    const blue = { r: 0, g: 0, b: 255 };

    const [r, g, b] = await sharp({ create: { width: 2, height: 2, channels: 4, background: red } })
      .composite([{
        input: { create: { width: 2, height: 2, channels: 4, background: blue } },
        top: 1,
        left: 1
      }])
      .raw()
      .toBuffer();

    assert.strictEqual(red, { r, g, b });
  });
```
If this works, please feel free to add it as a test case to `test/unit/composite.js`.

SHG42:
Hi, I keep getting a test failure on that. Just to make sure, should I be adding it at a specific line in `test/unit/composite.js`? I tried adding it inside `describe('composite')` after `describe("validation")`, and I also tried it inside `describe("validation")` after `it('invalid gravity')` just to be sure. I also tried both placements with both `npm run test` and `npm run test-unit`. Did I miss a step in the testing process? I'm also curious why this test uses async/await when the rest of the unit tests in the file don't?
This is the error I get: 
```
 AssertionError [ERR_ASSERTION]: Values have same structure but are not reference-equal:

{
  b: 0,
  g: 0,
  r: 255
}
```

lovell:
Sorry, try `deepStrictEqual` rather than `strictEqual` - the docs for these are at https://nodejs.org/dist/latest/docs/api/assert.html

I'm slowly moving tests to using async/await where it makes sense to do so as I find it usually simplifies the flow, makes the intent clearer and offloads more error handling onto the test runner.

SHG42:
Thanks, that worked! Gonna go ahead and add it to the test file.

lovell (approved):
Brilliant, thank you for fixing this, I'll get this merged for v0.28.0.

lovell:
Landed via commit https://github.com/lovell/sharp/commit/34a2e14a144b8768f326751a88c81c0e2d19c59b

PR #2602, opened 2021-03-02
Update the instruction on mirror settings

nevill:
According to the [source](https://github.com/lovell/sharp/blob/master/install/libvips.js#L26-L27), using `npm_config_sharp_libvips_binary_host` is enough to install the binary from mirror host.
In fact, it doesn't work if `sharp_binary_host` is set to `"https://npm.taobao.org/mirrors/sharp"` (should be `"https://npm.taobao.org/mirrors/sharp-libvips/v8.10.0/"`).

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -130,7 +130,6 @@ Alibaba provide a mirror site based in China containing binaries for both sharp 
 To use this either set the following configuration:
 
 ```sh
-npm config set sharp_binary_host "https://npm.taobao.org/mirrors/sharp"
 npm config set sharp_libvips_binary_host "https://npm.taobao.org/mirrors/sharp-libvips"
 npm install sharp
 ```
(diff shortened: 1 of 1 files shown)


lovell:
Hi, there are potentially two binaries downloaded at install time, one for sharp and one for libvips.

https://sharp.pixelplumbing.com/install#chinese-mirror

> Alibaba provide a mirror site based in China containing binaries for both sharp and libvips.

The change in this PR removes use of the mirror for the sharp binary. Is this intentional?

lovell:
I've just tested the documented approach and it works for me. Please ensure you're using the latest version.

nevill:
Oh well, I see. 
I was thinking the problem here is the false document.

Because config variable `npm_config_sharp_binary_host` is not used at all according to the  [source](https://github.com/lovell/sharp/blob/master/install/libvips.js#L26-L27) (the latest).

```javascript
const distBaseUrl = process.env.npm_config_sharp_dist_base_url || process.env.SHARP_DIST_BASE_URL || `${distHost}/v${minimumLibvipsVersionLabelled}/`;
```

> Hi, there are potentially two binaries downloaded at install time, one for sharp and one for libvips.

Then, I realized that another binary is downloaded via package `prebuild-install`. So, `npm_config_sharp_binary_host` is considered. 😃

PR #2608, opened 2021-03-04
Install: fail on incomplete download and clean up tempfile

abradley:
The libvips install script should fail when the connection closes before the response is complete. The existing script quietly terminates in this case, leaving the tempfile behind; the `error` and `close` listeners for `tmpFile` do not execute. The failure of the installation may not become apparent until the user hits errors like

```
libvips-cpp.so.42: cannot open shared object file: No such file or directory
```

later while trying to use `sharp`.


lovell:
Great spot Alex, thank you very much for fixing this.

PR #2641, opened 2021-03-28
Extend the range of "speed" option in avif output params

janaz:
There's a new speed setting added to [`aom` in v 3.0.0](https://aomedia.googlesource.com/aom/+/refs/tags/v3.0.0)
> New speed 9 setting for faster realtime encoding.

This tiny PR extends the speed range for `avif` output from 0..8 to 0..9


janaz:
It seems that the "FreeBSD 13.0" check is failing on the "bijou" branch, so most likely the broken test is not caused by my change

lovell:
Hi Tomasz, good to hear from you, thanks for the PR.

Have you been able to test this with libvips? I ask as there's currently a 0-8 range limit on the `speed` parameter that will probably need lifting first:

https://github.com/libvips/libvips/blob/77ccd04cbe3939c5680b1b4e0aeebbb71febd6d8/libvips/foreign/heifsave.c#L509-L514

(See also https://github.com/libvips/libvips/pull/1819, which introduced the feature and has some discussion about what range would be suitable given the state of dependent libraries at that time.)

janaz:
Hi Lovell,

I haven't tested the change with libvips, but I'm going to do that as soon as possible and report if it works or not. If there's a limit of speed=8 in libvips then obviously the change either won't have any effect or will raise an error. If that's the case then I'll try to compile patched libvips locally to do some further testing.

I don't think there's a limit for the `speed` setting in AVIF encoders. I might be wrong but it looks like the speed values just turn on or off some toggles in the encoder's algorithm and each encoder may define its own range for the `speed` setting. For that reason, it might be better not to limit it in `libvips` or `sharp` as future versions of `aom` of `rav1e` are likely to introduce yet another speed value.

lovell:
It looks like this PR was auto-closed when I removed the work-in-progress `bijou` branch and now I don't appear to be able to reopen it or change the destination branch, sorry Tomasz. Please feel free to reopen when you're ready!

PR #2654, opened 2021-04-06
flatten jsdoc: object instead of string

timohausmann:
I just noticed that the docs show `flatten` with a string.

`flatten` expects an options object with background property,  passing a color string always results in black background.


lovell:
Good spot, thank you very much for fixing this Timo.

PR #2664, opened 2021-04-15
install: fix proxy authentication when password contains special character

msalettes:
Hello,

When we install `sharp` module on a machine which is behind a proxy, we have an error  `ERR! sharp tunneling socket could not be established, statusCode=407` if the password contains special characters.

This problem is because the `agent()` function returns a `password`where special characters are encoded. 
To fix this problem, we just need  to decode password before to return it. 

Open to any suggestions of course.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -25,7 +25,7 @@ module.exports = function () {
       ? tunnelAgent.httpsOverHttps
       : tunnelAgent.httpsOverHttp;
     const proxyAuth = proxy.username && proxy.password
-      ? `${proxy.username}:${proxy.password}`
+      ? `${decodeURIComponent(proxy.username)}:${decodeURIComponent(proxy.password)}`
       : null;
     return tunnel({
       proxy: {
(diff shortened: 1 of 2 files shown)


lovell:
Merci Matthieu, this fix makes sense based on the WHATWG spec for userinfo parsing. We should probably also apply the same change to the username too.

It's interesting to note that this bug/feature also exists in the recently-exposed `urlToHttpOptions` feature of Node.js itself.

https://github.com/nodejs/node/blob/e46c680bf2b211bbd52cf959ca17ee98c7f657f5/lib/internal/url.js#L1308

msalettes:
@lovell : You're welcome. I asked myself the same question about the username but i was not sure there was a use case to use special characters inside username (except for `.`, `-`, `_` , but those are not encoded) . But you're right, we can do it , it doesn't cost much to do it. I will update my pr to apply the change on username too.

As you said, it's interesting to the bug also exist on node. Maybe i will submit a pr .

msalettes:
One more thing, the CI fails when testing on freebsd . If i understand well, the error is because FreeBSD 13 uses the version `8.10.5` of `vips` but the current version of `sharp` requires the version `8.10.6`. 
Someone has a solution ?

lovell:
Yes, the FreeBSD failure is due to the very slightly older vips package (I should probably mark it as being allowed to fail).

Thanks again for the PR, this will be in v0.28.2.

Issue #1599, opened 2019-03-05
Enhancement: support premultiplied raw pixel input


justewg:
Thanks for utility, great work.
I use the utility inside the map-tile-server-gl node-app. Server configuration when rendering raster tiles allows you to set the color of the lines in rgba. But as a result, PNG-images enclose a gray or black substrate under the translucent white color. Result on the attached screenshot.
Images on the server are created as follows:
```
var image = sharp(data, {
          raw: {
            width: params.width * scale,
            height: params.height * scale,
            channels: 4
          }
        })
...
image.png()
```
I tried using .embed (), .ensureAlpha (), passing the create parameter to the constructor, pointing to background with transparent white, but it didn’t work.
How do I get a white backing in white translucent color, or do I get a 24-bit PNG supporting translucent colors?
![image](https://user-images.githubusercontent.com/16482115/53805477-4d1f9500-3f5b-11e9-8520-a6a3ff5d3dbb.png)

lovell:
Hello, what is the source of `data`? Could it have been premultiplied?

What does `...` refer to in this code sample? Are you able to provide a standalone code sample and input image data that exhibits this behaviour?

justewg:
https://github.com/klokantech/tileserver-gl/blob/master/src/serve_rendered.js
lines from 417 to 451

lovell:
https://github.com/mapbox/mapbox-gl-native/issues/9124

> "The Node binding however, returns the premultiplied image buffer as is"

justewg:
Sorry, I didn't quite understand what it was about in this post.
The point is that in any case, the image will be with a gray stroke under the translucent pixels?

lovell:
It looks like `tileserver-gl` uses `mapbox-gl-native` to render, which returns RGBA data that has been premultiplied. This is what's causing the artifacts you're seeing on (semi-)transparent pixels.

The quick fix would probably be to modify `tileserver-gl` to unpremultiply pixel values before passing them to sharp - there's an example of what's required to do this in https://github.com/consbio/mbgl-renderer/issues/25

justewg:
Thanks for help, I've add the issue to server's author, because haven't found the solution myself. The are no using mbgl-renderer as javascript-library, I see only mbgl/renderer c-sources and can't move futher, unfortunately

lovell:
v0.28.2 now available with support for this via the following API:
```js
sharp(input, { raw: { premultiplied: true, ... } })...
```

PR #2685, opened 2021-04-29
Accept premultiplied raw inputs

mnutt:
Addresses https://github.com/lovell/sharp/issues/1599.

I'm using `tileserver-gl` which works around using premultiplied inputs like so: https://github.com/maptiler/tileserver-gl/blob/4036d528ec9e/src/serve_rendered.js#L280-L294. This works, but from some simple benchmarks I've run it looks like it's faster to do in libvips than in javascript, especially for large images.

I ran into a bit of trouble with the unit test using `assertSimilar`, as the tests passed whether or not the image was unpremultiplied first so I ended up comparing the raw output directly.

If this approach looks acceptable I can add public API documentation.


src/common.cc
```diff
@@ -96,6 +96,10 @@ namespace sharp {
       descriptor->rawWidth = AttrAsUint32(input, "rawWidth");
       descriptor->rawHeight = AttrAsUint32(input, "rawHeight");
     }
+    // Raw pixel input is premultiplied
+    if (HasAttr(input, "rawPremultiplied")) {
```
lovell, line 100: If possible, it might be better to ensure `inputDescriptor.rawPremultiplied` is always set in the JavaScript code to avoid a conditional statement here.

lovell (commented):
Thank you very much for the PR, this looks great and it's always a pleasure to see unit tests. Please go ahead and add JSDocs for this, plus I've left one small comment inline.

lib/input.js
```diff
@@ -97,6 +97,12 @@ function _createInputDescriptor (input, inputOptions, containerOptions) {
         inputDescriptor.rawWidth = inputOptions.raw.width;
         inputDescriptor.rawHeight = inputOptions.raw.height;
         inputDescriptor.rawChannels = inputOptions.raw.channels;
+
+        if (is.bool(inputOptions.raw.premultiplied)) {
+          inputDescriptor.rawPremultiplied = inputOptions.raw.premultiplied;
+        } else {
+          inputDescriptor.rawPremultiplied = false;
+        }
```
mnutt, line 105: This could also just be `inputDescriptor.rawPremultiplied = !!inputOptions.raw.premultiplied;` if that's preferable.

docs/api-constructor.md
```diff
@@ -35,6 +35,7 @@ Implements the [stream.Duplex][1] class.
         -   `options.raw.width` **[number][8]?** integral number of pixels wide.
         -   `options.raw.height` **[number][8]?** integral number of pixels high.
         -   `options.raw.channels` **[number][8]?** integral number of channels, between 1 and 4.
+        -   `options.raw.premultiplied` **[boolean][7]?** specifies that the raw input has already been premultiplied, set to `true` to avoid sharp premultiplying the image. (optional, default `false`)
```
lovell, line 38: Please can you add this to the JSDocs at https://github.com/lovell/sharp/blob/master/lib/constructor.js#L141 then re-generate the markdown using `npm run docs-build`

lib/input.js
lovell: Yes please, sounds good.

lovell (changes requested):

lovell (approved):

lovell:
Brilliant, thank you very much, this will be in v0.28.2.

dantovbein:
Hi @lovell @mnutt,
I'm figuring out this error since today morning and it's blocking me an important deploy on Tuesday morning. 
```
> node-gyp rebuild
[91mERROR: Download of https://dl.bintray.com/lovell/sharp/libvips-8.4.2-linux-x64.tar.gz failed: Response code 403 (Forbidden)
[0m[91mgyp: Call to 'LDD_VERSION="ldd (Debian GLIBC 2.24-11+deb9u4) 2.24
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
Written by Roland McGrath and Ulrich Drepper." node -e "require('./binding').download_vips()"' returned exit status 1 while in binding.gyp. while trying to load binding.gyp
[0m[91mgyp ERR! configure error
```

I'll appreciate your help.
Dan

PR #2687, opened 2021-04-30
Detect empty input and throw a helpful error

### LINKED ISSUE #2681: Another "Input buffet contains unsupported image format"
> Are you using the latest version?

Yes: `0.28.1`

> What are the steps to reproduce?

1. Use `fs.watch` to observe a (USB-3) flash drive
1. On `"rename"`, call the below (where `pathToFile` is like `'/Volumes/FlashDrive/_MG_7743.JPG'`)
1. Drag `_MG_7743.JPG` onto the flash drive.

```js
const file = await fs.promises.readFile(pathToFile);

// …

sharp(file)
  .resize({ width: 800 })
  .toFile(dest);
```

> What is the expected behaviour?

For most images, this works fine. However, for seemingly random images (which otherwise appear to be fine), the Input Buffer unsupported image format is consistently thrown. I took all the images myself with the same camera, so I'm certain they're actually jpeg (unlike #1852, where the image was secretly a bitmap in disguise).

Curiously, the error is NOT thrown without `FSWatcher` in the mix (eg when the image is already on the flash drive).

Also curiously, the error is NOT thrown when `fs.watch` observes a same-disk-to-same-disk move (eg moving the image from elsewhere on the local hard drive to the watched folder on the local hard drive, or moving the image from elsewhere on the flash drive to the watched folder on the hard drive). That makes me think it's some kind of latency issue, BUT:

The `file` buffer appears to be intact in both scenarios, and the `readFile` is `await`ed, so it should have finished before the buffer is passed to Sharp.

However, in a case against latency (in addition to it being always the same images): It does not appear to be affected by file-size: some of the erroring ones are significantly smaller than others that non-error (ex one erroring is 2.5 MB, and one non-erroring is 4.5 MB). If it were a latency issue, larger files should be more prone to errorure than smaller (out of my test set of 18 images, 8 error, and their sizes are roughly evenly distributed in a range of 2.5 MB to 4.5 MB, and the 10 non-erroring are similarly ranged in size).

And finally, the error does not actually seem to matter—the operation still succeeds: the 800-width image gets written to `dest`, and AFAICT, the new file is fine.

My apologies if the problem is actually unrelated to Sharp (that seems possible since `fs.watch` appears to be a critical piece, but that doesn't make much sense either).

> Are you able to provide a minimal, standalone code sample, without other dependencies, that demonstrates this problem?

Yes, I can put one together if needed.

> Are you able to provide a sample image that helps explain the problem?

Yes, can do.

> What is the output of running `npx envinfo --binaries --system`?

```
  System:
    OS: macOS 10.15.7
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 15.41 GB / 32.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 15.14.0 - ~/.nvm/versions/node/v15.14.0/bin/node
    npm: 7.7.6 - ~/.nvm/versions/node/v15.14.0/bin/npm
```

JakobJingleheimer:
resolves #2681

### DIFF
--- a/lib/input.js
+++ b/lib/input.js
@@ -25,6 +25,9 @@ function _createInputDescriptor (input, inputOptions, containerOptions) {
     limitInputPixels: Math.pow(0x3FFF, 2),
     sequentialRead: false
   };
+  if (input && input.length === 0) {
+    throw Error('Input contains no data (its size is 0)');
+  }
   if (is.string(input)) {
     // filesystem
     inputDescriptor.file = input;
--- a/lib/input.js
+++ b/lib/input.js
@@ -25,17 +25,20 @@ function _createInputDescriptor (input, inputOptions, containerOptions) {
     limitInputPixels: Math.pow(0x3FFF, 2),
     sequentialRead: false
   };
-  if (input && input.length === 0) {
-    throw Error('Input contains no data (its size is 0)');
-  }
   if (is.string(input)) {
     // filesystem
     inputDescriptor.file = input;
   } else if (is.buffer(input)) {
     // Buffer
+    if (input.length === 0) {
+      throw Error('Input Buffer is empty');
+    }
     inputDescriptor.buffer = input;
   } else if (is.uint8Array(input)) {
     // Uint8Array or Uint8ClampedArray
+    if (input.length === 0) {
+      throw Error('Input Bit Array is empty');
+    }
     inputDescriptor.buffer = Buffer.from(input.buffer);
   } else if (is.plainObject(input) && !is.defined(inputOptions)) {
     // Plain Object descriptor, e.g. create
--- a/package.json
+++ b/package.json
@@ -74,7 +74,8 @@
     "Christian Flintrup <chr@gigahost.dk>",
     "Manan Jadhav <manan@motionden.com>",
     "Leon Radley <leon@radley.se>",
-    "alza54 <alza54@thiocod.in>"
+    "alza54 <alza54@thiocod.in>",
+    "Jacob Smith <jacob@frende.me>"
   ],
   "scripts": {
     "install": "(node install/libvips && node install/dll-copy && prebuild-install) || (node-gyp rebuild && node install/dll-copy)",
--- a/test/unit/io.js
+++ b/test/unit/io.js
@@ -335,17 +335,6 @@ describe('Input/output', function () {
     });
   });
 
-  it('Fail when input is empty Buffer', function (done) {
-    sharp(Buffer.alloc(0)).toBuffer().then(function () {
-      assert(false);
-      done();
-    }).catch(function (err) {
-      assert(err instanceof Error);
-      assert.strictEqual('Input buffer contains unsupported image format', err.message);
-      done();
-    });
-  });
-
   it('Fail when input is invalid Buffer', function (done) {
     sharp(Buffer.from([0x1, 0x2, 0x3, 0x4])).toBuffer().then(function () {
       assert(false);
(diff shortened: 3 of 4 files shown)


JakobJingleheimer:
1 unit test is failing locally, but I believe it's unrelated to my change:

```
  1) Platform-detection
       Defaults to ARMv6 for 32-bit:
     TypeError: 'deleteProperty' on proxy: trap returned falsish for property 'arm_version'
      at Context.<anonymous> (test/unit/platform.js:47:27)
      at processImmediate (node:internal/timers:464:21)
```

I have an M1 / Apple Silicon mac, which is 64-bit ARM, so maybe that's confusing it?

lib/input.js
```diff
@@ -25,6 +25,9 @@ function _createInputDescriptor (input, inputOptions, containerOptions) {
     limitInputPixels: Math.pow(0x3FFF, 2),
     sequentialRead: false
   };
+  if (input && input.length === 0) {
+    throw Error('Input contains no data (its size is 0)');
```
lovell, line 29: Please can you move this into the `if (is.buffer(input))` and `if (is.uint8Array(input)` blocks, then the error message can contain the type e.g. `Input Buffer is empty`.

lovell (changes requested):
Thank you for the PR, there's one change that I think could improve the error messaging but otherwise this is good to merge. (This is a very slightly breaking change, but I think we're OK for a patch release.)

lib/input.js
JakobJingleheimer: That's actually how I first implemented it, but then I changed it to this to avoid the redundancy. I checked the docs, and all accepted inputs have a `length` property, so it seemed appropriate to do as a catch-all. I considered the lack of type in the error message, but reasoned that it didn't provide anything particularly important since the problem is not related to the type (and the extra type information might lead to a red herring for the consumer).

Can change though.

I'm thinking to also update the test to include an assertion on the error's message to ensure it's the correct error (and not something later).

lovell:
The unrelated test failure was due to a change in Node.js 16 that prevents `process.config.variables` being modified (which the tests really shouldn't have been doing anyway). Commit https://github.com/lovell/sharp/commit/a688468378a9377be3f3d2620645b471f8cbe0db should allow these to pass again.

lovell:
Thanks for the updates, did you see there's a linter error that's preventing the tests from completing?

JakobJingleheimer:
Nope. I don't see anything about a lint error.

I see the `install` step failed in Task Summary due to `fatal error: 'vips/vips8' file not found` (which I definitely didn't touch).

And Travis appears to be failing due to the same ARM64 problem (despite having rebased onto upstream master).

JakobJingleheimer:
Ahhh, I see it buried in Travis under the ARM64 task and as a sub-step to `npm test`; I was running `test-unit`.

Fixed.

Might be worthwhile to add a specific npm script so it doesn't get lost for it and maybe also an eslint or similar config so IDEs can pick that up automatically 🙂

lovell:
Thank you very much Jakob, hopefully this will help prevent others running in to the same problem.

lovell:
Commit https://github.com/lovell/sharp/commit/309918a878240e912b2b8ee62530a4e9b32caf60 moves the linter tasks into their own script entry to make things a bit clearer.

If you use vscode and hadn't already seen it, https://github.com/standard/vscode-standard provides a handy extension that detects the use of standard/semistandard in any workspace and displays warnings inline as you edit.

JakobJingleheimer:
Awesome, thank you!

I'll check out that VS Code plugin 🙂

PR #2726, opened 2021-05-20
Add clahe operator

### LINKED ISSUE #1066: Enhancing clarity of an image
Hey I have an image that I will be passing into tesseract, but before passing it in I need to enhance the "clarity" of the image for ocr. I can do this through windows photos "Edit=>Clarity" but want to know how I can get the same effect using sharp

Thanks

baparham:
This should fix #1066 ~but it still needs tests.~

Is something like this along the right track?

@lovell Do you have any advice on an overall methodology for testing something like this? Should I generate some sample images or something directly with libvips and use those as expected inputs and outputs for a test? What's the standard practice here?


baparham:
This really just wraps the vips hist_local command, so perhaps the arguments to this operator need to match the valid inputs to vips:

```
C:\vips-dev-8.10\bin>vips.exe hist_local
local histogram equalisation
usage:
   hist_local in out width height [--option-name option-value ...]
where:
   in           - Input image, input VipsImage
   out          - Output image, output VipsImage
   width        - Window width in pixels, input gint
                        default: 1
                        min: 1, max: 10000000
   height       - Window height in pixels, input gint
                        default: 1
                        min: 1, max: 10000000
optional arguments:
   max-slope    - Maximum slope (CLAHE), input gint
                        default: 0
                        min: 0, max: 100
operation flags: sequential
```

I believe I can generate some sample expected outputs directly with libvips and use those for some unit tests, as demonstrated by something like the unit tests for the `.blur()` operator. I'll do that tonight when I get some more time.

I think it also makes sense to update the `.clahe()` valid arguments to accept even values too, and be more clear with the min and max as defined by the libvips help above.

test/fixtures/index.js
```diff
@@ -121,6 +121,8 @@ module.exports = {
 
   inputV: getPath('vfile.v'),
 
+  inputJpgClahe: getPath('concert.jpg'),
```
lovell, line 124: A reverse search for the test image suggests the source might be https://www.flickr.com/photos/mars_/14389236779/ - if so, please can you add a link to this in the test fixtures file so we know it's in the public domain.

src/pipeline.cc
```diff
@@ -1330,6 +1336,9 @@ Napi::Value pipeline(const Napi::CallbackInfo& info) {
   baton->linearB = sharp::AttrAsDouble(options, "linearB");
   baton->greyscale = sharp::AttrAsBool(options, "greyscale");
   baton->normalise = sharp::AttrAsBool(options, "normalise");
+  baton->claheWidth = sharp::AttrAsInt32(options, "claheWidth");
```
lovell, line 1339: Can use the unsigned `AttrAsUint32` for dimensions as they will always be positive.

lib/operation.js
```diff
@@ -350,6 +350,40 @@ function normalize (normalize) {
...
+  }
+  if (!('height' in options) || !is.integer(options.height) || options.height <= 0) {
+    throw is.invalidParameterError('height', 'integer above zero', options.height);
+  } else {
+    this.options.claheHeight = options.height;
+  }
+  if (!is.defined(options.maxSlope)) {
+    this.options.claheMaxSlope = 3.0;
+  } else if (!is.number(options.maxSlope) || options.maxSlope < 0 || options.maxSlope > 100) {
+    throw is.invalidParameterError('maxSlope', 'positive number 0.0-100.0 or zero', options.maxSlope);
```
lovell, line 380: Could remove the "or zero" part as the range covers this.

lib/operation.js
```diff
@@ -350,6 +350,40 @@ function normalize (normalize) {
   return this.normalise(normalize);
 }
 
+/**
+ * Perform contrast limiting adaptive histogram equalization
```
lovell, line 354: Perhaps we could add a short description about using this operation to improve the clarity of an image, if only to help people find it (e.g. via keyword search)? Might be worth adding a link to https://en.wikipedia.org/wiki/Adaptive_histogram_equalization#Contrast_Limited_AHE too.

lib/operation.js
```diff
@@ -368,7 +402,7 @@ function normalize (normalize) {
  *
  * @param {Object} kernel
  * @param {number} kernel.width - width of the kernel in pixels.
- * @param {number} kernel.height - width of the kernel in pixels.
+ * @param {number} kernel.height - height of the kernel in pixels.
```
lovell, line 405: Good spot, thank you for fixing this.

lovell (changes requested):
This looks great, thank you very much Brad. I've left a few small questions/comments inline that might need addressing but otherwise this is good to merge.

test/fixtures/index.js
baparham: yes! that is where I found it and had the url copied to my clipboard to do so. Nice catch.

baparham:
Fixed your comments with the [force push](https://github.com/lovell/sharp/compare/43601f9dd851aa5252feaba5ea2bcca992fc27a6..aca67457da8350804f7ea186254a41b92f76ec45). Thanks for the quick feedback.

baparham:
Do types automatically get rebuilt when you do a release or should I include some update to the typings for Typescript in this pull request?

baparham:
Hmm, vips actually looks (and behaves, now that I've tried this with non-integers like 0.5 or 2.5) like it only takes integers as the maxSlope, so I'll need to modify the expected type and argument checking of the `maxSlope` argument.

baparham:
> Hmm, vips actually looks (and behaves, now that I've tried this with non-integers like 0.5 or 2.5) like it only takes integers as the maxSlope, so I'll need to modify the expected type and argument checking of the `maxSlope` argument.

I made these changes, so now the maxSlope should be an int, which matches what the libvips hist_local api accepts.

It was just truncating the decimel before, so 0.5 would be the same as 0, and 1.0, 1.1, and 1.9 would all return the same result as if maxSlope = 1.

baparham:
Darn, not quite, I missed a few things that make the build fail, and now that I've changed everything, the tests are failing with similarity thresholds that are too high. I'll manually compare the the images to see if they look as they should compared with the libvips outputted images.

baparham:
> Darn, not quite, I missed a few things that make the build fail, and now that I've changed everything, the tests are failing with similarity thresholds that are too high. I'll manually compare the the images to see if they look as they should compared with the libvips outputted images.

The threshold failing tests were resolved after deleting and rebuilding `build`, so I think it was just that something was out of sync in the built library with the js interface perhaps. I also fixed a few c++ errors that I had missed before when changing from double to int on maxSlope.

It should be good now, and I've tested it manually a bit outside the tests to double check.

lovell (approved):

lovell:
Marvellous, thanks again.

> Do types automatically get rebuilt when you do a release or should I include some update to the typings for Typescript in this pull request?

These are managed by the https://github.com/DefinitelyTyped/DefinitelyTyped repo. If you're able to add these (after v0.28.3 is released), that would be great.

baparham:
I have the patch (since I needed it locally for testing my client code anyway) so I'll push it into a draft pull request over there to time when 0.28.3 is released and keep you in the loop.

Thanks for the help getting this in!

Issue #1454, opened 2018-11-08
Enhancement: set pixel depth on raw output


akidee:
I read a quite large uncompressed TIFF (107 MB) into a buffer:

`img.raw().toBuffer().then(buffer => { console.log(buffer.length) })`

Result: `53496576` which is only half of the expected value:
The image has 3 channels a 16 bits and its dimensions are 5208 x 3424, and so I'd expect `buffer.length` to be 5208 * 3424 * 3 * 2 = 106993152

It seems that the buffer is manipulated by reducing depth to 8 bit.

lovell:
Hello, returning 8-bit integer pixel values is the current documented behaviour.

> "Force output to be raw, uncompressed uint8 pixel data."
> http://sharp.pixelplumbing.com/en/stable/api-output/#raw

akidee:
I request this feature. Supporting 16 bit by request would be great since "raw" should at least imply that all information from the original image is accessible.

lovell:
Optional parameters could be added to the [raw](http://sharp.pixelplumbing.com/en/stable/api-output/#raw) function, allowing a `depth` property e.g. `uchar` (default), `ushort`, `float` etc. to match [libvips' nicknames](https://github.com/libvips/libvips/blob/7da31386c73cacc17116529463529014f8525cc1/libvips/iofuncs/enumtypes.c#L748-L757).

**Proposed API:**
```javascript
sharp(input)
  .raw({ depth: 'ushort' })
  .toBuffer()
  ...
```

Happy to accept/help with a PR if you're able.

lovell:
v0.29.0 now available with this feature.

PR #2762, opened 2021-06-21
Add support for different bit depths with raw input and output

mart-jansink:
Fixes https://github.com/lovell/sharp/issues/1454 and fixes https://github.com/lovell/sharp/issues/1289. I've mostly added just the logic to support input or output raw data in different bit depths for now. If you like it, I'll also adjust the documentation accordingly and ensure test coverage of course.

***

### Raw input example
```javascript
const input = new Uint16Array( 48 );
for ( let i = 1; i < 48; i += 3 ) {
  input[ i ] = 0xffff;
}
sharp( input, {
  raw: {
    width: 4,
    height: 4,
    channels: 3,
  },
} ).toFile( "16bpc.png" );
```
This generates a green 4x4 .png file with 16 bits per pixel per color.

***

### Raw output example
```javascript
sharp( "16bpc.png" )
  .toColourspace( "rgb16" )
  .raw( {
    depth: "ushort",
  } )
  .toBuffer( ( error, data, {
    width,
    height,
    channels,
    size,
  } ) => {
    console.log( size / width / height / channels * 8 );
    console.log( new Uint16Array( data.buffer ) );
  } );
```
This takes the above .png file and reads the raw pixels from it as an `Uint16Array`. Note that the raw output still defaults to 8-bit results, so you have to explicitly use `{ depth: "ushort" }`. Also, `.toColourspace( "rgb16" )` is needed to prevent downsampling earlier on in the pipeline with 
https://github.com/lovell/sharp/blob/61640fb5c7ed834c264756e224ec81079f8baec2/src/pipeline.cc#L711
Maybe you have a suggestion on how to do this automatically?

***

Please let me know your thoughts on this PR, I'd happy to work some more on it to get it merged.

### DIFF
--- a/lib/constructor.js
+++ b/lib/constructor.js
@@ -253,6 +253,7 @@ const Sharp = function (input, options) {
     heifCompression: 'av1',
     heifSpeed: 5,
     heifChromaSubsampling: '4:4:4',
+    rawDepth: 'uchar',
     tileSize: 256,
     tileOverlap: 0,
     tileContainer: 'fs',
--- a/lib/constructor.js
+++ b/lib/constructor.js
@@ -92,8 +92,9 @@ const debuglog = util.debuglog('sharp');
  *  }
  * }).toFile('noise.png');
  *
- * @param {(Buffer|Uint8Array|Uint8ClampedArray|string)} [input] - if present, can be
- *  a Buffer / Uint8Array / Uint8ClampedArray containing JPEG, PNG, WebP, AVIF, GIF, SVG, TIFF or raw pixel image data, or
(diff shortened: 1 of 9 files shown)


lovell:
Hello Mart, thank you for your PR, this will be a welcome addition.

If you hadn't seen it, I'm planning to review https://github.com/lovell/sharp/pull/2704 very soon, and I think it's worth getting that merged first as it should then help improve what's in this proposed change.

mart-jansink:
Perfect! The main reason that I need this modifications is to prevent unnecessary banding while processing 16-bit images, so #2704 will definitely help.

lovell:
This looks great, thank you. Please can you go ahead and add tests for all this useful new logic.

mart-jansink:
Apologies for the delay, but I've now rebased and added some unit tests for this functionality. I'm not getting 100% coverage on `output.js`, but that's already not the case with the `circle` branch itself.

lovell (approved):

lovell:
Bedankt Mart!

PR #2804, opened 2021-07-19
Docs: add stats usage note and example (#2803)

### LINKED ISSUE #2803: Clarify usage of stats on processed/extracted images
Is my understanding correct that `stats()` works solely on the source image and not on the result of any operations done on it - i.e. it cannot be piped this way:

```js
await sharp( ... ).extract( ... ).stats();
```

but needs a buffer to store the intermediate result like so:

```js
const buf = await sharp( ... ).extract( ... ).toBuffer();
await sharp(buf).stats();
```

If that's the case, could this usage perhaps be documented [here](https://sharp.pixelplumbing.com/api-input#stats)? Happy to send a PR your way.

For context: I was trying to do something similar to https://github.com/lovell/sharp/issues/2708 and couldn't work out for the life of me why `dominant` was giving me unexpected results.

reiv:
Fixes #2803

Hopefully this makes the usage a bit clearer. Feel free to change the order/wording.

### DIFF
--- a/lib/input.js
+++ b/lib/input.js
@@ -356,6 +356,9 @@ function metadata (callback) {
  * - `sharpness`: Estimation of greyscale sharpness based on the standard deviation of a Laplacian convolution, discarding alpha channel if any (experimental)
  * - `dominant`: Object containing most dominant sRGB colour based on a 4096-bin 3D histogram (experimental)
  *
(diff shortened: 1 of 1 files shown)


lovell:
Thank you!

Issue #2845, opened 2021-08-16
Feature request: lightness


Tenpi:
What are you trying to achieve?

I want to more closely mimic Photoshop's HSL adjustment in my app. From my understanding, the lightness operation adds either white or black to the image (where +100 is fully white and -100 is fully black). This results in a slightly different effect than brightness, which brightens the colors in the image. 

Have you searched for similar feature requests?

Yes, and I didn't find anything similar.

What would you expect the API to look like?

```js
sharp(input).lightness(value) // Range from -100 to 100
```

What alternatives have you considered?

Jimp has a lightness operation, but it is a much slower library than sharp is.

Is there a sample image that helps explain?

Lightness example (from Photoshop):
![lightness](https://user-images.githubusercontent.com/37512637/129565391-1af42f49-74da-4a3b-9d65-6ae76d8d798d.png)
![darkness](https://user-images.githubusercontent.com/37512637/129565405-f62a2002-b050-4ffa-9bb4-a3a6cf256fb1.png)

This is different from the brightness operation in sharp, which doesn't affect the blacks in the image:
![lightness-sharp](https://user-images.githubusercontent.com/37512637/129565536-3199119e-295b-4593-8862-32ccd836f001.png)
![darkness-sharp](https://user-images.githubusercontent.com/37512637/129565552-9860761b-695e-4722-8c88-913a1d6475cc.png)

lovell:
The existing [modulate](https://sharp.pixelplumbing.com/api-operation#modulate) operation converts to LCH (luminance, chroma, hue) and _multiplies_ the L channel by the `brightness` parameter.

https://github.com/lovell/sharp/blob/7467fa8b50d280780dfa9fe9a7248deaa499c7a1/src/operations.cc#L172-L193

As you suggest, Photoshop's "lightness" control appears to be _additive_. The range of the L channel is 0 (black) to 100 (white), so we could extend modulate to add a new `lightness` property that adds the provided value to the L channel, with clipping.

```js
// PROPOSED API
.modulate({ lightness: 10 })
```

Happy to accept a PR if you're able.

Tenpi:
> Happy to accept a PR if you're able.

I'm not strong with C++ but I could try.

PR #2846, opened 2021-08-17
Feature: Add lightness option to modulate

Tenpi:
This adds a lightness option to modulate in order to more closely mimic Photoshop's lightness adjustment. It does this by adding the provided value to the luminance channel instead of multiplying by it. Related to my earlier feature request (#2845).

Image Example:

Code Example:
```js
await sharp(image)
      .modulate({lightness: 10})
      .toBuffer()
```


lovell:
Thank you very much for this, it looks great.

Is it worth adding a sentence to the JSDocs to clarify that lightness is additive whereas brightness is multiplicative?

Please can you add a unit test for the new error handling path too.

Tenpi:
> Please can you add a unit test for the new error handling path too.

Do you mean adding it to the "invalid options" test? If so, I have updated it.

lovell:
Landed via commit https://github.com/lovell/sharp/commit/4a9267ce127c435c5c04dfa33dfdd1396d7de50c, thank you. This will be in v0.29.1

Issue #2864, opened 2021-08-27
ImageMagick7 and "Input file contains unsupported image format"


paulstraw:
Hi there. I'm running into an issue which seems very similar to #2303. My goal is to get sharp running with ImageMagick and lots of other tools (libheif, libwebp, etc) on a Docker image destined for AWS Lambda. Everything is working well except for ImageMagick. It's detected during `libvips` installation, and can be used with e.g. `vips magickload`, but trying to load a `.bmp` or other IM-compatible file with sharp gives "Input file contains unsupported image format". I've created a basic repro of the problem here: https://github.com/paulstraw/sharp-magick-lambda

Thanks very much for your help, and your work on sharp. Please let me know if there's any additional information I can provide.

---

> Did you see the [documentation relating to installation](https://sharp.pixelplumbing.com/install)?

Yep, I've referenced this doc extensively, and it's been very helpful in getting this far.

> Have you ensured the architecture and platform of Node.js used for `npm install` is the same as the architecture and platform of Node.js used at runtime?

Yes, there is only one Node installed on the Docker image I'm using.

> Are you using the latest version? Is the version currently in use as reported by `npm ls sharp` the same as the latest version as reported by `npm view sharp dist-tags.latest`?

Yes, sharp `0.29.0`. 

> If you are using npm v6 or earlier and installing as a `root` or `sudo` user, have you tried with the `npm install --unsafe-perm` flag?

Yes, although I've also tried without, and installation appears to work successfully either way.

> If you are using npm v7, does the user running `npm install` own the directory it is run in?

N/A

> If you are using the `ignore-scripts` feature of `npm`, have you tried with the `npm install --ignore-scripts=false` flag?

N/A

> What is the complete output of running `npm install --verbose sharp`? Have you checked this output for useful error messages?

Yes. I don't see any errors, and the system libvips is detected.

<details>
<summary>

`npm install --verbose sharp`

</summary>

```bash
# npm install --verbose sharp
npm info it worked if it ends with ok
npm verb cli [
npm verb cli   '/var/lang/bin/node',
npm verb cli   '/var/lang/bin/npm',
npm verb cli   'install',
npm verb cli   '--verbose',
npm verb cli   'sharp'
npm verb cli ]
npm info using npm@6.14.14
npm info using node@v14.17.5
npm verb npm-session e53379bbc505ec01
npm http fetch GET 200 https://registry.npmjs.org/sharp 406ms
[...]

PR #2865, opened 2021-08-27
Add ImageMagick 7 loaders

paulstraw:
This is a fix for #2864. I've updated my [reproduction repo for that issue](https://github.com/paulstraw/sharp-magick-lambda) to point at this branch, and it does indeed fix the bug.


lovell:
Brilliant, thank you, this will be in v0.29.1.

Issue #2245, opened 2020-06-05
Sharp convert to jp2 format (JPEG2000)


hegyre:
Hello,
I'm using latest versions of sharp and libvips

What are you trying to achieve?
I want to convert images like .png to .jp2
I've compiled and installed libvips globally, with imagemagick, itself compiled with jpeg2000 support.

Have you searched for similar questions?
Only found that one from 2018: https://github.com/lovell/sharp/issues/1489

Are you able to provide a minimal, standalone code sample that demonstrates this question?
If you could help me understand what should I update in the code to support something like:
```
sharp(input)
  .toFormat('jp2')
  .toBuffer();
```

Thanks in advance.

lovell:
The work to expose magick-based output has started with PR #2012

hegyre:
OK so it seems that one should adapt that PR #2012 to allow support for additional imagemagick formats (like .jp2)

Thanks for your answer @lovell ! much appreciated.

LosTigeros:
So is this finally a thing? I would like to use either `.toFormat('jp2')` or `.toFormat('jxr')`.

lovell:
In the same way that https://github.com/lovell/sharp/pull/2012 exposed a `.gif()` function for GIF output via *magick, this will need someone to do similar for `.jp2()`. Always happy to accept PRs, if you're able.

inukshuk:
@lovell with 8.11 jp2 should be supported directly by libvips if compiled with libopenjpeg2, right? I suspect that this will make supporting the a dedicated `.jp2()` function more straightforward. I'd be happy to take a stab at this, if you would like to see this in included (I could also add openjpeg2 to sharp-libvips if you would like to include it by default).

lovell:
@inukshuk Great, yes please, libvips 8.11.0 added `jp2ksave` (and `jp2kload`) so a PR to expose this API is welcome, although I'd prefer not to include openjpeg in the prebuilt binaries.

hegyre:
Hello @lovell ,
Just curious about not including openjpeg in prebuild binaries.
Any specific reason ?
Thanks!

lovell:
v0.29.1 is now available with support for JP2 when you compile sharp from source using a globally-installed libvips that has been compiled with support for OpenJPEG.

https://sharp.pixelplumbing.com/api-output#jp2

PR #2868, opened 2021-08-28
Feat: Add support for JP2

Zaruike:
Requires `libvips 8.11.0`

This add the support for made JP2 file and JP2 buffer.

More info:
#2245

Code Example:

```js
await sharp(image)
      .jp2({quality: 90})
      .toBuffer()
```


lib/constructor.js
```diff
@@ -234,6 +234,11 @@ const Sharp = function (input, options) {
     pngQuality: 100,
     pngBitdepth: 8,
     pngDither: 1,
+    jp2Quality: 80,
+    jp2TileHeight: 256,
```
lovell, line 238: libvips has default tile dimensions of 512 - I'd prefer to keep things consistent unless there's a reason for the lower value.

lib/output.js
```diff
@@ -511,6 +516,80 @@ function gif (options) {
   return this._updateFormatOut('gif', options);
 }
 
+/**
+ * Use these JP2 options for output image.
+ *
```
lovell, line 521: Please can you add info about the lack of prebuilt support for this, something like:
```
 * Requires libvips compiled with support for OpenJPEG.
 * The prebuilt binaries do not include this - see
 * {@link https://sharp.pixelplumbing.com/install#custom-libvips installing a custom libvips}.
```

lib/output.js
```diff
@@ -511,6 +516,80 @@ function gif (options) {
...
+    if (is.defined(options.lossless)) {
+      if (is.bool(options.lossless)) {
+        this.options.jp2Lossless = options.lossless;
+      } else {
+        throw is.invalidParameterError('lossless', 'boolean', options.lossless);
+      }
+    }
+
+    if (is.defined(options.tileWidth)) {
+      if (is.integer(options.tileWidth) && options.tileWidth > 0) {
```
lovell, line 569: Tile dimensions should be in the 1-32768 range.

lib/output.js
```diff
@@ -511,6 +516,80 @@ function gif (options) {
...
+ *
+ * @param {Object} [options] - output options
+ * @param {number} [options.quality=80] - quality, integer 1-100
+ * @param {boolean} [options.lossless=false] - use lossless compression mode
+ * @param {number} [options.tileWidth=256] - horizontal tile size
+ * @param {number} [options.tileHeight=256] - vertical tile size
+ * @param {string} [options.chromaSubsampling='4:4:4'] - set to '4:2:0' to use chroma subsampling
+ * @returns {Sharp}
+ * @throws {Error} Invalid options
+ */
```
lovell, line 545: Might need to add `/* istanbul ignore next */` in here to avoid test coverage failure - see `gif()` for similar.

lib/output.js
```diff
@@ -13,10 +13,15 @@ const formats = new Map([
...
-  ['gif', 'gif']
+  ['gif', 'gif'],
+  ['jp2', 'jp2'],
+  ['jpx', 'jp2'],
+  ['j2k', 'jp2'],
+  ['j2c', 'jp2']
 ]);
 
 const errMagickSave = new Error('GIF output requires libvips with support for ImageMagick');
+const errJp2Save = new Error('JP2 output requires libvips with support for openjpeg');
```
lovell, line 24: Might need to change `openjpeg` to `OpenJPEG` as the unit tests match on `/JP2 output requires libvips with support for OpenJPEG/`

lovell (changes requested):
Thank you very much for this, I've left a few questions/comments inline that we should consider then this is ready to merge.

lib/constructor.js
Zaruike: There is no particular reason it was while I was performing my tests I forgot to reset it to 512 😅

lovell:
Landed via commit https://github.com/lovell/sharp/commit/b7fbffb3f7e4854da2d9629950bd3dbb63e26c45 - merci beaucoup!

PR #2878, opened 2021-09-06
Resolve paths before comparing input/output destination

### LINKED ISSUE #2873: Improve detection of input/output file name matching
Reading from and writing to the same file should be supported

rexxars:
This fixes an issue where if you try to write to the same destination as the input file but you are not using absolute (or the same relative path) for both the input and output, sharp/vips might produce errors such as:

```
someFile.jpg: unable to open for write
unix error: No such file or directory
```

Fixes #2873

### DIFF
--- a/lib/output.js
+++ b/lib/output.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const path = require('path');
 const is = require('./is');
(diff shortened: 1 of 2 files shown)


lovell:
Thank you!

PR #2879, opened 2021-09-06
Allow using speed 9 for AVIF/HEIC encoding

### LINKED ISSUE #2861: Support speed=9 for AVIF output
Since this is released in 8.11, we can add speed=9 support in sharp as well: https://github.com/libvips/libvips/issues/2180

rexxars:
Fixes #2861

Do we need to check for specific versions of libvips or libaom, or is this fine as it stands?


lovell:
Thanks again!

Issue #2775, opened 2021-06-30
Allowing alternate tif format for tiff


vaclavbarta:
Similar to jpeg / jpg we would appreciate adding alternate `tif` delegating to `tiff` format in formats map which is used in toFormat method.

https://github.com/lovell/sharp/blob/7467fa8b50d280780dfa9fe9a7248deaa499c7a1/lib/output.js#L14

RhnSharma:
I would like to take this one.

PR #2893, opened 2021-09-16
Allow alternate tif format for tiff for #2775

erf:
Should solve #2775 by adding an alternative `tif` format for `tiff`.

Added a unit test which calls `toFormat('tif')` and asserts format is `tiff`.


lovell:
Thank you / Takk !

PR #2952, opened 2021-10-28
Impute TIFF xres/yres from withMetadata({density})

### LINKED ISSUE #2941: Enhancement: use density provided to withMetadata as fallback value for TIFF xres/yres
_What are you trying to achieve?_

Set a specific density/resolution on an output image. (I realize this changes nothing inherent about the image, but some of our users are getting confused by the default print dimensions of an image based on its stored resolution and I'd like to make the problem go away.)

_Have you searched for similar questions?_

Yes. Other questions/issues have come close, but haven't quite addressed the difference between output formats when it comes to metadata and density.

_Are you able to provide a minimal, standalone code sample that demonstrates this question?_

Given this input file:
```
$ identify -verbose input.tif
Image:
  Filename: input.tif
  Format: TIFF (Tagged Image File Format)
  Mime type: image/tiff
  Class: DirectClass
  Geometry: 2998x4454+0+0
  Resolution: 3000x3000
  Print size: 0.999333x1.48467
  Units: PixelsPerInch
```

I run the following:
```
sharp('input.tif').withMetadata({ density: 96 }).jpeg().toFile('output.jpg');
sharp('input.tif').withMetadata({ density: 96 }).png().toFile('output.png');
sharp('input.tif').withMetadata({ density: 96 }).tiff().toFile('output.tif');
```

The output looks like this:
```
$ identify -verbose output.jpg
Image:
  Filename: output.jpg
  Format: JPEG (Joint Photographic Experts Group JFIF format)
  Mime type: image/jpeg
  Class: DirectClass
  Geometry: 2998x4454+0+0
  Resolution: 96x96
  Print size: 31.2292x46.3958
  Units: PixelsPerInch

$ identify -verbose output.png
Image:
  Filename: output.png
  Format: PNG (Portable Network Graphics)
  Mime type: image/png
  Class: DirectClass
  Geometry: 2998x4454+0+0
  Resolution: 37.8x37.8
  Print size: 79.3122x117.831
  Units: PixelsPerCentimeter

$ identify -verbose output.tif
Image:
  Filename: output.tif
  Format: TIFF (Tagged Image File Format)
  Mime type: image/tiff
  Class: DirectClass
  Geometry: 2998x4454+0+0
  Resolution: 25.4x25.4
  Print size: 118.031x175.354
  Units: PixelsPerInch
```

I get why the JPEG and the PNG are using different units (because the PNG header always expresses density in pixels per meter), but at least the density math works out to 96 PPI. The TIFF keeps the original units, but gets written out using sharp's default density of 1000 pixels per meter.

FWIW, `sharp` is capable of _reading_ the density from the source TIFF:

```
> await sharp('input.tif').metadata();
{
  format: 'tiff',
  width: 2998,
  height: 4454,
  space: 'b-w',
  channels: 1,
  depth: 'uchar',
  density: 3000,
  isProgressive: false,
  pages: 1,
  hasProfile: true,
  hasAlpha: false,
  orientation: 1,
  ...
}
```

_Are you able to provide a sample image that helps explain the question?_

Yes. (It's zipped just to prevent GitHub from trying to create a derivative or inlining it into the issue.)
[input.tif.zip](https://github.com/lovell/sharp/files/7400799/input.tif.zip)

mbklein:
Closes #2941 

This PR uses the `density` value passed to `withMetadata()` as a fallback if the output format is TIFF and netter `xres` nor `yres` is provided in the TIFF options. I've also updated the API documentation with to include a short explanation of this behavior.

### DIFF
--- a/docs/api-output.md
+++ b/docs/api-output.md
@@ -125,7 +125,7 @@ sRGB colour space and strip all metadata, including the removal of any ICC profi
     *   `options.orientation` **[number][9]?** value between 1 and 8, used to update the EXIF `Orientation` tag.
     *   `options.icc` **[string][2]?** filesystem path to output ICC profile, defaults to sRGB.
     *   `options.exif` **[Object][6]<[Object][6]>** Object keyed by IFD0, IFD1 etc. of key/value string pairs to write as EXIF data. (optional, default `{}`)
-    *   `options.density` **[number][9]?** Number of pixels per inch (DPI).
+    *   `options.density` **[number][9]?** Number of pixels per inch (DPI). (If the output format is TIFF, this value will only apply if the TIFF options _do not_ explicitly specify a non-default `xres` and/or `yres`.)
 
 ### Examples
 
--- a/src/pipeline.cc
+++ b/src/pipeline.cc
@@ -1492,6 +1492,9 @@ Napi::Value pipeline(const Napi::CallbackInfo& info) {
   baton->tiffTileHeight = sharp::AttrAsUint32(options, "tiffTileHeight");
   baton->tiffXres = sharp::AttrAsDouble(options, "tiffXres");
   baton->tiffYres = sharp::AttrAsDouble(options, "tiffYres");
+  if (baton->tiffXres == 1.0 && baton->tiffYres == 1.0 && baton->withMetadataDensity > 0) {
+    baton->tiffXres = baton->tiffYres = baton->withMetadataDensity / 25.4;
+  }
   // tiff compression options
   baton->tiffCompression = static_cast<VipsForeignTiffCompression>(
   vips_enum_from_nick(nullptr, VIPS_TYPE_FOREIGN_TIFF_COMPRESSION,
--- a/test/unit/tiff.js
+++ b/test/unit/tiff.js
@@ -188,6 +188,26 @@ describe('TIFF', function () {
       )
   );
 
+  it('TIFF imputes xres and yres from withMetadataDensity if not explicitly provided', async () => {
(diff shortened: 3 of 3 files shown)


lovell:
Hi Michael, thank you very much for this PR, and always a pleasure to see unit tests too.

I wonder if it might be a little simpler to have this logic in the C++ layer, perhaps something like the following:

```diff
--- a/src/pipeline.cc
+++ b/src/pipeline.cc
@@ -1496,6 +1496,9 @@ Napi::Value pipeline(const Napi::CallbackInfo& info) {
   baton->tiffTileHeight = sharp::AttrAsUint32(options, "tiffTileHeight");
   baton->tiffXres = sharp::AttrAsDouble(options, "tiffXres");
   baton->tiffYres = sharp::AttrAsDouble(options, "tiffYres");
+  if (baton->tiffXres == 1.0 && baton->tiffYres == 1.0 && baton->withMetadataDensity > 0) {
+    baton->tiffXres = baton->tiffYres = baton->withMetadataDensity / 25.4;
+  }
   // tiff compression options
   baton->tiffCompression = static_cast<VipsForeignTiffCompression>(
   vips_enum_from_nick(nullptr, VIPS_TYPE_FOREIGN_TIFF_COMPRESSION,
```
I think this should still satisfy the test cases you added.

mbklein:
I skimmed `pipeline.cc` for an appropriate spot to add it, but came up short. This is much more concise. I'll make the change and force-push my branch.

lovell:
Brilliant, thank you, landed via commit https://github.com/lovell/sharp/commit/342de369737f7618324867def4b330f938a22168

Issue #2934, opened 2021-10-17
Resizing png 16-bit colorspace stopped working after 0.29.1 update


gforge:
After updating to 0.29.1 from 0.29.0 my tests start failing when parsing buffers after resize

What are the steps to reproduce?
I have a resize function that uses `sharp` and where the data is sent over a buffer where `sharp` fails to reconstruct the `16-bit` data

To reproduce using `sharp`s own unit tests:
```
  it("16-bit PNG after resize should retain 16-bit space", async function () {
    const img = sharp(fixtures.inputPngWithTransparency16bit);
    const orgSpace = (await img.metadata()).space;

    const space = await new Promise((resolve, reject) => {
      img
        .resize(32, 16, { fit: "contain" })
        .toBuffer(async function (err, data, info) {
          if (err) throw err;

          resolve((await sharp(data).metadata()).space);
        });
    });
    assert.equal(space, orgSpace);
  });
```

lovell:
Hi, v0.29.0 and v0.29.1 (and v0.28.3) produce the same result for me, namely that `orgSpace` is "rgb16" and `space` is "srgb", which is expected.

For "rgb16" output use `toColourspace()`, there's an example in the docs at https://sharp.pixelplumbing.com/api-colour#tocolourspace

gforge:
Wow, thanks for the quick reply. Maybe I missed something when writing up my test. My original problem is with "grey16", the  debugging code in my app looks like this:

```ts
  const orgImg = imageCache(path);
  const resized = await orgImg.resize({
    width: maxWidthOrHeight,
    height: maxWidthOrHeight,
    fit: 'inside',
    withoutEnlargement: true,
  });
  const reverted = await new Promise((resolve) =>
    resized.toBuffer(async (err, data) => {
      resolve(await (await sharp(data).metadata()).space);
    }),
  );
  console.log({ org: (await orgImg.metadata()).space, resized: (await resized.metadata()).space, reverted });
```

In v0.29.0 I get the output:
`{ org: 'grey16', resized: 'grey16', reverted: 'grey16' }`

while if I upgrade to v0.29.1 I get:
`{ org: 'grey16', resized: 'grey16', reverted: 'b-w' }`

is this just as expected?

gforge:
So I've been trying to dig a little deeper into this rabbit hole... Are these assumptions correct?

* A sharp object that is called `toBuffer()` on will be converted to 8-bit format
[...]

PR #2958, opened 2021-10-31
Addition of pngBitdepth option to png

gforge:
The suggested fix for issue #2934. Note that this causes warning of type:

> (process:25977): GLib-GObject-WARNING **: 22:25:02.057: value "16" of type 'gint' is invalid or out of range for property 'bitdepth' of type 'gint'

### DIFF
--- a/lib/output.js
+++ b/lib/output.js
@@ -393,6 +393,13 @@ function png (options) {
         throw is.invalidParameterError('compressionLevel', 'integer between 0 and 9', options.compressionLevel);
       }
     }
+    if (is.defined(options.pngBitdepth)) {
+      if ([8, 16].includes(options.pngBitdepth)) {
+        this.options.pngBitdepth = options.pngBitdepth;
+      } else {
+        throw is.invalidParameterError('pngBitdepth', 'integer of 8 or 16', options.png);
+      }
+    }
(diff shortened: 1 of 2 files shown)


lovell:
I've updated the branch for this PR to use `main`, which provides libvips 8.12.0-rc1.

Please can you update the code to take an approach similar to https://github.com/lovell/sharp/issues/2934#issuecomment-959852511

gforge:
Thanks, I've updated the code with the `Is16Bit` function. The test still complains though with:

> (process:38668): GLib-GObject-WARNING **: 17:16:45.992: value "16" of type 'gint' is invalid or out of range for property 'bitdepth' of type 'gint'

but I may have missed something when I compiled the library.

lovell:
Thanks for the updates, please can you rebase your branch against the upstream `main` rather than `master`.

gforge:
Ok, I think I rebased it although I have to admit I have absolutely no idea if everything turned out fine.... at one time I think I did an empty commit and then stuff seemed to work but if it doesn't perhaps it's easier add the fixes manually.

lovell:
Thank you, landed via https://github.com/lovell/sharp/commit/2deced0fb93f51667d75250457f08d3d62f5207d

PR #2971, opened 2021-11-11
Remove double entry for mbklein

mbklein:
My name was already in `humans.txt` from the merge of #1483.


lovell:
Michael, your contributions will always count as double ;)

PR #2992, opened 2021-11-28
Heroku docs: Add VIPS buildpack dependencies

josephfrazier:
I used https://github.com/brandoncc/heroku-buildpack-vips to install VIPS in my heroku app, and discovered that I needed to install some extra packages to get it working. I opened up a PR on that repo, but then realized that it might be more relevant here.

There's some related documentation here: https://gist.github.com/bigtiger/be41fffcd103d7e98c5fe4fdff0af3f9

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -199,6 +199,16 @@ Set
 [NODE_MODULES_CACHE](https://devcenter.heroku.com/articles/nodejs-support#cache-behavior)
 to `false` when using the `yarn` package manager.
 
+If you are using https://github.com/brandoncc/heroku-buildpack-vips to install VIPS, you may also need the following packages:
+
+- libglib2.0-0
+- libglib2.0-dev
+- libpoppler-glib8
+- libpoppler-glib-dev
+- libheif-dev
+- libfftw3-dev
+- libwebp-dev
+
 ## AWS Lambda
 
 The `node_modules` directory of the


lovell:
Hi, thanks for the PR, as far as I know the buildpack you reference isn't required to use sharp on Heroku. Is my understanding correct?

josephfrazier:
If I recall correctly, sharp bundles VIPS by default, so it's not generally required. However, my use case involves decoding HEIC images, so I needed to install it separately. Maybe I should make that clearer in the docs?

lovell:
Thanks for confirming. Given this is not a requirement to install sharp on Heroku, I'd prefer not to mention it in the sharp docs otherwise people will get confused and I'll end up fielding support questions relating to its maintenance that I can't help with.

In addition, this buildpack is one of many forks of https://github.com/alex88/heroku-buildpack-vips/network/members all with slightly differing dependencies and installation requirements and I'd prefer if each of these were self documenting rather than trying to cover every possible combination in the sharp docs.

Perhaps you could maintain an active fork that is well-documented and explicitly mentions sharp. If so, you should highlight that your buildpack makes all code running in it GPL (e.g. poppler, fftw). It's also worth noting that if you're planning to distribute binaries that can en/docode HEIC then you will be subject to HEVC patent claims.

josephfrazier:
Thanks for the detailed response, I see why you wouldn't want to put these docs here, and agree that a fork might be a better place for me to put them. Also appreciate the reminder about HEVC patent claims! I'll go ahead and close this, thanks again for the consideration

Issue #2999, opened 2021-12-03
Missing error message when destination folder does not exist


albancho:
sharp is working well when the destination folder exists. But when the destination folder does not, the code crashes, with a generic error, which doesn't state anything about the issue.

here is the code I'm using:

```
console.log('imgToResize', imgToResize); // C:/my/path/to/origin/my-img.jpg
console.log('img960', img960); // C:/my/path/to/960/my-img.jpg (folder "960" doesn't exist)
console.log('dimensionsFull', dimensionsFull); // { height: 1335, width: 2000, type: 'jpg' }
console.log('new dim: 960x', Math.ceil(dimensionsFull.height * 960 / dimensionsFull.width)); // 641
const sharpOptions = {
    fit: 'contain',
    background: {
       r: 0,
       g: 0,
       b: 0,
       alpha: 0
    }
}
await sharp(imgToResize)
                    .resize(960, Math.ceil(dimensionsFull.height * 960 / dimensionsFull.width), sharpOptions)
                    .toFile(img960);
```

```
npm ERR! code ELIFECYCLE
npm ERR! errno 3221226505
npm ERR! my-app-name@0.0.0 imgmin: `node scripts/my-script.js`
npm ERR! Exit status 3221226505
npm ERR!
npm ERR! Failed at the my-app-name@0.0.0 my-script script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\my-user\AppData\Roaming\npm-cache\_logs\xxx-debug.log
```

The issue is solved if I add this line:
```
fse.ensureDirSync(path.dirname(img960));
```

I think sharp should check if destination folder exists at init (before sharping the actual img), and if not, throw an error which states the cause. It should be even better is sharp could make the folder if it doesn't exists, but not sure if it should.

lovell:
I see the following error when attempting to write to a missing directory:
```
Error: does-not-exist/out.png: unable to open for write
unix error: No such file or directory
```
Perhaps you need to add some try/catch error handling to your script?

https://sharp.pixelplumbing.com/api-output#tofile

> The caller is responsible for ensuring directory structures and permissions exist.

albancho:
Thank you for your suggestion, but I had already tried this code, with no better message:

```
            try {
                // no console.log triggered
                sharp(imgToResize)
[...]

PR #3002, opened 2021-12-06
Windows: make invalid param errors recoverable by default

kleisauke:
Context: #2999 and https://github.com/libvips/libvips/pull/2571.


lovell:
Brilliant, thanks Kleis.

PR #3006, opened 2021-12-07
added support for graphicsMagicks < resize behaviour closes #1085

### LINKED ISSUE #1085: Add option not to reduce images
Hey,

In my opinion, as there is the withoutEnlargement() option, it should also be the opposite of it. (withoutReduction() or something withoutShrink())
Is it expected to be implemented? Do you need help with this?

thx,
Ran

christopherbradleybanks:


lovell:
Hi Chris, thank you very much for this PR. Please can you rebase against the `main` branch as that will form the next release.

lovell (changes requested):
I think you'll need to set a default value for `withoutReduction` in the same way as `withoutEnlargement`.

https://github.com/lovell/sharp/blob/513fb40f40d0464931971ada333a36afc287fa7b/lib/constructor.js#L168

christopherbradleybanks:
> I think you'll need to set a default value for `withoutReduction` in the same way as `withoutEnlargement`.
> 
> https://github.com/lovell/sharp/blob/513fb40f40d0464931971ada333a36afc287fa7b/lib/constructor.js#L168

totally an oversight on my part, I've made the correction

christopherbradleybanks:
> Hi Chris, thank you very much for this PR. Please can you rebase against the `main` branch as that will form the next release.
> 
> UPDATE: I changed the base branch via GitHub and it seems to have handled this without conflict, so a rebase might not be required.

Thanks for making that adjustment, I wasn't sure if I should put it against master or main but your contribution doc seemed to suggest it should be master. I tested it against both just to be sure and am glad there wasn't any issue rebasing.

test/unit/resize.js
```diff
@@ -357,6 +357,127 @@ describe('Resize dimensions', function () {
...
+  });
+
+  it('Do resize when fit = outside and withoutReduction are true and input height > height and input width > width ', function (done) {
+    sharp(fixtures.inputJpg)
+      .resize(3000, 3000, { fit: 'outside', withoutReduction: true })
+      .toBuffer(function (err, data, info) {
+        if (err) throw err;
+        assert.strictEqual(true, data.length > 0);
+        assert.strictEqual('jpeg', info.format);
+        assert.strictEqual(2725, info.width);
```
lovell, line 475: The assertion on width appears to be failing:
```
  1) Resize dimensions
       Do resize when fit = outside and withoutReduction are true and input height > height and input width > width :

      Uncaught AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

2725 !== 3674

      + expected - actual

      -2725
      +3674
      
      at Object.<anonymous> (test/unit/resize.js:475:16)
```

lovell (changes requested):
Thanks for the update, it looks like there's one test that needs fixing up, otherwise this is good to go.

test/unit/resize.js
christopherbradleybanks: > The assertion on width appears to be failing:
> 
> ```
>   1) Resize dimensions
>        Do resize when fit = outside and withoutReduction are true and input height > height and input width > width :
> 
>       Uncaught AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
> 
> 2725 !== 3674
> 
>       + expected - actual
> 
>       -2725
>       +3674
>       
>       at Object.<anonymous> (test/unit/resize.js:475:16)
> ```

Good catch assertion corrected.

lovell (approved):

lovell:
Brilliant, thank you, this will be in v0.30.0

PR #3010, opened 2021-12-11
Ensure operations without multi-page support reject rather than silently fail

lovell:
@kleisauke as we discussed in https://github.com/lovell/sharp/pull/2789#issuecomment-985072211

I didn't apply this to the `composite` operation as this should still work, plus can be used to watermark an animation via tiled composition.


kleisauke (approved):
LGTM!

PR #3023, opened 2021-12-23
Added resolution unit option in tiff method

ompal-sisodiya:
- Added resolution unit in option.param while using tiff method
- Added required test cases


lovell (changes requested):
Thank you Ompal for the PR, looks good. Please can you add a unit test that verifies the EXIF metadata in the output has the expected resolution unit based on this new option.

lovell:
Sorry, I should have been clearer, I was thinking of a unit test that verifies the EXIF metadata of the output image (rather than just inspecting internal options), perhaps using an approach similar to the existing metadata tests:

https://github.com/lovell/sharp/blob/1ff84b20b7d9b351786fb06899255cea59aa123c/test/unit/metadata.js#L53-L56

ompal-sisodiya:
@lovell  All the required change are done on our end. could you review once.

nishantjiit:
@lovell : Can you please review and approve this PR. We are waiting to use the resolution unit feature with our application. Thanks in advance !!!

lovell:
@ompal-sisodiya Thanks for the updates. I don't see any verification of output image EXIF metadata in the newly-added tests; am I missing something? (I note you've added to an existing unit test, but that logic is unrelated to this PR, I was merely using it as an example of using `exif-reader` to parse the output.)

ompal-sisodiya:
@lovell Thanks for your response. please review the below test for the verification of output image with EXIF metadata.
    if this is fine then i will commit the code else you can help me with the  test.   
```
it('TIFF setting resolution unit(cm)', async () => {
    await sharp(fixtures.inputJpgWithExif)
      .tiff({ resolutionUnit: 'inch' })
      .metadata(function (err, metadata) {
        if (err) throw err;
        // EXIF
        assert.strictEqual('object', typeof metadata.exif);
        assert.strictEqual(true, metadata.exif instanceof Buffer);
        const exif = exifReader(metadata.exif);
        assert.strictEqual('object', typeof exif);
        assert.strictEqual('object', typeof exif.image);
        assert.strictEqual('number', typeof exif.image.XResolution);
        assert.strictEqual('number', typeof exif.image.ResolutionUnit);
      });
  });

lovell:
@ompal-sisodiya Please can you add an assertion on the value of `exif.image.ResolutionUnit` (as well as the type).

nishantjiit:
@lovell : Can you please review the PR?

ompal-sisodiya:
@lovell  Could please review pr?

lovell:
Thanks again for taking the time to submit a PR, I will review this when I am ready to do so. Coordinated pestering won't make that happen any sooner.

lovell:
Landed via commit https://github.com/lovell/sharp/commit/f7bed69ffb5a533881faeb02163111deb4ff219a

I also added the (normalised) `resolutionUnit` property to the response from `metadata()`, which had the added advantage of making the tests a bit simpler.

This will be in v0.30.0, thanks again for the PR.

PR #3057, opened 2022-01-27
Vulnerability: update simple-get package version

simon-tannai:
According to [Snyk's vulnerability page](https://security.snyk.io/vuln/SNYK-JS-SIMPLEGET-2361683), simply updating `simple-get` package version to fix this vulnerability.


lovell:
Thanks for the PR.

The current range already includes the fixed version so automated audit tools can be silenced with a lockfile update.

I've also checked the "vulnerability" and it has no impact on sharp users.

PR #3059, opened 2022-01-27
Take page parameter into account when all frames are read

kleisauke:
Otherwise, it could lead to funny results.

<details>
  <summary>Before / after examples</summary>

```bash
$ curl -LO https://storage.googleapis.com/downloads.webmproject.org/webp/images/dancing_banana2.lossless.webp
$ node -e "require('sharp')('dancing_banana2.lossless.webp', { pages: -1, page: 2 }).resize(128).toFile('x.gif')"
```
| Before | After |
| :---: |  :---: |
| before | after |

</details>

A test case was added to prevent further regressions.


lovell:
Lovely stuff, thank you Kleis.

Issue #3061, opened 2022-01-27
Preserve delay when resizing animated images


janicduplessis:
I've been testing the 0.30 version in the main branch for resizing animated images. The only issue I encountered is that delay seem to be lost when resizing / converting to webp.

Not sure exactly where the delay is lost, but I've been doing the following operations:

- input gif
- resize
- output webp

```js
const image = sharp(someGif, { animated: true });
const processed = image.rotate().resize(200, 200).webp();
```

The result animates, but a lot faster than the source.

As a workaround I used the `metadata` function to get the delay and pass that to .webp.

```js
const image = sharp(someGif, { animated: true });
const meta = await image.metadata();
const processed = image.rotate().resize(200, 200).webp({ delay: meta.delay });
```

kleisauke:
Are you able to provide the test image that exhibits this behavior? I could not reproduce this with:
```bash
$ curl -LO https://storage.googleapis.com/downloads.webmproject.org/webp/images/dancing-banana.gif
$ vipsheader dancing-banana.gif -f delay
100 100 100 100 100 100 100 100
$ node -e "require('sharp')('dancing-banana.gif', { animated: true }).rotate().resize(200, 200).toFile('x.webp')"
$ vipsheader x.webp -f delay
100 100 100 100 100 100 100 100
```

The only bug that I'm aware of is https://github.com/libvips/libvips/issues/2582, but that was fixed in libvips v8.12.2 and causes the animated image to be slower (not faster).

janicduplessis:
Hey @kleisauke! Thanks for having a look.

Here's the image I had the issue with: https://cdn.th3rdwave.coffee/merchants/24IRazCBNSlALF2blNYuCEKhkPX.gif

kleisauke:
Ah, the image I tested with had all delays set to 100ms, which is also the default duration for WebP images.
https://github.com/libvips/libvips/blob/0a7e96a4e2f97dc2b933ebeaa763ab573adb7f6d/libvips/foreign/vips2webp.c#L336-L341

I was able to reproduce this with the image you provided. PR #3062 should fix this. Thanks for reporting/testing!

janicduplessis:
Thanks for the quick fix, will test this today!

janicduplessis:
I can confirm this works

PR #3062, opened 2022-01-28
Ensure delays of the origin image is preserved

kleisauke:
By not initializing the delay vector, since we assume it's empty by default  in `SetAnimationProperties`.
https://github.com/lovell/sharp/blob/68823a5edb0c913fb230cd36c1ce22554e48f9ad/src/common.cc#L495-L503

A test case was added to prevent further regressions.

See: #3061.


lovell:
Bedankt!

PR #3080, opened 2022-02-08
Update tests and comments after 7faacd9

kleisauke:


lovell:
The arm64 builds will fail as it looks like there's a possible lcms2 rounding error difference that needs a slight threshold increase (the default threshold for most tests is 7, so this is already very strict) - see https://github.com/lovell/sharp/commit/02e0c2dfc98ae639c987922d46fabdc27f4dd1b0

lovell:
Thank you Kleis!

Issue #3067, opened 2022-02-02
Possible regression in `.resize` with `withoutEnlargement: true` and `fit: sharp.fit.contain`


medanat:
Hey, after upgrading to 0.30.0 we noticed an issue with the resizing method with `withoutEnlargement: true` and `fit: sharp.fit.contain`. The images are being enlarged despite the `withoutEnlargement` flag. 

The image correctly ignores the new width when `fit: sharp.fit.contain` is removed. Not sure if this is intended behaviour or a regression. I've added the failing test steps below.

Super stoked to see 0.30.0 come out 😄, thanks!

Are you using the latest version? Is the version currently in use as reported by `npm ls sharp` the same as the latest version as reported by `npm view sharp dist-tags.latest`?

Yes

```
 sharp@0.30.0
 0.30.0
 ```

What are the steps to reproduce?

- resize image to a larger width than original with `withoutEnlargement: true` and `fit: sharp.fit.contain`

What is the expected behaviour?

Image width should not grow bigger than original

Are you able to provide a minimal, standalone code sample, without other dependencies, that demonstrates this problem?

I've included the test (passing in 0.29.3/failing in 0.30.0)

```
  it('does not make an image bigger', async () => {
    const { readFileSync } = require('fs');
    const sharp = require('sharp');
    const image = readFileSync('image-1.png');

    const { info: { width: originalWidth, height: originalHeight } } = await sharp(image).toBuffer({ resolveWithObject: true });
    const newWidth = originalWidth + 333;
    const { info: { width, height } } = await sharp(image)
      .resize(newWidth, undefined, {
        withoutEnlargement: true,
        fit: sharp.fit.contain
      })
      .toBuffer({ resolveWithObject: true });

    expect(width).to.not.eq(newWidth);
    expect(width).to.eq(originalWidth);
    expect(height).to.eq(originalHeight);
  });
```

Are you able to provide a sample image that helps explain the problem?

![image-1](https://user-images.githubusercontent.com/969646/152117082-73f6f420-b780-468b-a992-278d55b672dd.png)

What is the output of running `npx envinfo --binaries --system`?

```
  System:
    OS: Linux 5.10 Alpine Linux
    CPU: (5) arm64 unknown
    Memory: 2.06 GB / 3.84 GB
    Container: Yes
    Shell: 1.34.1 - /bin/ash
  Binaries:
    Node: 14.18.2 - /usr/local/bin/node
    Yarn: 1.22.15 - /usr/local/bin/yarn
[...]

PR #3081, opened 2022-02-08
Move `withoutReduction` logic to `ResolveShrink`

kleisauke:
Ensures that it does not interfere with cropping/embedding.

See: #3067.


lovell:
(Zoals altijd) bedankt!

Issue #3086, opened 2022-02-11
Enhancement: statically link Windows C++ runtime


dertieran:
I ran into a very strange issue when upgrading from `sharp@0.29.3` to `sharp@0.30.1` on Windows.

```sh
Error:
Something went wrong installing the "sharp" module

The specified module could not be found.
\\?\C:\...\node_modules\sharp\build\Release\sharp-win32-x64.node
```

But when I check the diorectory the module exists

```
ls node_modules\sharp\build\Release

    Directory:
C:\...\node_modules\sharp\build\Release

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a---            2/1/2022 10:49 AM        1580544 libglib-2.0-0.dll
-a---            2/1/2022 10:49 AM         292352 libgobject-2.0-0.dll
-a---            2/1/2022 10:49 AM       20575232 libvips-42.dll
-a---            2/9/2022  5:47 PM         257024 libvips-cpp.dll
-a---            2/9/2022  5:47 PM         226304 sharp-win32-x64.node
```

I also check other Issues I found related to this

- I don't have `canvas` installed
- Deleting `_prebuilds` doesn't help
- Installing with different flags doesn't help

The only thing that worked was downgrading to `sharp@0.29.3` again.

### System Info:
```
npx envinfo --binaries --system

  System:
    OS: Windows 10 10.0.19043
    CPU: (2) x64 Intel(R) Celeron(R) CPU G3900 @ 2.80GHz
    Memory: 7.34 GB / 11.87 GB
  Binaries:
    Node: 16.14.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.15 - C:\Program Files\nodejs\yarn.CMD
    npm: 8.3.1 - C:\Program Files\nodejs\npm.CMD
```

lovell:
Do you have any other dependencies in the same `node_modules` directory that use sharp? What's the output of `npm ls sharp`?

Doko-Demo-Doa:
I have the same issue. Output for `npm ls sharp` is:

```
└── sharp@0.30.1
```

No other module uses sharp.

mariusa:
FYI works for me on linux

Palad1NPT:
any news on this?

wardpeet:
I don't have the issue on my windows machine 🤔 - what about `npm rebuild sharp`

lovell:
The two main reasons for seeing this error on Windows are:

[...]

PR #3110, opened 2022-02-25
Windows: ensure C++ runtime is linked statically

kleisauke:
And remove the empty invalid parameter handler, which should
be present in the C layer instead.

This partially reverts commit 659cdabd8e8c284901a0ea17ccdcb13e17350a8e, the added test case in
that commit is still preserved.

See: #3086.


lovell:
Bedankt Kleis!

Issue #3124, opened 2022-03-09
Get actual (rotated) width/height from metadata


mindplay-dk:
### What are you trying to achieve?

In an image server project, I need to check the image dimensions via `metadata()`, so I can decide whether to resize an image, or instead respond with a redirect.

### When you searched for similar feature requests, what did you find that might be related?

There is `rotate()`, which will fix the rotation - the script will use this if it decides to resize the image, but I'm trying to avoid loading images if I'm not going to resize them, so I need to use `metadata()` instead.

The information is of course available in the metadata object as `width`, `height` and `orientation` - but the documentation explicitly says that `orientation` is not taken into account with regards to `width` and `height`, which that means the *meaning* of `width` and `height` might actually be reversed.

### What would you expect the API to look like?

I would propose adding a second pair of properties for the actual (right-side-up) dimensions.

I don't know how you would name these... It's already a bit confusing to have `width` and `height` that might actually be reversed - but changing that would be a nasty breaking change, so I'm not suggesting that.

Instead maybe a new pair of properties named `normalWidth`, `rotatedWidth` or `actualWidth`, etc.?

### What alternatives have you considered?

I know I can implement this myself by checking the `orientation` - but it seems like this requirement is going to be common enough that it's worth adding?

Alternatively, the documentation could include an example of this? It currently states that "EXIF orientation is not taken into consideration", but doesn't offer any advice on how to deal with that.

If I had to guess, [based on this answer](https://stackoverflow.com/questions/7489742/php-read-exif-data-and-adjust-orientation), I would say maybe something along the lines of...

```js
function getNormalSize({ width, height, orientation }) {
  return orientation >= 5
    ? { width: height, height: width }
    : { width, height };
}
```

Did I get it right? EXIF orientation is confusing, and I *know* I'm going to get this wrong, I've done it before. 😅

lovell:
A PR to add your example to the metadata documentation somewhere around [here](https://github.com/lovell/sharp/blob/23033e2050d9ef099ecb02dde0f56f9f23e8d30d/lib/input.js#L345) would be most welcome, thank you.

mindplay-dk:
Still considering this for the API, or do you want to close it? 🙂

lovell:
Let's close for now and see how your documentation addition goes, thanks again for doing that.

[...]

PR #3127, opened 2022-03-09
add example of how to get the right-side-up width/height

mindplay-dk:
See #3124


lovell:
Tak Rasmus!

mindplay-dk:
Selv tak 😄

PR #3132, opened 2022-03-11
Check loaded bindings not only on Windows

marvinroger:


lovell:
Perhaps we should only additionally show this help when the error message mentions a symbol-related problem?
```js
if (process.platform === 'win32' || /symbol/.test(err.message)) { ...
```

marvinroger:
I rebased and amended the commit with your suggestion, therefore I force-pushed to have a single commit as specified in the CONTRIBUTING.md

lovell:
C'est super, merci Marvin.

marvinroger:
Merci à toi !

PR #3146, opened 2022-03-19
[#3145] byteOffset of typedArray is not handled properly

### LINKED ISSUE #3145: byteOffset of typedArray is not handled properly
https://github.com/lovell/sharp/blob/b609df4b48b46b6f6e2f0a9d81bb950fae80236f/lib/input.js#L38-L42

i think L42 should be like below to handle typedArray with byteOffset.

```
inputDescriptor.buffer = Buffer.from(input.buffer, input.byteOffset, input.byteLength);
```

using byteOffset seems unusal case but it will be nice to handle it.

codepage949:
Closes #3145

### DIFF
--- a/lib/input.js
+++ b/lib/input.js
@@ -39,7 +39,7 @@ function _createInputDescriptor (input, inputOptions, containerOptions) {
     if (input.length === 0) {
       throw Error('Input Bit Array is empty');
     }
-    inputDescriptor.buffer = Buffer.from(input.buffer);
+    inputDescriptor.buffer = Buffer.from(input.buffer, input.byteOffset, input.byteLength);
   } else if (is.plainObject(input) && !is.defined(inputOptions)) {
     // Plain Object descriptor, e.g. create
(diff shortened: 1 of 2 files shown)


lovell:
Thank you very much for the PR. Is it possible to add a test for this?

codepage949:
@lovell @gacord23 added test.

lovell:
Thanks for the updates. It looks like this PR was affected by today's [GitHub Actions outage](https://www.githubstatus.com/history) as the CI job is still in a queued and non-cancellable state - see https://github.com/lovell/sharp/actions/runs/2021845098

Hopefully the backlog will clear soon and we can get this merged.

lovell:
It looks like those CI jobs aren't going to run. Please can you rebase against the latest upstream/main branch and push to see if that might be enough to force the tests to run.

codepage949:
ok, i did rebase.

lovell:
Thank you!

PR #3160, opened 2022-03-28
Allow specifying libc for Linux install

### LINKED ISSUE #3159: Cross-platform Linux installation, target uses glibc but host uses musl
## Possible install-time or require-time problem

- [X] I have read the [documentation relating to installation](https://sharp.pixelplumbing.com/install).
- [ ] I have ensured that the architecture and platform of Node.js used for `npm install` is the same as the architecture and platform of Node.js used at runtime.

- [X] I am using the latest version of `sharp` as reported by `npm view sharp dist-tags.latest`.

If you cannot confirm this, please upgrade to the latest version and try again before opening an issue.

If you are using another package which depends on a version of `sharp` that is not the latest, please open an issue against that package instead.

### Is this a problem with filesystem permissions?

If you are using npm v6 or earlier and installing as a `root` or `sudo` user, have you tried with the `npm install --unsafe-perm` flag?

If you are using npm v7 or later, does the user running `npm install` own the directory it is run in?

If you are using the `ignore-scripts` feature of `npm`, have you tried with the `npm install --ignore-scripts=false` flag?

### What is the complete output of running `npm install --verbose sharp`?

<details>

</details>

I'm trying to cross-platfrom install from an alpine image targeting both win32-x64 and linux-x64 platform, but both are failing because during the installation procedure musl is detected as libc, and the binary installed are for musl.

Instead, when I cross-platform install when using windows targeting linux, it works fine.

Am I doing something wrong here, or is using an alpine image not supported for cross-compiling to linux (glibc)?

When I run:
npm install --arch=x64 --platform=linux sharp
Installation seems ok, but I found the following path revealing something not expected:

"/node_modules/sharp/vendor/8.12.2/linuxmusl-x64"

joonamo:
This is a sibling PR to PR in upstream dependency [prebuild-install](https://github.com/prebuild/prebuild-install/pull/176) and should fix #3159.

### DIFF
--- a/lib/platform.js
+++ b/lib/platform.js
@@ -7,10 +7,11 @@ const env = process.env;
 module.exports = function () {
   const arch = env.npm_config_arch || process.arch;
   const platform = env.npm_config_platform || process.platform;
-  /* istanbul ignore next */
-  const libc = (platform === 'linux' && detectLibc.isNonGlibcLinuxSync()) ? detectLibc.familySync() : '';
+  const libc = process.env.LIBC || process.env.npm_config_libc || 
+    (detectLibc.isNonGlibcLinuxSync() ? detectLibc.familySync() : '');
+  const libcId = platform !== 'linux' || libc === detectLibc.GLIBC ? '' : libc;
 
-  const platformId = [`${platform}${libc}`];
+  const platformId = [`${platform}${libcId}`];
 
   if (arch === 'arm') {
     const fallback = process.versions.electron ? '7' : '6';
--- a/lib/platform.js
+++ b/lib/platform.js
@@ -7,7 +7,8 @@ const env = process.env;
 module.exports = function () {
   const arch = env.npm_config_arch || process.arch;
   const platform = env.npm_config_platform || process.platform;
-  const libc = process.env.LIBC || process.env.npm_config_libc || 
+  const libc = process.env.npm_config_libc || 
+    /* istanbul ignore next */
     (detectLibc.isNonGlibcLinuxSync() ? detectLibc.familySync() : '');
   const libcId = platform !== 'linux' || libc === detectLibc.GLIBC ? '' : libc;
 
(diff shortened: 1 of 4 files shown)


joonamo:
@lovell This PR is now ready for review. You can find a test case for darwin and alpine linux in [this gist](https://gist.github.com/joonamo/ce903a0d496e2460e080347ff4fd719e). Let me know if you would need some further verification.

lib/platform.js
```diff
@@ -7,10 +7,11 @@ const env = process.env;
 module.exports = function () {
   const arch = env.npm_config_arch || process.arch;
   const platform = env.npm_config_platform || process.platform;
-  /* istanbul ignore next */
-  const libc = (platform === 'linux' && detectLibc.isNonGlibcLinuxSync()) ? detectLibc.familySync() : '';
+  const libc = process.env.LIBC || process.env.npm_config_libc ||
```
lovell, line 10: Do we need to check both of these environment variables or would `env.npm_config_libc` suffice on its own?

lovell (commented):
Thank you very much for the PR, I've left one question inline. It might also be worth adding a unit test for overriding `npm_config_libc` to https://github.com/lovell/sharp/blob/main/test/unit/platform.js for consistency with the other `npm_config_*` env vars we support.

lib/platform.js
joonamo: prebuild-install reads both LIBC-env value and npm_config_libc in this order. I made it this way to align with prebuild-install scripts, but if you want, LIBC-variable can be removed. I left it there mainly for backwards compatibility and their documentation no longer mentions it.

joonamo: After considering this, I removed reading of LIBC env value completely. Sharp installation has never supported it in the past and prebuild-install documentation now instructs the use of `--libc` or `npm_config_libc ` env value.

lovell:
Kiitos!

joonamo:
Good morning! Was this included in v0.30.5? I can't see it in the changelog 🤔

lovell:
Yes, sorry, I somehow missed adding it to the changelog. It's there now via commit https://github.com/lovell/sharp/commit/93b29057e4d759ff9251cdef374c62416a16d26b - thanks again Joona for all your work on this!

PR #3178, opened 2022-04-04
Add missing file name to 'Input file is missing' error message

### LINKED ISSUE #2360: Show meaningful error message on missing file
What are you trying to achieve?
I have the following sequence using sharp
```js
const result = await sharp(inputBuffer)
        .resize(460)
        .composite([{ input: watermark, gravity: 'center' }])
        .toFormat('webp', { quality: 90 })
        .toBuffer();
```
Everything should work as expected, the issue is:
- I've actually moved the current file where this processing takes place, therefore the `watermark` input file has a different path.
- One could argue I should use fully qualified paths but the issue I have is with the error message

```
Error: Input file is missing
```

I Found it super strange as this used to work just fine, but outputting this generic error got me through a google+stackoverflow+this-repo-issues rabbit hole thinking it was actually an issue with the `inputBuffer`.

### Suggestion

A better output would be 
```
Error: Could not <composite|process|etc> Input file at <provided path> is missing
```

Brodan:
Fixes #2360

This PR updates the error message 'Input file is missing' to include the file that is missing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,8 @@
     "Taneli Vatanen <taneli.vatanen@gmail.com>",
     "Joris Dugué <zaruike10@gmail.com>",
     "Chris Banks <christopher.bradley.banks@gmail.com>",
-    "Ompal Singh <ompal.hitm09@gmail.com>"
+    "Ompal Singh <ompal.hitm09@gmail.com>",
+    "Brodan <christopher.hranj@gmail.com"
   ],
   "scripts": {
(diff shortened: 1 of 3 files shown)


Brodan:
I tested it locally within the module and it seemed to work as expected. However, when I tried to test it in a separate local project I was seeing some weird behavior. I feel like it may have been related to how I was building sharp locally or perhaps I was using `npm pack` incorrectly.

lovell:
This is great, very helpful, thank you Chris.

Issue #3188, opened 2022-04-12
Unable to install with local custom-prebuilt-binaries


ankurparihar:
## Not able to use locally downloaded prebuilt binaries

- [x] I have read the [documentation relating to installation](https://sharp.pixelplumbing.com/install).
- [x] I have ensured that the architecture and platform of Node.js used for `npm install` is the same as the architecture and platform of Node.js used at runtime.

### Are you using the latest version of sharp?

- [x] I am using the latest version of `sharp` as reported by `npm view sharp dist-tags.latest`.

If you cannot confirm this, please upgrade to the latest version and try again before opening an issue.

If you are using another package which depends on a version of `sharp` that is not the latest, please open an issue against that package instead.

### Is this a problem with filesystem permissions?

If you are using npm v6 or earlier and installing as a `root` or `sudo` user, have you tried with the `npm install --unsafe-perm` flag?

If you are using npm v7 or later, does the user running `npm install` own the directory it is run in?

If you are using the `ignore-scripts` feature of `npm`, have you tried with the `npm install --ignore-scripts=false` flag?

### What is the complete output of running `npm install --verbose sharp`?

 > I am running this to also specify local binary directory path 
`export npm_config_sharp_local_prebuilds="/Users/ankur/workspace/phonepe/test-sharp/libs" && npm install --verbose sharp`

<details>

```
// Trimming a few good lines from above
npm info lifecycle prebuild-install@7.0.1~install: prebuild-install@7.0.1
npm info lifecycle sharp@0.30.3~install: sharp@0.30.3

> sharp@0.30.3 install /Users/ankur/workspace/phonepe/test-sharp/node_modules/sharp
> (node install/libvips && node install/dll-copy && prebuild-install) || (node install/can-compile && node-gyp rebuild && node install/dll-copy)

sharp: Downloading https://github.com/lovell/sharp-libvips/releases/download/v8.12.2/libvips-8.12.2-darwin-x64.tar.br
sharp: Installation error: connect ECONNREFUSED 127.0.0.1:443
sharp: Please see https://sharp.pixelplumbing.com/install for required dependencies
npm verb lifecycle sharp@0.30.3~install: unsafe-perm in lifecycle true
[...]

PR #3196, opened 2022-04-18
Feature to use locally downloaded libvips

ankurparihar:
Changes include

- Feature to install the prebuilt libvips binaries from a directory on the local filesystem
- Installation doc update
  - Separate instructions for prebuilt sharp and libvips binaries to avoid confusion
  - Added instructions to install prebuilt libvips binaries from local filesystem directory

Closes https://github.com/lovell/sharp/issues/3188

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -135,8 +135,8 @@ set the `sharp_libvips_binary_host` npm config option
 or the `npm_config_sharp_libvips_binary_host` environment variable.
 
 To install the prebuilt libvips binaries from a directory on the local filesystem,
-set the `sharp_local_libvips_dir` npm config option
-or the `npm_config_sharp_local_libvips_dir` environment variable.
+set the `sharp_libvips_local_prebuilds` npm config option
+or the `npm_config_sharp_libvips_local_prebuilds` environment variable.
 
 The version subpath and file name are appended to these.
 For example, if `sharp_libvips_binary_host` is set to `https://hostname/path`
--- a/docs/install.md
+++ b/docs/install.md
@@ -118,6 +118,8 @@ Building from source requires:
 
 This is an advanced approach that most people will not require.
 
+### Prebuilt sharp binaries
+
 To install the prebuilt sharp binaries from a custom URL,
 set the `sharp_binary_host` npm config option
 or the `npm_config_sharp_binary_host` environment variable.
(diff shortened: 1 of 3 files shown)


ankurparihar:
Hi @lovell 

Created this PR for https://github.com/lovell/sharp/issues/3188
Let me know in case of any issues.

Thanks

docs/install.md
```diff
@@ -126,10 +128,16 @@ To install the prebuilt sharp binaries from a directory on the local filesystem,
...
 or the `npm_config_sharp_local_prebuilds` environment variable.
 
+### Prebuilt libvips binaries
+
 To install the prebuilt libvips binaries from a custom URL,
 set the `sharp_libvips_binary_host` npm config option
 or the `npm_config_sharp_libvips_binary_host` environment variable.
 
+To install the prebuilt libvips binaries from a directory on the local filesystem,
+set the `sharp_local_libvips_dir` npm config option
```
lovell, line 138: Would naming this something like `sharp_libvips_local_prebuilds` provide a more consistent approach?

install/libvips.js
```diff
@@ -156,6 +157,15 @@ try {
     if (fs.existsSync(tarPathCache)) {
       libvips.log(`Using cached ${tarPathCache}`);
       extractTarball(tarPathCache, platformAndArch);
+    } else if (localLibvipsDir) {
+      const tarPathLocal = path.join(path.resolve(localLibvipsDir), `v${minimumLibvipsVersionLabelled}`, tarFilename);
+      // If localLibvipsDir is given try to use binaries from local directory
+      if (fs.existsSync(tarPathLocal)) {
```
lovell, line 163: It might be simpler to remove the exists check and always call `extractTarball`. This means when it fails the error message from `createReadStream` will provide more context e.g. ENOENT vs EACCES etc., which I think will be more helpful.

package.json
```diff
@@ -82,7 +82,8 @@
     "Joris Dugué <zaruike10@gmail.com>",
     "Chris Banks <christopher.bradley.banks@gmail.com>",
     "Ompal Singh <ompal.hitm09@gmail.com>",
-    "Brodan <christopher.hranj@gmail.com"
+    "Brodan <christopher.hranj@gmail.com",
+    "Ankur Parihar <ankur.github@gmail.com> (https://ankurparihar.github.io)"
```
lovell, line 86: Please can you remove the URL.

lovell (changes requested):
Hello, thank you very much for submitting this PR, I've left a few questions/comments inline.

package.json
ankurparihar: sure

docs/install.md
ankurparihar: updated

install/libvips.js
ankurparihar: Yes, updated

lovell (approved):

lovell:
Thank you!

ankurparihar:
Thank you for this awesome library!

PR #3201, opened 2022-04-22
Fix typo in api-output.md

humphd:
Removed an extra `)` on code example for converting to `gif`

### DIFF
--- a/docs/api-output.md
+++ b/docs/api-output.md
@@ -334,7 +334,7 @@ The first entry in the palette is reserved for transparency.
 // Convert PNG to GIF
 await sharp(pngBuffer)
   .gif()
-  .toBuffer());
+  .toBuffer();
 ```
 
 ```javascript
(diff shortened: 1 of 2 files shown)


lovell:
Thank you for fixing this David. Would you be able to update the incorrect source JSDoc too as part of this PR?

https://github.com/lovell/sharp/blob/d396a4e65d5ed744dc4300c89ee1809fd7188638/lib/output.js#L519

The `docs/api-*.md` files are auto-generated and would otherwise be overwritten again next time `npm run docs-build` is used.

https://github.com/lovell/sharp/blob/d396a4e65d5ed744dc4300c89ee1809fd7188638/package.json#L96

humphd:
Ah, thanks for pointing that out.  I've made the change.

lovell:
Brilliant, thank you!

humphd:
Actually, thank *you*! Sharp is amazing.  I've used it so many times for lots of projects.  Recently I had my cloud computing students use it for a final project, and (as always) it worked flawlessly for them, even though they were all new to it.  The docs especially set it apart.

Keep up the great work.

Issue #3024, opened 2021-12-23
Chaining rotate() and extract() not working for mirrored Exif-Orientations (Orientation 2 ,4 ,5 ,7)


phellmayr1:
**The issue is the following:**
When chaining rotate and extract the extraction result is wrong for EXIF-Orientations of 2,4,5,7.
When **not** chaining and first rotating => to Buffer => and then extracting, it works just fine.

**Example**:
Chaining Rotate and Extract **not** working:
```
        sharp(imageBuffer)
            .rotate()
            .extract({"top":0,"left":600,"width":600,"height":600})
            .toBuffer((err: Error, data: Buffer, info: OutputInfo) => {
                    //Result is wrong and mirroring is not considered
                     fs.writeFileSync('C:\\Users\\xxx\\Pictures\\sharp-bug\\' + uuid() + '.jpg', data);
         });
```

**Workaround which can be used**
```
        sharp(imageBuffer)
            .rotate()
            .toBuffer((err: Error, res: Buffer, info: OutputInfo) => {

                        sharp(res)
                            .extract({"top":0,"left":600,"width":600,"height":600})
                            .toBuffer((err: Error, data: Buffer, info: OutputInfo) => {
                                    //Result is Correct and mirroring is considered
                                     fs.writeFileSync('C:\\Users\\xxx\\Pictures\\sharp-bug\\' + uuid() + '.jpg', data);
                         }); 
         });
```

- 

**What are the steps to reproduce?**
Use an image with exif-orientation 2 ,3 ,4 or 7 and chain rotate and extract like in the above example.

**What is the expected behaviour?**
Expected behaviour: Chaining rotate and extract works the same as it does when using it after each other.

**Are you able to provide a minimal, standalone code sample, without other dependencies, that demonstrates this problem?**
see code samples above

**Are you able to provide a sample image that helps explain the problem?**
Can be reproduced with this image:
https://github.com/recurser/exif-orientation-examples/blob/master/Landscape_7.jpg

lovell:
Thanks for reporting, I agree, this looks like a bug where the operation ordering should be rotate>flip>extract whereas for this code sample it's doing rotate>extract>flip. I'll take a look.

lovell:
Commit https://github.com/lovell/sharp/commit/d8f129851110a59916cd12352b8e6b06b37411bc fixes this and adds a test that would previously have failed. This will be in v0.30.0.

lovell:
v0.30.0 now available, thanks again for reporting.

PR #3218, opened 2022-05-09
fix rotate-then-extract for EXIF orientation 2

jakob0fischl:
the fix for #3024 isn't handling exif orientation 2 correctly

orientation 2 is flipped but has no rotation and wasn't applied correctly because of that


lovell:
Thank you very much for fixing this and adding tests.

PR #3232, opened 2022-05-17
Update api-resize.md

oleg-andreyev:


lovell:
Hello, thanks for the PR, I think the original wording is correct, but perhaps could be made clearer with something like:
```diff
- when using a `fit` of `contain`
+ when `fit` is `contain`
```
Please could you make this change to the JSDocs in the source too, as the `docs/api-*.md` files are autogenerated from these.

https://github.com/lovell/sharp/blob/04c31b35a7e2d88931e1cb7cde0fe1720e4685b7/lib/resize.js#L214

lovell:
Landed via commit f903e1465e2f8cce8de1c0bc01ef79631800aa51 - thank you!

PR #3241, opened 2022-05-30
docs(input): correct getNormalSize with EXIF orientation example

Fonger:
The example is wrong. `>=` has precedence over `||` so the example always switchs the width and height if orientation has non-zero value and this leads to incorrect size.


lovell:
Good spot, thank you for fixing this.

Issue #512, opened 2016-07-15
Enhancement: expose vips_text to create an image containing rendered text


gargsms:
I was fiddling with SVG overlays, and tried a SVG containing text inside a foreignObject div, but it didn't render.
My use case is to make the text fit the container dynamically. I won't be knowing the text length beforehand, but I know how wide the container will be.
Here is the SVG image I am using: http://jsbin.com/jowaboluga/1/edit?html,output
Since there is no support of drawing text over an image, I had to make a SVG with the text.

Is there any way to achieve the desired effect?

lovell:
Hello, a quick Stack Overflow search revealed the following, which may help:

http://stackoverflow.com/questions/2938779/svg-scaling-text-to-fit-container
http://stackoverflow.com/questions/283346/how-to-resize-text-in-svg-dynamically

gargsms:
@lovell I did look up those links already. And I believe they are not catering to my use case.
I want the text to wrap by itself if it is overflowing the container - something easily achievable with foreignObject support. Anyhow, if it is not there, I will have to devise some method to make it work.

By the way, any plans to include the watermark functionality by default? https://github.com/h2non/bimg/ has the watermark support built in.

I would be more than willing to work on a patch if it is possible - need a few pointers, though.

lovell:
I suspect librsvg (as used by libvips for SVG rendering) doesn't support the foreignObject extension to SVG and it's likely *magick has the same limitation.

I'd be happy to accept a PR that exposes [vips_text](http://www.vips.ecs.soton.ac.uk/supported/current/doc/html/libvips/libvips-create.html#vips-text), a wrapper around pango, for image creation. I need to think a little about how this should fit into the existing API (e.g. should work alongside #470).

gargsms:
I should be able to work this out and submit a PR soon. Not sure about *magick, though.
Instead of using it with #470, I think it could be used better as an extension to `.overlayWith()`.

lovell:
In terms of the API to expose this, I see it working a bit like the proposed "create image" helper described in https://github.com/lovell/sharp/issues/470#issuecomment-239620622
```javascript
sharp(null, {
  text: {
    text: string,
    font: string,
    width: integer,
    align: string,
    dpi: integer,
    spacing: integer
  }
})...
```

gargsms:
I have implemented the text API, though not for the pencil branch.

[...]

PR #3252, opened 2022-06-10
Expose vips_text to create an image containing rendered text #512

brahima:
Hello @lovell ,

This a PR regarding the vips_text feature (#512)

Some notes about this PR :

- I started from your [last proposal API](https://github.com/lovell/sharp/issues/512#issuecomment-817599154)
  - What @rorz did helped me getting started and  i removed all the color related stuffs but integrated the rgba option
  - maybe the spacing parameter should be renamed as lineSpacing ?
- I did not take into consideration the [autofit_dpi](https://www.libvips.org/API/current/libvips-create.html#vips-text) parameter exported by libvips. Should i ?
- Regarding the tests
    - I have all my new tests passing but some existing ones are failing (even on your main branch). I tried with libvips installed via brew and recompiled locally with no success.
  - I added a font (Bitstream Vera from the Gnome Project) for my tests.
    - What do you think of the copyright/license ? If not ok, do you think of another font that could be embedded ? Is it in the right place (fixtures) ?
  - I was not able to run the memory-leak test also (no valgrind in mac) and i struggled a lot to install kcachegrind (macport, compile ... seems to be a known problem on last MacOs versions) with no success.

Thanks for your insights on my modifications and any advice about my macos related problems regarding the tests.

Regards,
Brahim

### DIFF
--- a/docs/api-constructor.md
+++ b/docs/api-constructor.md
@@ -151,15 +151,15 @@ await sharp({
 ```
 
 ```javascript
-// Generate a color image from text using pango markup and font
+// Generate an rgba image from text using pango markup and font
 await sharp({
   text: {
     text: '<span foreground="red">Red!</span><span background="cyan">blue</span>',
-    font: 'sans'
+    font: 'sans',
     rgba: true,
     dpi: 300
   }
-}).toFile('text_color.png');
+}).toFile('text_rgba.png');
 ```
 
 *   Throws **[Error][17]** Invalid parameters
--- a/docs/api-constructor.md
+++ b/docs/api-constructor.md
@@ -51,6 +51,18 @@ Implements the [stream.Duplex][1] class.
             *   `options.create.noise.type` **[string][12]?** type of generated noise, currently only `gaussian` is supported.
             *   `options.create.noise.mean` **[number][14]?** mean of pixels in generated noise.
             *   `options.create.noise.sigma` **[number][14]?** standard deviation of pixels in generated noise.
+    *   `options.text` **[Object][13]?** describes a new text image to be created.
+
+        *   `options.text.text` **[string][12]?** text to render as a UTF-8 string. It can contain Pango markup, for example `<i>Le</i>Monde`.
+        *   `options.text.font` **[string][12]?** font name to render with.
+        *   `options.text.fontfile` **[string][12]?** absolute filesystem path to a font file that can be used by `font`.
+        *   `options.text.width` **[number][14]** integral number of pixels to word-wrap at. Lines of text wider than this will be broken at word boundaries. (optional, default `0`)
+        *   `options.text.height` **[number][14]** integral number of pixels high. When defined, `dpi` will be ignored and the text will automatically fit the pixel resolution defined by `width` and `height`. Will be ignored if `width` is not specified or set to 0. (optional, default `0`)
+        *   `options.text.align` **[string][12]** text alignment (`'left'`, `'centre'`, `'center'`, `'right'`). (optional, default `'left'`)
+        *   `options.text.justify` **[boolean][15]** set this to true to apply justification to the text. (optional, default `false`)
+        *   `options.text.dpi` **[number][14]** the resolution (size) at which to render the text. Does not take effect if `height` is specified. (optional, default `72`)
+        *   `options.text.rgba` **[boolean][15]** set this to true to enable RGBA output. This is useful for colour emoji rendering, or support for pango markup features like `<span foreground="red">Red!</span>`. (optional, default `false`)
+        *   `options.text.spacing` **[number][14]** text line height in points. Will use the font line height if none is specified. (optional, default `0`)
 
 ### Examples
 
--- a/docs/api-constructor.md
+++ b/docs/api-constructor.md
@@ -139,6 +139,29 @@ await sharp({
 }).toFile('noise.png');
 ```
 
+```javascript
+// Generate an image from text
+await sharp({
+  text: {
+    text: 'Hello, world!',
+    width: 400, // max width
+    height: 300 // max height
+  }
+}).toFile('text_bw.png');
+```
+
+```javascript
+// Generate a color image from text using pango markup and font
+await sharp({
+  text: {
+    text: '<span foreground="red">Red!</span><span background="cyan">blue</span>',
+    font: 'sans'
+    rgba: true,
+    dpi: 300
+  }
+}).toFile('text_color.png');
+```
+
 *   Throws **[Error][17]** Invalid parameters
 
 Returns **[Sharp][18]** 
--- a/lib/constructor.js
+++ b/lib/constructor.js
@@ -126,6 +126,17 @@ const debuglog = util.debuglog('sharp');
  * @param {string} [options.create.noise.type] - type of generated noise, currently only `gaussian` is supported.
  * @param {number} [options.create.noise.mean] - mean of pixels in generated noise.
  * @param {number} [options.create.noise.sigma] - standard deviation of pixels in generated noise.
+ * @param {Object} [options.text] - describes a new text image to be created.
+ * @param {string} [options.text.text] - text to render as a UTF-8 string. It can contain Pango markup, for example `<i>Le</i>Monde`.
+ * @param {string} [options.text.font] - font name to render with.
+ * @param {string} [options.text.fontfile] - absolute filesystem path to a font file that can be used by `font`.
+ * @param {number} [options.text.width=0] - integral number of pixels to word-wrap at. Lines of text wider than this will be broken at word boundaries.
+ * @param {number} [options.text.height=0] - integral number of pixels high. When defined, `dpi` will be ignored and the text will automatically fit the pixel resolution defined by `width` and `height`. Will be ignored if `width` is not specified or set to 0.
+ * @param {string} [options.text.align='left'] - text alignment (`'left'`, `'centre'`, `'center'`, `'right'`).
+ * @param {boolean} [options.text.justify=false] - set this to true to apply justification to the text.
+ * @param {number} [options.text.dpi=72] - the resolution (size) at which to render the text. Does not take effect if `height` is specified.
+ * @param {boolean} [options.text.rgba=false] - set this to true to enable RGBA output. This is useful for colour emoji rendering, or support for pango markup features like `<span foreground="red">Red!</span>`.
+ * @param {number} [options.text.spacing=0] - text line height in points. Will use the font line height if none is specified.
  * @returns {Sharp}
  * @throws {Error} Invalid parameters
  */
--- a/lib/constructor.js
+++ b/lib/constructor.js
@@ -92,6 +92,27 @@ const debuglog = util.debuglog('sharp');
  *  }
  * }).toFile('noise.png');
  *
+ * @example
+ * // Generate an image from text
+ * await sharp({
+ *   text: {
+ *     text: 'Hello, world!',
+ *     width: 400, // max width
+ *     height: 300 // max height
+ *   }
+ * }).toFile('text_bw.png');
+ *
+ * @example
+ * // Generate a color image from text using pango markup and font
+ * await sharp({
+ *   text: {
+ *     text: '<span foreground="red">Red!</span><span background="cyan">blue</span>',
+ *     font: 'sans'
+ *     rgba: true,
+ *     dpi: 300
+ *   }
+ * }).toFile('text_color.png');
+ *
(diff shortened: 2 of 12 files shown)


brahima:
Hi again @lovell,

I installed a fresh ubuntu image via docker with libvips dependencies, node 16.15, valgrind ... and now have only one error remaining on the existing tests (even on main branch). The error occurs in test/util.js (Vendor -> Contains expected attributes) : 

```  
1) Utilities
       Vendor
         Contains expected attributes:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:

true !== false

      + expected - actual

      -true
      +false
      
      at Context.<anonymous> (test/unit/util.js:143:14)
      at processImmediate (node:internal/timers:466:21)
```

The _sharp.vendor.installed_ variable in my case contains this (the installed array is empty) :
`{ current: 'linux-x64', installed: [] }`

Any clue about that ?

Also i have some of my own tests failing right now ... i'm having a look.

Thanks for your help

brahima:
Hello @lovell,

This is me again. I now understand why i have failures on my own (new) tests. Generated image based on text can be quite different depending on the OS and font combination. That's why i introduced a specific font file for the tests ... but that's not that easy :-)

I did some tests using different vips command to see the difference between MacOs / Linux (Windows is not even in the equation). Here's only two of them :

`vips text /tmp/text_macos.png "<span>This is a text to justify</span>" --font 'Arial 100px' --align centre --width 300`
image

`vips text /tmp/text_macos.png "<span>This is a text to justify</span>" --font 'Bitstream Vera Sans Bold Oblique 100px' --align centre --width 300`
image
That makes me think that the Bitstream Vera font is not a good fit for my unit tests.

Identifying a font that can give the « same » results across the different Oses and keeping the fixtures.assertSimilar() in the unit tests may not be a viable option. 

Maybe @jcupitt can give some insights about this ?

Thanks in advance for your help regarding this issue.

PS : I still have the vendor unit test failing locally

Best regards,
Brahim

jcupitt:
@brahima Text rendering is really complex, and tiny changes to your system can completely alter the bitmap you get. You can't expect to get the same result from ubuntu 21.04 and 21.10, for example, never mind between macos and linux. You can't even expect two users both on ubuntu 22.04 to get the same result.

I would just check that the generated text image has a not-crazy width and height, has one band, and that the image average is between 0 and 255.

Here's how libvips tests text:

https://github.com/libvips/libvips/blob/master/test/test-suite/test_create.py#L403-L418

brahima:
Hi @jcupitt and thanks for your feedback, makes sense.
Just after sending my last message, i had the idea to check the unit test of libvips.

Also, @lovell, after playing around with my version and testing, i think that having default values for width (0), height (0) and dpi (72) is not a good idea and makes the default behavior strange and different from vips (unless making some adjustments)

For exemple, for the following commands (vips default behavior vs. sharp) :

`vips text /tmp/text_vips.png 'Hello, world!' --width 500 --height 500 `
image
Here, since we set the dpi to 72 by default ... it is not computed by libvips based on width and height.

I see 2 options : 
1. remove these default values and let libvips « do its work » and as you mentioned [here](https://github.com/lovell/sharp/issues/512#issuecomment-817366966), let user provides only dpi or height.
2. add some rules (ex : if width and height are set => do not set default value for dpi ...).

I do think that the 1st option is better but would like to have your point of view on this.

Also, [here after getting the image from libvips](https://github.com/lovell/sharp/pull/3252/commits/90ca1d21bc8b5b4e0eab6b7a2bf33fb5cbf4e076#diff-53a01d31aa585021f8f316c6613dadef86bf1f9ac2070d52daa68adbefbb6ef4R450), even if i set the image type manually to VIPS_INTERPRETATION_B_W, i always get a 3 channel image in sharp when testing (despite the image.bands() being equals to 1). I expected to have only one channel image with its space being 'b-w' and not 'srgb'.

Sorry for all my questions and hope i made my self clear :-)

Regards,
Brahim

brahima:
Hello,

I just re-pushed with updated tests :
- i removed the font 
- kept the default values with some more checks (height and dpi input are mutually exclusive)

I'll wait for your review before doing any further changes.

Regards,
Brahim

lovell:
@brahima Thank you very much for the PR, I plan to review this next week.

brahima:
Hello @lovell,

Thanks in advance for your time. I stay tuned :-)

kantorcodes:
> @brahima Thank you very much for the PR, I plan to review this next week.

Any updates on this one? Would love to utilize this feature.

lovell (changes requested):
Thank you very much for taking the time to work on this feature, it's looking really good, only a couple of small inline comments.

The tests appear to be failing on Linux due to a lack of fonts in the CI environment. Whilst adding fonts should be straightforward, it highlights that we need to do something (code? docs?) to catch this problem and aid Linux users. I imagine it'll be common in FaaS environments such as AWS Lambda that lack control over the OS, or minimal containers that don't install fonts by default.

I'm going to switch the base branch to the `eagle` branch for inclusion in sharp v0.31.0, so you may need to rebase as there have been a couple of changes relating to the forthcoming libvips v8.13.0 upgrade.

Thanks again, this is an oft-requested feature and will be greatly appreciated by many.

src/common.cc
```diff
@@ -394,6 +430,31 @@ namespace sharp {
...
+        } else if (descriptor->textDpi > 0) {
+          textOptions->set("dpi", descriptor->textDpi);
+        }
+        if (descriptor->textFont.length() > 0) {
+          textOptions->set("font", const_cast<char*>(descriptor->textFont.data()));
+        }
+        if (descriptor->textFontfile.length() > 0) {
+          textOptions->set("fontfile", const_cast<char*>(descriptor->textFontfile.data()));
+        }
+        image = VImage::new_memory().text(const_cast<char *>(descriptor->textValue.data()), textOptions);
```
kleisauke, line 455: I think you can just call `VImage::text` here.
```suggestion
        image = VImage::text(const_cast<char *>(descriptor->textValue.data()), textOptions);
```

src/common.cc
```diff
@@ -394,6 +430,31 @@ namespace sharp {
...
+        image.get_image()->Type = image.guess_interpretation();
```
kleisauke, line 456: This is probably redundant when RGBA output is enabled. libvips will only tag the image as `VIPS_INTERPRETATION_MULTIBAND` when the `@rgba` option is not set.
```suggestion
        if (!descriptor->textRgba) {
          image = image.copy(VImage::option()->set("interpretation", VIPS_INTERPRETATION_B_W));
        }
```

lib/constructor.js
```diff
@@ -126,6 +126,17 @@ const debuglog = util.debuglog('sharp');
  * @param {string} [options.create.noise.type] - type of generated noise, currently only `gaussian` is supported.
  * @param {number} [options.create.noise.mean] - mean of pixels in generated noise.
  * @param {number} [options.create.noise.sigma] - standard deviation of pixels in generated noise.
+ * @param {Object} [options.text] - describes a new text image to be created.
```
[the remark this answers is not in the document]
brahima: Ok added 2 examples. Let me know if this is enough

src/common.cc
```diff
@@ -394,6 +430,31 @@ namespace sharp {
...
+        imageType = ImageType::RAW;
```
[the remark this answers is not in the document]
brahima: Are you talking about the info object exposed via output options (toFile / toBuffer ...) ?

Also, libvips seems to set the autofit_dpi calculated on the [options object](https://github.com/libvips/libvips/blob/c7baa4535a72bf3be057df184e9813161b44019d/libvips/create/text.c#L330)
However, the VOption class have only setters and no getters (with a private Pair structure). 
Any hints on how to access the computed option value ?

Thanks

src/common.cc
brahima: Done

src/common.cc
brahima: Done

brahima:
Hello @lovell and @kleisauke and thanks for the time you took reviewing my PR.

> Thank you very much for taking the time to work on this feature, it's looking really good, only a couple of small inline comments.

You're welcome :-)

> The tests appear to be failing on Linux due to a lack of fonts in the CI environment. Whilst adding fonts should be straightforward, it highlights that we need to do something (code? docs?) to catch this problem and aid Linux users. I imagine it'll be common in FaaS environments such as AWS Lambda that lack control over the OS, or minimal containers that don't install fonts by default.

About adding the fonts for [the CI env](https://github.com/lovell/sharp/blob/eagle/.github/workflows/ci.yml), do you have specific packages in mind ? 

Regarding font problem detection : 
- what should be the better place for the doc ? constructor page or installation page 
- for the code part, aren't the fontconfig/pango errors sufficient enough ? Do you want sharp to wrap them in its own specific error messages ?

```
Fontconfig error: Cannot load default config file: No such file: (null)
(process:626): Pango-CRITICAL **: 17:59:42.755: pango_font_describe: assertion 'font != NULL' failed
```

> I'm going to switch the base branch to the `eagle` branch for inclusion in sharp v0.31.0, so you may need to rebase as there have been a couple of changes relating to the forthcoming libvips v8.13.0 upgrade.

Ok, done :-)
 
> Thanks again, this is an oft-requested feature and will be greatly appreciated by many.

src/common.cc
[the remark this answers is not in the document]
jcupitt: You can set a reference to the variable to assign the value to, eg.

```C++
    ->set("autofit_dpi", &output_dpi)
```

And it'll assign the value on return.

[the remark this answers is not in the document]
brahima: Thanks @jcupitt, ok makes sense. Way easier than what i was thinking about :-)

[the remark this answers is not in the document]
brahima: @lovell Just pushed a commit with the autofitDdpi parameter added to the info output object.
Still need to add some lines to the documentation but waiting to know if the value is exposed at the right place.
Thanks

lovell:
Thanks for the updates, it looks like there are a couple of linting errors.

For the Linux-based CI environments, we can probably use the Open Sans typeface as that's widely available (add `open-sans-fonts` package on Centos, `font-opensans` on Alpine etc.)

I'm happy to write the documentation for the use of fonts after we merge this PR. There have been a few questions in the past about SVG font rendering too, so it would be good to get it all down in one place.

lovell:
Commit https://github.com/lovell/sharp/commit/254944f8ab93d2bcc610e41f554754ed5ae3e6a1 on the `eagle` branch ensures the Noto Sans font is available in all the CI environments, which should allow the generic "sans" font name to be used in tests.

brahima:
> Thanks for the updates, it looks like there are a couple of linting errors.
My bad ... should be fixed now.

brahima:
@lovell Hi,
Any feedback regarding the  autofitDpi exposed via the info object ?

src/common.h
```diff
@@ -71,6 +71,17 @@ namespace sharp {
...
+    std::string textFont;
+    std::string textFontfile;
+    int textWidth;
+    int textHeight;
+    VipsAlign textAlign;
+    bool textJustify;
+    int textDpi;
+    bool textRgba;
+    int textSpacing;
+    int autofitDpi;
```
lovell, line 84: I know I originally suggested `autofitDpi` as a name so this is my fault, but perhaps we should rename this as `textAutofitDpi` throughout to keep in line with existing feature-related output properties such as `trimOffsetTop`.

docs/api-constructor.md
```diff
@@ -127,6 +139,29 @@ await sharp({
...
+  text: {
+    text: 'Hello, world!',
+    width: 400, // max width
+    height: 300 // max height
+  }
+}).toFile('text_bw.png');
+```
+
+```javascript
+// Generate a color image from text using pango markup and font
```
lovell, line 154: Thank you for adding examples, these are great. May I suggest we replace `color` with `RGBA` in the comment and filename to avoid the English spelling conundrum.

lovell (changes requested):
Thank you very much for all the updates, a couple of very minor inline comments and we're good to go.

lovell:
I've added some documentation about font discovery via commit https://github.com/lovell/sharp/commit/3e327a586cda725a05f0f09c1e4f423c96d5744f

docs/api-constructor.md
brahima: Done.
I'm not a native english speaker so learned a new word (conundrum) :-)

brahima:
@lovell also added 2 lines to api-output documentation regarding autofit dpi.

Hope we're good to go but do not hesitate if you have any feedback.

lovell (approved):

lovell:
Thank you very much for working on and persisting with this PR, it will be included in v0.31.0.

Issue #3259, opened 2022-06-13
Setting WebP effort to 0 in constructor does not change the object's effort option to 0


AlexanderTheGrey:
### Is this a possible bug in a feature of sharp, unrelated to installation?

- [ X] Running `npm install sharp` completes without error.
- [ X] Running `node -e "require('sharp')"` completes without error.

If you cannot confirm both of these, please open an [installation issue](https://github.com/lovell/sharp/issues/new?labels=installation&template=installation.md) instead.

### Are you using the latest version of sharp?

- [ X] I am using the latest version of `sharp` as reported by `npm view sharp dist-tags.latest`.

If you cannot confirm this, please upgrade to the latest version and try again before opening an issue.

If you are using another package which depends on a version of `sharp` that is not the latest, please open an issue against that package instead.

### What is the output of running `npx envinfo --binaries --system --npmPackages=sharp --npmGlobalPackages=sharp`?

System:
    OS: macOS 10.15.7
    CPU: (6) x64 Intel(R) Core(TM) i5-8500B CPU @ 3.00GHz
    Memory: 16.41 MB / 16.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 16.14.2 - ~/.nvm/versions/node/v16.14.2/bin/node
    npm: 8.5.0 - ~/.nvm/versions/node/v16.14.2/bin/npm
  npmPackages:
    sharp: ^0.30.6 => 0.30.6

### What are the steps to reproduce?

Setting WebP effort to 0 (fastest) in the constructor does not change the effort option to 0.

However, setting WebP effort 1 in the constructor correctly changes the effort option to 1.

https://sharp.pixelplumbing.com/api-output#webp

### What is the expected behaviour?

The effort option should be set to 0.

### Please provide a minimal, standalone code sample, without other dependencies, that demonstrates this problem

<pre>
const image = sharp(input.gif, { animated: true }).webp({ quality: 100, effort: 0 })

image.toFile(output.webp, (err) => {
    console.log("WebP Effort ====>", image.options.webpEffort);
    if (err === null) {
      console.log("Success");
    } else {
      console.log("Error")
    }
  });
</pre>

### Please provide sample image(s) that help explain this problem

https://en.wikipedia.org/wiki/GIF#/media/File:Rotating_earth_(large).gif

lovell:
Good spot, the problem is here:

[...]

PR #3261, opened 2022-06-14
Fix setting WebP effort to 0 in constructor #3259

AlexanderTheGrey:
Setting WebP effort to 0 in the constructor caused the effort to use the
default value of 4.

### DIFF
--- a/lib/output.js
+++ b/lib/output.js
@@ -495,7 +495,7 @@ function webp (options) {
     if (is.defined(options.smartSubsample)) {
       this._setBooleanOption('webpSmartSubsample', options.smartSubsample);
     }
-    const effort = options.effort || options.reductionEffort;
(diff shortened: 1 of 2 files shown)


lovell:
Would something like the following work to ensure we still support the deprecated `reductionEffort` property?
```js
const effort = is.defined(options.effort) ? options.effort : options.reductionEffort;
```
Please can you remove the unrelated layout changes too.

AlexanderTheGrey:
> Would something like the following work to ensure we still support the deprecated reductionEffort property?

Ah, yes, I modified the code as if it were removed.

> Please can you remove the unrelated layout changes too.

Yeah, my VS Code automatically formatted it for some reason.

lovell:
Brilliant, thank you!

PR #3267, opened 2022-06-17
Use vips to upsample webp 

### LINKED ISSUE #3262: Resizing then extracting Webp images past 16,383 now fails from sharp 0.30.x

- [ x ] Running `npm install sharp` completes without error.
- [ x ] Running `node -e "require('sharp')"` completes without error.

- [ x ] I am using the latest version of `sharp` as reported by `npm view sharp dist-tags.latest`.

Example  https://github.com/blacha/sharp-resize

In sharp 0.29.x I can resize a webp image to basically any size and then extract a region from the enlarged image without any issue. In sharp 0.30.x this now fails with `Error: webp: bad image dimensions`

```typescript
await sharp(buffer).resize(32768, 32768)
  .extract({ top: 16384, left: 16384, width: 1024, height: 1024 })
  .webp().toBuffer();
```

I see some work was done in libvips recently to improve image resizing especially with SVGs, is there a easy way to go back to the old resizing method?

### Please provide a minimal, standalone code sample, without other dependencies, that demonstrates this problem

Example  https://github.com/blacha/sharp-resize

### Please provide sample image(s) that help explain this problem

Sample image in repo.

blacha:
When scaling WebP images up past 16383 x 16383 use vips scaling logic to prevent `Error: webp: bad image dimensions`

Fixes #3262

### DIFF
--- a/src/pipeline.cc
+++ b/src/pipeline.cc
@@ -188,8 +188,10 @@ class PipelineWorker : public Napi::AsyncWorker {
           if (jpegShrinkOnLoad > 1 && static_cast<int>(shrink) == jpegShrinkOnLoad) {
             jpegShrinkOnLoad /= 2;
           }
-        } else if (inputImageType == sharp::ImageType::WEBP ||
-                   inputImageType == sharp::ImageType::SVG ||
+        } else if (inputImageType == sharp::ImageType::WEBP && shrink > 1.0) {
+          // Avoid upscaling via webp
+          scale = 1.0 / shrink;
+        } else if (inputImageType == sharp::ImageType::SVG ||
                    inputImageType == sharp::ImageType::PDF) {
           scale = 1.0 / shrink;
         }
@@ -217,7 +219,7 @@ class PipelineWorker : public Napi::AsyncWorker {
           ->set("access", baton->input->access)
           ->set("scale", scale)
           ->set("fail_on", baton->input->failOn);
-        if (scale < 1.0 && inputImageType == sharp::ImageType::WEBP) {
+        if (inputImageType == sharp::ImageType::WEBP) {
           option->set("n", baton->input->pages);
           option->set("page", baton->input->page);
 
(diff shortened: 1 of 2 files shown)


src/pipeline.cc
```diff
@@ -217,7 +217,7 @@ class PipelineWorker : public Napi::AsyncWorker {
           ->set("access", baton->input->access)
           ->set("scale", scale)
           ->set("fail_on", baton->input->failOn);
-        if (inputImageType == sharp::ImageType::WEBP) {
+        if (scale < 1.0 && inputImageType == sharp::ImageType::WEBP) {
```
kleisauke, line 220: This causes a small memory leak for the dynamic-allocated `*option` when `scale > 1.0`, since it's never `delete`'d. How about doing this?
```diff
--- a/src/pipeline.cc
+++ b/src/pipeline.cc
@@ -188,8 +188,10 @@ class PipelineWorker : public Napi::AsyncWorker {
           if (jpegShrinkOnLoad > 1 && static_cast<int>(shrink) == jpegShrinkOnLoad) {
             jpegShrinkOnLoad /= 2;
           }
-        } else if (inputImageType == sharp::ImageType::WEBP ||
-                   inputImageType == sharp::ImageType::SVG ||
+        } else if (inputImageType == sharp::ImageType::WEBP && shrink > 1.0) {
+          // Avoid upscaling via libwebp
+          scale = 1.0 / shrink;
+        } else if (inputImageType == sharp::ImageType::SVG ||
                    inputImageType == sharp::ImageType::PDF) {
           scale = 1.0 / shrink;
         }
```

lovell:
Brilliant, thank you for fixing this (and thanks Kleis for reviewing).

blacha:
Awesome than you both! Libvips/sharp are amazing tools! 

When do you think a release for this will happen?

lovell:
@blacha This will be part of v0.30.7, probably some time later this week ahead of the work to test the forthcoming libvips v8.13.0.

PR #3303, opened 2022-07-23
Vectorized linear function

antonmarsden:
@lovell This looked easy enough to implement, and adds some good value. This pull request is not really production quality and my knowledge of the code base is very limited. A few things to consider:
- do the additional variables make sense (linearVecA/linearVecB) or could linearA/linearB be both number & number[]?
- are there other functions that can be generalised in a similar fashion?
- probably needs a good test case :)
- unclear what the behaviour is (or should be) if the arrays are different lengths or if only one array is defined


lovell:
Hi, thanks for the PR.

How about we switch the (internal) `this.options` so `linearA` and `linearB` are always arrays, and if someone provides a single value (the current behaviour) we convert to a single element array.
```js
.linear(a, b)
// would be equivalent to
.linear([a], [b])
// but now we would also support
.linear([a1, a2, a3], [b1, b2, b3])
```
This should make the C++ code path a bit simpler as it will only deal with arrays of doubles.

antonmarsden:
@lovell Great suggestion, I have implemented it. Added some fancy logic for dealing with the alpha channel that you may wish to review. Only outstanding thing is test cases, which I can hopefully make some progress on shortly.

antonmarsden:
Handling the multi-band expansion for linear() looked tricky in the context of sharp's colourspace handling.

src/pipeline.cc
```diff
@@ -688,8 +688,14 @@ class PipelineWorker : public Napi::AsyncWorker {
         image = sharp::Gamma(image, baton->gammaOut);
       }
 
+      // Linear arg check
```
lovell, line 691: Can this array length check move to occur earlier, in the JavaScript rather than C++? We should probably add a test for it too.

lib/constructor.js
```diff
@@ -283,8 +283,8 @@ const Sharp = function (input, options) {
     tileCentre: false,
     tileId: 'https://example.com/iiif',
     timeoutSeconds: 0,
-    linearA: 1,
-    linearB: 0,
+    linearA: [1.0],
```
lovell, line 286: We could default these to an empty array, which should simplify the "should we perform a linear operation?" logic in the C++.

lib/operation.js
```diff
@@ -205,7 +205,7 @@ function affine (matrix, options) {
  * const data = await sharp(input)
  *   .sharpen({
  *     sigma: 2,
- *     m1: 0
+ *     m1: 0,
```
lovell, line 208: Good spot, thank you.

lovell (changes requested):
Thanks for the updates. I've left a couple of small comments/questions inline but otherwise this is good to go.

Please can you rebase against the `upstream/eagle` branch.

antonmarsden:
@lovell Done... all yours.

antonmarsden:
@lovell Done, all yours.

lib/operation.js
```diff
@@ -629,25 +629,37 @@ function boolean (operand, operator, options) {
...
  * Apply the linear formula a * input + b to the image (levels adjustment)
- * @param {number} [a=1.0] multiplier
- * @param {number} [b=0.0] offset
+ * @param {number | Array<number>} [a=[]] multiplier
+ * @param {number | Array<number>} [b=[]] offset
  * @returns {Sharp}
  * @throws {Error} Invalid parameters
  */
 function linear (a, b) {
+  if (!is.defined(a) && is.number(b)) {
```
lovell, line 638: Would it make things easier/simpler if we change the API to enforce that both `a` and `b` are required, and are no longer optional? Calling `linear` without parameters is effectively a no-op anyway. (The type would still be `number | Array<number>`.)

antonmarsden: Up to you. It could be a breaking change for some dependent apps.

lovell: I'm OK with breaking changes that simplify logic, but if it makes minimal difference to the complexity then let's keep the API as it is.

lovell:
Please can you add unit test(s) to ensure all the new logic is fully-covered - see https://coveralls.io/builds/51160601

lovell:
Landed via commit https://github.com/lovell/sharp/commit/74e3f73934df4fa8096fe87ce1b57b9cf0d50a5f - this will be in v0.31.0, thanks again for working on this.

PR #3323, opened 2022-08-06
Docs: fix broken link to Got stream documentation

kidonng:


lovell (changes requested):
Thank you for the PR. Please can you update the JSDocs in the source then use `npm run docs-build` to update the markdown, otherwise this file will be overwritten next time there's a change to the source.

https://github.com/lovell/sharp/blob/e40a881ab4a5e7b0e37ba17e31b3b186aef8cbf6/lib/constructor.js#L349

lovell:
Thank you!

PR #3326, opened 2022-08-08
package.json: add node-gyp as an optional peer dep

wmertens:
This way, the implicit dependency on node-gyp for local builds is explicit, and build tools like dream2nix don't need workarounds to add it :-)

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,6 +152,14 @@
     "rimraf": "^3.0.2",
     "semistandard": "^16.0.1"
   },
+  "peerDependencies": { 
+    "node-gyp": "*"
+  },
+  "peerDependenciesMeta": {
+    "node-gyp": {
+      "optional": true
+    }
+  },
   "license": "Apache-2.0",
   "config": {
     "libvips": "8.12.2",


package.json
```diff
@@ -152,6 +152,14 @@
     "rimraf": "^3.0.2",
     "semistandard": "^16.0.1"
   },
+  "peerDependencies": { 
+    "node-gyp": "*"
```
lovell, line 156: Could the use of _any_ version cause problems? Perhaps we should limit to the major version number of the minimum `node-gyp` version provided by the bundled version of `npm` for the earliest Node.js we support (npm 6)? If so, this would be `node-gyp` version 5.

package.json
```diff
@@ -152,6 +152,14 @@
...
+  },
+  "peerDependenciesMeta": {
+    "node-gyp": {
+      "optional": true
```
lovell, line 160: I recall some versions of npm auto-installing these erroneously - are you aware of any remaining problems this might cause?

lovell (commented):
Thanks for the PR, I've left a couple of questions inline.

package.json
wmertens: Ah - the optional peer dependencies need npm v7 :(

package.json
wmertens: Worst case, you'll get a message that the peer dependency is missing?

package.json
wmertens: Which means npm 6 will just complain if node-gyp isn't installed.

lovell: Thanks, I'd prefer to avoid dealing with support issues about perceived missing dependencies, so it sounds like we should we wait until Node.js 14, and therefore npm v6, reaches EOL.

wmertens:
You know what - I now see that node-gyp ships with npm by default (thanks for pointing that out) and it's just a matter of making sure it's in the path while building outside of npm. I was calling node-gyp directly and it was missing. Thanks again and sorry for the disturbance :+1:

PR #3332, opened 2022-08-10
Add the option to provide a specific background colour to trim

mart-jansink:
I've got a use case to trim just a specific background colour of an image. Mostly to detect letter and pillar boxes in images by trimming off any black areas. This pull request adds the option to do so, by providing this colour as a string or object, which will be parsed by the `color` module.

For me, it'd be more typical to provide a specific colour and use the default threshold, so I made the former the first optional argument. However, the Javascript side will handle the case where a `threshold` is provided as the first argument to retain backwards compatibility. I'm not sure if I made it fully clear in the docs that both arguments are optional, maybe you can improve my phrasing if you agree?

The output of the `color` module is always 8-bit, so it's scaled accordingly if the image is 16-bit. I've noticed that the `threshold` option is used as an absolute value in [find_trim](https://github.com/libvips/libvips/blob/0f56b099d0b2e001dc5a6f686cb09d2b0e44fc34/libvips/arithmetic/find_trim.c#L127) instead of being scaled relatively to the interpretation, therefore that's now multiplied as well.

Instead of `.flatten()`ing the background color if the image `HasAlpha()`, the alpha value is now simply discarded. I believe that this should give the same result because of how [find_trim](https://github.com/libvips/libvips/blob/0f56b099d0b2e001dc5a6f686cb09d2b0e44fc34/libvips/arithmetic/find_trim.c#L105) uses the same  background color when it flattens the image. At least all existing unit tests still pass.

Finally, I've based this pull request on the `eagle` branch because e0d3c6e05dc6e4caacb4b3b2160178fb7e6f5ca2 and cbf741cac7cc78ef20d5962131d5c6b83630046d affect the same part of the code. And I also encountered some failing tests when running the `main` branche on my M1 MacBook Pro.


test/unit/trim.js
```diff
@@ -168,4 +168,54 @@ describe('Trim borders', function () {
       });
     });
   });
+
+  describe('Specific background colour', function () {
```
lovell, line 172: Always a pleasure to see tests, thank you.

lib/resize.js
```diff
@@ -438,11 +438,17 @@ function extract (options) {
...
  * will contain `trimOffsetLeft` and `trimOffsetTop` properties.
  *
- * @param {number} [threshold=10] the allowed difference from the top-left pixel, a number greater than zero.
+ * @param {String|Object} [background='top-left pixel'] background colour, parsed by the [color](https://www.npmjs.org/package/color) module, defaults to that of the top-left pixel.
+ * @param {number} [threshold=10] the allowed difference from the above colour, a number greater than zero.
  * @returns {Sharp}
  * @throws {Error} Invalid parameters
  */
-function trim (threshold) {
+function trim (background, threshold) {
```
lovell, line 446: Perhaps we should make the signature accept an Object of options, so `background` would be specified via:
```js
sharp().trim({ background: 'red' })...
```
so if `threshold` is also required, this would be via:
```js
sharp().trim({ background: 'red', threshold: 20 })...
```
whilst ensuring the existing, number-is-threshold approach still works:
```js
sharp().trim(20)...
```

lib/resize.js
```diff
@@ -429,7 +429,7 @@ function extract (options) {
 }
 
 /**
- * Trim "boring" pixels from all edges that contain values similar to the top-left pixel.
+ * Trim pixels from all edges that contain values similar to the given background colour, which defaults to that of the top-left pixel.
```
lovell, line 432: Thanks for updating the docs, please can you add an `@example` of providing a colour.

lovell (changes requested):
Thank you very much for the PR, this is a welcome addition, I've left a couple of questions/comments inline.

src/pipeline.cc
```diff
@@ -118,7 +118,20 @@ class PipelineWorker : public Napi::AsyncWorker {
       // Trim
       if (baton->trimThreshold > 0.0) {
         MultiPageUnsupported(nPages, "Trim");
-        image = sharp::Trim(image, baton->trimThreshold);
+        // Background colour
```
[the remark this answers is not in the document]
mart-jansink: That'd be fine with me, I just copied the logic from https://github.com/lovell/sharp/blob/94fc1a30a9438cf479c82ea1317be72339b11a7b/src/pipeline.cc#L337-L344 where it's also done in the pipeline. Of course, `flatten` isn't abstracted to a separate operation, so there's something to say for either way. Please let me know what you'd prefer.

lib/resize.js
mart-jansink: Ah, that looks like a clear solution, thank you!

lib/resize.js
mart-jansink: Will do

src/pipeline.cc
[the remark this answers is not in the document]
lovell: The flatten logic should probably move into its own operation, I can fix that later, thanks for highlighting this.

mart-jansink:
Could you please have another look? I've modified the pull request according to your suggestions.

Also, it now allows the `threshold` to be zero to trim just the pixels with the exact same color. I couldn't think of any reason why this shouldn't work. Of course I'll revert it if there was one, but https://github.com/libvips/libvips/blob/f4a3c7eb5478ee77bd206353308eab3b02db923a/libvips/arithmetic/find_trim.c#L193 seems to be fine with anything between (and including) `0` and `INFINITY `.

lovell (approved):

lovell:
Dankjewel Mart!

PR #3349, opened 2022-09-05
fix: limitInputPixels for resolutions higher than uint32 limit

marcosc90:
This PR fixes an issue with `limitInputPixels` not working for images with resolutions exceeding the uint32 limit.

- [65535×65535 pixels PNG](https://alexandre.alapetite.fr/doc-alex/large-image/65535.png) works correctly, error is triggered
- [65536×65536 pixels PNG](https://alexandre.alapetite.fr/doc-alex/large-image/65536.png) error is not triggered and the image is processed (takes a lot of time)

The added test fails on `main` branch.


src/common.cc
```diff
@@ -507,9 +507,10 @@ namespace sharp {
         }
       }
     }
+
     // Limit input images to a given number of pixels, where pixels = width * height
     if (descriptor->limitInputPixels > 0 &&
-      static_cast<uint64_t>(image.width() * image.height()) > descriptor->limitInputPixels) {
+      static_cast<uint64_t>(image.width()) * static_cast<uint64_t>(image.height()) > descriptor->limitInputPixels) {
```
kleisauke, line 513: You could also cast only `image.width()` to `uint64_t`, `image.height()` is then implicitly converted to `uint64_t`. Then, the multiplication is still done with two operands of type `uint64_t`.
```suggestion
      static_cast<uint64_t>(image.width()) * image.height() > descriptor->limitInputPixels) {
```
(helps to avoid the 120 column width, which this line almost exceeds)

marcosc90: Good to know :), Haven't written a line of C++ in 11 years. I'll push it in a bit.

lovell:
Thank you!

PR #3350, opened 2022-09-05
doc(resize): images examples of resize types

Aarbel:
Sharp is missing a lot of visual examples in its doc.

image

### DIFF
--- a/docs/api-resize.md
+++ b/docs/api-resize.md
@@ -12,6 +12,8 @@ When both a `width` and `height` are provided, the possible methods by which the
 *   `inside`: Preserving aspect ratio, resize the image to be as large as possible while ensuring its dimensions are less than or equal to both those specified.
 *   `outside`: Preserving aspect ratio, resize the image to be as small as possible while ensuring its dimensions are greater than or equal to both those specified.
 
+![image](https://user-images.githubusercontent.com/25119847/188477841-22a9651c-993c-4735-a7bf-70c2828d3150.png)
+
 Some of these values are based on the [object-fit][1] CSS property.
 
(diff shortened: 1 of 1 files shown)


lovell:
Hi, thank you for this, it will be very useful.

Please can you add this image to the repo itself as part of this PR in [docs/image](https://github.com/lovell/sharp/tree/main/docs/image).

Please can you add a (markdown-formatted) reference to this image using an absolute URL `https://cdn.jsdelivr.net/gh/lovell/sharp@main/docs/image/resize-fit.png` as part of the following JSDocs then use `npm run docs-build` to regenerate `docs/api-resize.md`.

https://github.com/lovell/sharp/blob/eee0dd36d92bf6304da5cd0b6298eb2ae7db2349/lib/resize.js#L102-L112

The "image" description / alt text should be descriptive too, perhaps "result of applying fit options to a landscape image".

lovell:
I've added a slightly cleaned-up version of this image to the repo via commit https://github.com/lovell/sharp/commit/efbb0c22fdb82d40f71253b097ce3a4459f76fa5 with some HTML/CSS to avoid a layout shift.

Thank you very much for taking the time to submit a PR for this originally.

kleisauke:
Perhaps it would be neater to recreate that image in SVG? See for example (I could open a PR for this):
https://t0.nl/api-resize-fit.svg

Note that this placeholder image was the default for broken images in Windows 10. Microsoft had redesigned its icons as part of Windows 11 (and preview builds of Windows 10).
https://blogs.windows.com/windows-insider/2021/03/24/announcing-windows-10-insider-preview-build-21343/

kleisauke:
@lovell I just opened PR #3735 for this, which also uses a different base image to avoid any copyright implications.

Issue #3369, opened 2022-09-15
Bundled, official TypeScript definitions?


rexxars:
Hi Lovell!

I've been helping maintain the `@types/sharp` module lately, and was wondering if you had any thoughts on whether or not you'd like to see these integrated into (bundled with) the sharp module?

There are a couple of annoyances with having a separate type module:

- You have to install it separately (duh)
- Someone (usually not the author of the feature in sharp) has to submit a PR to the types repo, hopefully understanding the changes that have been done in sharp well enough to type them
- The types often trail the sharp releases, so using the latest version from typescript can be a bit of a pain (having to `(sharp as any)` things)
- People tend to add the stuff to the typings that they need to use, but not necessarily all of the features that have been added
- Sharp is still on the 0.x semver range, which makes it hard for the types module to be versioned properly. If there is a correction to the types, we have to bump the patch version - which often leads to being ahead of sharp in terms of the version number. If sharp follows semver and is on the 1.x or above range, we could provide corrections in patches that theoretically is "ahead" of sharp, but at least follows semver. Example:
  1. Someone sends a PR to fix a typing issue in 0.32.0
  2. The PR is merged and released as `@types/sharp@0.32.1` (sharp is still at `0.32.0`, but the types are at `0.32.1`)
  3. Sharp adds a new feature and releases 0.32.1
  4. A new PR is merged to add typing support for the feature, and is released as `0.32.2`

The situation isn't terrible, but I was wondering if you had any opinion on the matter. I'm sure everyone is happy to continue contributing to the types module, but I'm also sure people would appreciate them being part of the bundle :)

lovell:
Hi Espen, thanks for helping to maintain the TypeScript definitions, I'd be happy to make these "official" and integrate into the sharp repo.

The DefinitelyTyped repo is MIT licensed, which should allow us to copy a snapshot of the current implementation and redistribute under Apache 2.0 but with an MIT notice in the definition file.

We should introduce some tooling to help verify types, I'm thinking at least https://www.npmjs.com/package/tsd - anything else you're aware of?

Would including type definitions in the distributed package count as a (slightly) breaking change? If so, we should aim for v0.32.0 rather than a patch release.

lovell:
v0.32.0 now available with bundled TypeScript definitions, thanks for the PR.

Are you able to start the deprecation/removal process for the `@types/sharp` package? https://github.com/DefinitelyTyped/DefinitelyTyped#removing-a-package

[...]

PR #3370, opened 2022-09-17
Include TypeScript definitions in bundle

rexxars:
Treat this as a work-in-progress that we can use to base our discussions from.
Addresses #3369

In essence, this lifts the types from the `@types/sharp` module into the official package, along with the tests that were in that package as well.

Because the types reference node.js types such as `Buffer`, we also need to add a new dependency: `@types/node`. The DefinitelyTyped modules uses the wildcard version range (`*`) in order to prevent multiple copies of the package in the module tree, which often leads to hard to diagnose problems. An alternative would be for us to use `>=14.0.0` or similar. Open to suggestions.

As @lovell pointed out in #3369, this should be considered a minor breaking change, so should likely be released as 0.32.0. When that happens, we can also deprecate `@types/sharp` using the [official process](https://github.com/DefinitelyTyped/DefinitelyTyped#removing-a-package).

Some thoughts:
- While I don't _know_ of any errors in the typings, there is always a chance that there are. I wouldn't be surprised to see a few more PRs popping up to address issues.
- There is starting to be a bit of duplication in documenting things - there is a chance that the jsdoc and the tsdoc is slightly out of sync. Ideally we'd find a better process here, or even consider doing "typescript first" and compiling to js (this is probably controversial - just thinking aloud).
- The "type tests" are somewhat spotty in coverage, and very unstructured. Unless there are plans to migrate to TypeScript fully, I would probably want to restructure these tests a bit over time - splitting them into logical, "sorted" sections of some kind. I'm not sure if it's worth waiting for this work to be completed before a release however - open to suggestions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,6 +139,7 @@
     "tunnel-agent": "^0.6.0"
   },
   "devDependencies": {
+    "@types/node": "*",
     "async": "^3.2.4",
     "cc": "^3.0.1",
     "exif-reader": "^1.1.0",
--- a/package.json
+++ b/package.json
@@ -90,16 +90,18 @@
   "scripts": {
     "install": "(node install/libvips && node install/dll-copy && prebuild-install) || (node install/can-compile && node-gyp rebuild && node install/dll-copy)",
     "clean": "rm -rf node_modules/ build/ vendor/ .nyc_output/ coverage/ test/fixtures/output.*",
-    "test": "npm run test-lint && npm run test-unit && npm run test-licensing",
+    "test": "npm run test-lint && npm run test-unit && npm run test-licensing && npm run test-types",
     "test-lint": "semistandard && cpplint",
     "test-unit": "nyc --reporter=lcov --reporter=text --check-coverage --branches=100 mocha",
     "test-licensing": "license-checker --production --summary --onlyAllow=\"Apache-2.0;BSD;ISC;MIT\"",
     "test-leak": "./test/leak/leak.sh",
+    "test-types": "tsd",
     "docs-build": "node docs/build && node docs/search-index/build",
     "docs-serve": "cd docs && npx serve",
     "docs-publish": "cd docs && npx firebase-tools deploy --project pixelplumbing --only hosting:pixelplumbing-sharp"
   },
   "main": "lib/index.js",
+  "types": "lib/index.d.ts",
   "files": [
     "binding.gyp",
     "install/**",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -130,6 +132,7 @@
     "vips"
   ],
   "dependencies": {
+    "@types/node": "*",
```
lovell, line 135: Do we need this? My understanding is that it's the responsibility of a consuming TypeScript-based app to define this dependency, rather than doing so in every package that happens to provide type definitions. This approach also avoids the versioning problem you mention.
```sh
$ du -sh node_modules/@types/node
1.9M	node_modules/@types/node
```

lovell (commented):
Thank you for making a start on this, as you suggest it will need to be part of v0.32.0. I've left one question inline.

package.json
rexxars: 1.9M worth of typings 👀  Color me surprised.

I honestly don't quite know what the "right" approach is here, or if there even is one:
- A direct dependency will work and have typescript be happy when using the module
- A peer dependency marked as optional might be one option if we want to be explicit about what it depends on
- Dropping it altogether (or rather, moving it to a dev dependency - I think we'll need it for tsd to work) is probably also acceptable, since as you say - people using sharp + typescript is likely to install `@types/node` in their app anyway

lovell: Let's go with including it in `devDependencies`, which I think is the least intrusive option. We should also add a section to the installation docs about TypeScript support, including the need to install `@types/node` yourself. I can add this later.

I'm also bracing for future PRs that attempt to move `@types/node` to `dependencies` :)

rexxars: > I'm also bracing for future PRs that attempt to move `@types/node` to `dependencies` :)

Haha, it is bound to happen! 👍

rexxars: Moved to dev dependencies and force-pushed.

rexxars:
Let me know if you want any more changes done to this!

rexxars:
I have rebased this PR and lifted the changes from https://github.com/DefinitelyTyped/DefinitelyTyped/pull/63711 - which incorporates the changes that were added in sharp v0.31.3, along with the added tests. Squashed the changes into the previous commits to keep the commits clean. Also added @akwiatek as an author of the typings.

rexxars:
Nothing else I want to add! Rebased on `flow` and fixed the conflict 👍

lovell:
Thank you very much Espen, these will be published as part of v0.32.0.

Thanking you also in advance should I need to draw on your expertise to help review any PRs that arise due to this :+1:

lovell:
The definition "diff" for the forthcoming v0.32.0 release is in commit https://github.com/lovell/sharp/commit/a532659b0fefbeb0c2c18eba7f56dd36932abaa4 - if you spot anything else out of sync, please do open a PR.

PR #3377, opened 2022-09-20
GitHub Workflows security hardening

sashashura:
This PR adds explicit [permissions section](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions) to workflows. This is a security best practice because by default workflows run with [extended set of permissions](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) (except from `on: pull_request` [from external forks](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)). By specifying any permission explicitly all others are set to none. By using the principle of least privilege the damage a compromised workflow can do (because of an [injection](https://securitylab.github.com/research/github-actions-untrusted-input/) or compromised third party tool or action) is restricted.
It is recommended to have [most strict permissions on the top level](https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions) and grant write permissions on [job level](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs) case by case.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,8 +2,15 @@ name: CI (GitHub)
 on:
   - push
   - pull_request
+permissions:
+  contents: read # to fetch code (actions/checkout)
+
 jobs:
   CI:
+    permissions:
+      contents: read # to fetch code (actions/checkout)
+      checks: write # to create new checks (coverallsapp/github-action)
+
     runs-on: ${{ matrix.os }}
     container: ${{ matrix.container }}
     strategy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,13 +2,11 @@ name: CI (GitHub)
 on:
   - push
   - pull_request
-permissions:
-  contents: read # to fetch code (actions/checkout)
-
+permissions: {}
 jobs:
   CI:
     permissions:
-      contents: read # to fetch code (actions/checkout)
+      contents: write # for npx prebuild to make release
       checks: write # to create new checks (coverallsapp/github-action)
 
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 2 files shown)


lovell:
Hello, thank you for the PR. As you may have noticed this repo is configured to always require manual approval before any workflow runs, but I'd be happy to strengthen this further, especially as this could help improve the security of forks.

The workflows require the ability to create tags, releases and release artefacts, which I think means we'll also require the addition of `contents: write` at the job level.

sashashura:
Permissions are not as granular as they could be. `contents: write` covers all the cases you mentioned.

sashashura:
But currently I set contents to read. Do I understand correctly that
```
        env:
          prebuild_upload: ${{ secrets.GITHUB_TOKEN }}
        run: npx prebuild --runtime napi --target 7
```
creates GitHub release and needs the write permission?

lovell:
Yes, the prebuild step creates a release (if it doesn't already exist) with the same name as the tag and adds artefacts to that release.

sashashura:
Just for the reference these are current permissions it runs with:
image

.github/workflows/ci.yml
```diff
@@ -2,8 +2,13 @@ name: CI (GitHub)
 on:
   - push
   - pull_request
+permissions: {}
 jobs:
   CI:
+    permissions:
```
lovell, line 8: Is there a benefit in listing permissions under a job instead of a workflow? Should these two files use a consistent approach?

lovell (commented):
Thanks for the update, one more question inline otherwise this is good to go.

.github/workflows/ci.yml
sashashura: The only benefit is that if later somebody adds a new job it will inherit the global permission automatically, so it is safer to have read-only at the top. I made it consistent now.

lovell:
Thank you!

PR #3410, opened 2022-10-17
Update word format in resize docs

ginoemiliozzi:
Just a little update to avoid confusing fit as a valid keyword for the transformation request


lovell:
Thank you for the PR, please can you update the JSDoc in the source then use `npm run docs-build` to regenerate the documentation.

https://github.com/lovell/sharp/blob/eacb8337fa0b868fcdffefd422acee4816fc428f/lib/resize.js#L114

ginoemiliozzi:
@lovell 
Sorry I didn't know it has to be done this way, now I did it through the JSDoc as you suggested :+1:

PR #3442, opened 2022-11-08
Fix errors for missing OpenJPEG

RReverser:
Fixes couple of minor issues with JP2 errors:

1. The tests passed as false-positives even if regex is changed to arbitary pattern, because the promise returned from `assert.rejects` was ignored and the test ended prematurely. This is fixed by removing `{ ... }` around the test function body.
2. This, in turn, hid an issue with `toFile` not throwing the expected error message which was instead propagating `Error: VipsOperation: class "jp2ksave" not found` from libvips. This is now fixed by manually checking the extension before calling into libvips.
3. Pre-creating error instances like `errJp2Save` did is sometimes tempting, but is problematic for debugging because it hides the actual stacktrace of the error (the stacktrace is collected at the moment of `new Error` creation). This is now turned into a function that creates error with the right stack.

### DIFF
--- a/lib/output.js
+++ b/lib/output.js
@@ -25,7 +25,9 @@ const formats = new Map([
   ['j2c', 'jp2']
 ]);
 
-const errJp2Save = new Error('JP2 output requires libvips with support for OpenJPEG');
+const jp2Regex = /\.jp[2x]|j2[kc]$/i;
+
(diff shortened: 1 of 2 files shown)


lovell:
Thank you very much Ingvar.

PR #3449, opened 2022-11-10
Run tests in parallel

RReverser:
- Run tests in parallel. On my ThinkPad laptop with 8 cores, this reduces tests execution time from 60-70 seconds to 30-35 seconds (essentially ~2x faster turnaround).
 - Move Mocha settings to a config file. This allows to easily invoke Mocha alone via `npx mocha`, or for e.g. VSCode Test Explorer to find & run tests with the correct settings automatically.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,7 +92,7 @@
     "clean": "rm -rf node_modules/ build/ vendor/ .nyc_output/ coverage/ test/fixtures/output.*",
     "test": "npm run test-lint && npm run test-unit && npm run test-licensing",
     "test-lint": "semistandard && cpplint",
-    "test-unit": "nyc --reporter=lcov --reporter=text --check-coverage --branches=100 mocha --slow=1000 --timeout=30000 ./test/unit/*.js",
+    "test-unit": "nyc --reporter=lcov --reporter=text --check-coverage --branches=100 mocha --parallel --slow=1000 --timeout=30000 ./test/unit/*.js",
     "test-licensing": "license-checker --production --summary --onlyAllow=\"Apache-2.0;BSD;ISC;MIT\"",
     "test-leak": "./test/leak/leak.sh",
     "docs-build": "documentation lint lib && node docs/build && node docs/search-index/build",
@@ -139,6 +139,7 @@
     "tunnel-agent": "^0.6.0"
   },
   "devDependencies": {
+    "@tybys/emnapi": "^0.19.2",
     "async": "^3.2.4",
     "cc": "^3.0.1",
     "documentation": "^14.0.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -139,6 +139,7 @@
     "tunnel-agent": "^0.6.0"
   },
   "devDependencies": {
+    "@tybys/emnapi": "^0.19.2",
```
kleisauke, line 142: This added dependency should probably be omitted.

kleisauke: ... for now, at least :)

RReverser: Yes, that's accidental leak from rebase.

lovell: Ha ha, this definitely doesn't give the game away ;)

RReverser: Wait for it 😛

lovell:
Thank you for this improvement Ingvar.

There are a few [util tests](https://github.com/lovell/sharp/blob/main/test/unit/util.js) (cache, concurrency, counters) that require sequential running and may sporadically fail or cause other tests to fail when run in parallel. We should ensure these run after other tests have completed, perhaps via some kind of "after all" hook.

RReverser:
> There are a few [util tests](https://github.com/lovell/sharp/blob/main/test/unit/util.js?rgh-link-date=2022-11-10T20%3A13%3A04Z) (cache, concurrency, counters) that require sequential running and may sporadically fail or cause other tests to fail when run in parallel.

I actually did run into that in my _ahem_ other branch, but not with native build, even though I tried even running those asserts in a tight loop while Mocha was executing other test files.

Note that Mocha's parallel mode only parallelises tests at file-level, rather than per-test level, so I guess what happens is that each individual test file gets clean environment with its own copy of native module anyway, hence the counters are indeed zero in the individual test.

RReverser:
If I try to add e.g. `console.log("sharp init")` at the top level of `constructor.js`, I also see number of logs corresponding to my number of CPUs:

```
sharp init
sharp init
sharp init
sharp init
sharp init
sharp init
sharp init
```

So it does reuse the instances, but within each instance tests are running one by one and can't step on each others' toes and concurrently modify globals - that's why counters test passes successfully.

lovell:
Great, it looks like the overall situation has improved since I last tried this, e.g. commits https://github.com/lovell/sharp/commit/c4bcd088fbc7bb3d0833372df9031cdc7569dbce and https://github.com/lovell/sharp/commit/d7d580ae6f9bada7edea941e8f2c0efdc324d793

Let's merge and should we experience flaky tests we can deal with specific failures, if any. Thanks again!

RReverser:
No worries, and yeah, it's certainly a valid concern and something I tried to test extensively before making the PR. And I'm still not 100% sure - parallel stuff is always hard 😅 Happy to make another one if something does turn out flaky.

PR #3450, opened 2022-11-11
Convert mocha hooks

RReverser:
Forgot to do this in #3449... Mocha hooks need to be added via root hook plugins instead of somewhere in the test files, specifically because each file is treated as individual test and might run in a separate thread in parallel mode.

This didn't change the test results, but best to do the beforeEach cleanup anyway.


lovell:
Good spot, thanks Ingvar, looks like there's a couple of linter failures to deal with then we can get this merged.
```
semistandard: Run `semistandard --fix` to automatically fix some problems.
  /__w/sharp/sharp/test/beforeEach.js:13:13: Missing space before function parentheses.
  /__w/sharp/sharp/test/beforeEach.js:19:12: Missing space before function parentheses.
  /__w/sharp/sharp/test/beforeEach.js:23:4: Unexpected trailing comma.
```

RReverser:
My bad, I was sleepy and didn't run lints. Will fix up once near the computer.

RReverser:
Fixed.

Btw, I wonder why Github says "First-time contributors need a maintainer to approve running workflows." even though I already sent #3449 so I'm not a first-time contributor 🤔

RReverser:
Also related: I didn't want to change it in case I don't understand it, but I find this line weird:

```js
const concurrency = process.env.VIPS_CONCURRENCY ||
  (libcFamily === detectLibc.MUSL || process.arch === 'arm' ? 1 : undefined);
```

My understanding is that passing `undefined` to `sharp.concurrency()` is no-op, as it will only act as a getter instead of a setter in that case. Is that intentional? Or was this meant to be `0` (zero) instead, to reset concurrency to the number of CPU cores before each test?

lovell:
> First-time contributors...

This is a mis-labelling by GitHub, as the repo was originally configured to require approval for "outside collaborators". The macOS ARM64 CI environment runs on an M1 Mac mini kindly provided by MacStadium and there was a spate of cr*pto-mining spam PRs affecting GitHub when I was first setting this up that I didn't want running on hardware I am responsible for. It's probably time to relax this setting, which I've now done.

>  I find this line weird

Yeah, sorry about all the weird historic internals of sharp you're having to look at, I think this was to allow the tests to run on various hardware/conditions/CI providers. These can definitely be simplified e.g. look for a single environment variable.

lovell:
Simplified via https://github.com/lovell/sharp/commit/4490a93430e5e85b03ed122bbde487c319e3acef - thank you for highlighting this.

PR #3460, opened 2022-11-18
CI: Upgrade to Python 3.11

cclauss:
https://docs.python.org/3.11/whatsnew/3.11.html
> Python 3.11 is between 10-60% faster than Python 3.10. On average, we measured a 1.25x speedup on the standard benchmark suite.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,11 +78,11 @@ jobs:
       - name: Dependencies (Linux musl)
         if: contains(matrix.container, 'alpine')
         run: apk add build-base git python3 font-noto --update-cache
-      - name: Dependencies (Python 3.10 - macOS, Windows)
+      - name: Dependencies (Python 3.11 - macOS, Windows)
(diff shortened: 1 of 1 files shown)


lovell:
Hi Christian, thank you for the PR.

Python is currently pinned to 3.10 as node-gyp version 6.x that ships with Node.js 14 does not work with Python 3.11.

It looks like it was you that fixed node-gyp to support Python 3.11, available since version 8.x, so thank you very much for doing so.

https://github.com/nodejs/node-gyp/commit/c3c510d89ede3a747eb679a49254052344ed8bc3

When we drop support for Node.js 14 sometime after May next year we can unpin Python. Happy to leave this PR open until then, or perhaps mark as draft, so we don't forget.

birkskyum:
I guess this can proceed now. [Node 16 is EOL within a month as well](https://nodejs.org/en/blog/announcements/nodejs16-eol).

lovell:
Thank you Christian.

Issue #1648, opened 2019-04-05
Replace Color with transparency


yasso1am:
What are you trying to achieve?

I have a known rgb value that surrounds the screenshot I am taking, and would like to replace it with a white, fully transparent background. I have not seen examples of how to do this.

Have you searched for similar questions?

Yes. It seems most people want to take a transparent background and provide it with a color. I am looking to take a known color and change it, as well as make it transparent.

Are you able to provide a standalone code sample that demonstrates this question?

```
const trimWhiteSpaceAndPad = (path) => {
    sharp(path)
    .trim()
    .resize(175, 180, {
        fit: 'contain',
        background: {
            r: 225,
            g: 227,
            b: 237,
        }
    })
    .extend({
        left: 5,
        right: 5,
        top: 0,
        bottom: 0,
        background: {
            r: 225,
            g: 227,
            b: 237,
            alpha: 0
        }
    })
    .toFile(`trimmed_${path}`, (err) => {
            err && console.log('Error trimming file:' + err)
        })
}
```

Are you able to provide a sample image that helps explain the question?

I would like to be able to take all of the off-white (225,227,237) area that edges up against the model of these teeth and turn it white and transparent. I acquired this photo with the code above.

Here is my input file:

![beforeLower-6caee480-ab4e-11e8-a183-6d4ab07faa80](https://user-images.githubusercontent.com/39067243/55645744-3359b380-578e-11e9-8d55-82f454a3b2c7.png)

Here is my output file:

![trimmed_beforeLower-6caee480-ab4e-11e8-a183-6d4ab07faa80](https://user-images.githubusercontent.com/39067243/55645766-4076a280-578e-11e9-8813-f1ea1181c4e8.png)

I would like if all the outside area was transparent. It would be great if I could call something like

`.replace({ input: {// color I want to replace here}, output: {// color I want it to be} })`

lovell:
Hello, although libvips provides some "flood fill" features, they are not yet exposed in sharp. I'll tag this as a future possible enhancement.

For now your best bet is to use [raw](https://sharp.pixelplumbing.com/en/stable/api-output/#raw) pixel output then loop over and selectively replace values.

yasso1am:
> Hello, although libvips provides some "flood fill" features, they are not yet exposed in sharp. I'll tag this as a future possible enhancement.
> 
> For now your best bet is to use [raw](https://sharp.pixelplumbing.com/en/stable/api-output/#raw) pixel output then loop over and selectively replace values.

[...]

PR #3461, opened 2022-11-19
Unflatten function - create an alpha channel and make white pixels fully transparent

antonmarsden:
@lovell A little something I'm working on. It creates a hard (0/255) alpha channel based on the pixel colour. Very keen to get your feedback on this. Note that it relates to https://github.com/lovell/sharp/issues/1648

It's currently limited to making lighter areas transparent (but easy enough to change how the threshold rules work). I'm still thinking about how to generalise it for mainstream use.

See provided images for example output (input was from the test image suite).

unflatten-swiss
unflatten-white-transparent


lib/operation.js
```diff
@@ -402,6 +402,27 @@ function flatten (options) {
   return this;
 }
 
+/**
+ * Unflatten - apply transparency (0 /255) to the image using a threshold.
+ *
+ * @example
+ * await sharp(rgbInput)
+ *   .unflatten({ unflattenThresholds: [255] })
```
lovell, line 410: Should this be `threshold` instead of the (internal) `unflattenThresholds`? I think the singular `threshold` might be better for the public API, plus maybe we could allow a `number` as well (then auto-generate a single-element array)?

src/operations.cc
```diff
@@ -351,6 +351,31 @@ namespace sharp {
     }
   }
 
+  /*
+   * Unflatten
+   */
+  VImage Unflatten(VImage image, std::vector<double> const thresholds) {
+    if (HasAlpha(image)) {
```
lovell, line 358: Would we ever want to discard the existing alpha channel?

src/operations.cc
```diff
@@ -351,6 +351,31 @@ namespace sharp {
...
+      throw VError("Cannot unflatten an image that already has an alpha channel");
+    }
+    size_t const bands = static_cast<size_t>(image.bands());
+    if (thresholds.size() != 1 && thresholds.size() != bands) {
+      throw VError("band / threshold count mismatch");
```
lovell, line 363: Perhaps add something to this message about the error originating from `unflatten`. Is it worth testing this path also?

lovell (commented):
Thanks again for this PR, sorry it took me a while to get around to reviewing it properly. I've added a couple of questions/thoughts inline.

antonmarsden:
@lovell Thanks for the feedback, I'll find some time to incorporate it over the next few weeks.

lib/operation.js
antonmarsden: Good call... I have simplified unflatten to convert only white to transparent, and placed it after **threshold** in the pipeline.

src/operations.cc
antonmarsden: Likely not :) The new implementation will convert white to transparent, and leave non-white alpha values as is.

antonmarsden:
@lovell Looks like prebuild is failing on arm64. I am doing an image compare with threshold=0. Not sure why it's breaking at this stage. Are there any platform-specific quirks to be aware of?

lovell:
Thank you for the updates Anton, this looks great.

Please can you add TypeScript definitions to [index.d.ts](https://github.com/lovell/sharp/blob/main/lib/index.d.ts) then this is good to merge.

antonmarsden:
> Thank you for the updates Anton, this looks great.
> 
> Please can you add TypeScript definitions to [index.d.ts](https://github.com/lovell/sharp/blob/main/lib/index.d.ts) then this is good to merge.

Done, good to go.

lovell (approved):
Thanks for the updates, looks great, I'll merge shortly.

fyyyyy:
If there was a backgroundColor and threshold option for unflatten, it would be very useful. I.e. to remove background colors typically found in video game sprites.
image

fyyyyy:
maybe a similar api to the trim function
https://github.com/lovell/sharp/pull/3332

Issue #3480, opened 2022-12-06
Expected behavior of `failOn`


LekoArts:
## Question about an existing feature

Hi! I'd like some clarification about the `failOn` feature and how it's expected to behave. Right now I'm not sure if I'm doing something wrong or if it just will work like that.

### What are you trying to achieve?

While working on https://github.com/gatsbyjs/gatsby/pull/37177 I wanted to add a test case to verify that the expected behavior works. We often wrap sharp like this:

```js
try {
  const pipeline = sharp({ failOn: pluginOptions.failOn })
  fs.createReadStream(file.absolutePath).pipe(pipeline)

  metadata = await pipeline.metadata()
} catch (err) {
  reportError(
    `Failed to retrieve metadata from image ${file.absolutePath}`,
    err,
    reporter
  )
  return null
}
```

The goal of the above PR is that when `failOn: 'none'` is set the `reportError` function shouldn't be called. For this test case I then created a `.txt` file, added some text to it and then renamed the file to have a `.png` extension. So a completely broken .png image really.

Now, even with failOn set to none, it still throws an error and in the catch the reportError is called. Is this expected behavior? Is there some threshold of what sharp will accept? I understood the option with set to none as "please don't throw any errors".

### Please provide a minimal, standalone code sample, without other dependencies, that demonstrates this question

See explanation above.

LekoArts:
Thanks for the answer!

When I tried that it errors with:

```shell
Failed to retrieve metadata from image /Users/lejoe/code/work/gatsby/packages/gatsby-plugin-sharp/src/__tests__/images/invalid.png [Error: Input buffer has corrupt header: pngload_buffer: end of stream]
```

When I tried any of the images in here https://github.com/YahooArchive/pngjs-image/tree/master/test/png/PngSuite/corrupt I get:

```shell
Failed to retrieve metadata from image /Users/lejoe/code/work/gatsby/packages/gatsby-plugin-sharp/src/__tests__/images/invalid.png [Error: Input buffer has corrupt header: pngload_buffer: invalid color type]
```

So I suppose (because I don't know how to check) that those corrupt png images have these 8 magic bytes at the beginning but it still fails?

lovell:
Apologies, there needs to be enough of a valid header beyond the magic bytes so that libvips can read out basic settings and therefore the various decoder libraries can attempt to start to decode pixel values.

[...]

PR #3481, opened 2022-12-07
chore(docs): Clarify `failOn` setting

LekoArts:
## Description

Clarification to the `failOn` setting coming from the discussion in https://github.com/lovell/sharp/issues/3480. Goal of this PR is to set correct expectations of when `failOn` is actually applied internally and which images are always expected to fail regardless of the setting of this

## Related Issues

Fixes https://github.com/lovell/sharp/issues/3480

### DIFF
--- a/docs/api-constructor.md
+++ b/docs/api-constructor.md
@@ -20,7 +20,7 @@ Implements the [stream.Duplex][1] class.
     JPEG, PNG, WebP, AVIF, GIF, SVG, TIFF or raw pixel image data can be streamed into the object when not present.
 *   `options` **[Object][13]?** if present, is an Object with optional attributes.
 
-    *   `options.failOn` **[string][12]** level of sensitivity to invalid images, one of (in order of sensitivity): 'none' (least), 'truncated', 'error' or 'warning' (most), highers level imply lower levels. (optional, default `'warning'`)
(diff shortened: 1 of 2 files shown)


docs/api-constructor.md
```diff
@@ -20,7 +20,7 @@ Implements the [stream.Duplex][1] class.
     JPEG, PNG, WebP, AVIF, GIF, SVG, TIFF or raw pixel image data can be streamed into the object when not present.
 *   `options` **[Object][13]?** if present, is an Object with optional attributes.
 
-    *   `options.failOn` **[string][12]** level of sensitivity to invalid images, one of (in order of sensitivity): 'none' (least), 'truncated', 'error' or 'warning' (most), highers level imply lower levels. (optional, default `'warning'`)
+    *   `options.failOn` **[string][12]** Level of sensitivity to invalid images in terms of pixel values. Invalid image headers/metadata will always fail regardless of this setting. One of (in order of sensitivity): 'none' (least), 'truncated', 'error' or 'warning' (most), highers level imply lower levels. (optional, default `'warning'`)
```
LekoArts, line 23: I don't understand enough of sharp to better word this sentence:

> Level of sensitivity to invalid images in terms of pixel values.

This needs more clarity

lovell:
Thank you very much for the PR, how about something like:

> when to abort processing of invalid pixel data, one of (in order of sensitivity): 'none' (least), 'truncated', 'error' or 'warning' (most), higher levels imply lower levels, invalid metadata will always abort

Please can you update the JSDocs then use `npm run docs-build` to automagically update the markdown.

https://github.com/lovell/sharp/blob/2a71f1830f7e649e5e8349565d32fd4e7c98d61e/lib/constructor.js#L122

LekoArts:
Done 👍

Issue #3239, opened 2022-05-26
Sharp cannot detect globally installed libvips (Apple M1)


hyunjinjeong:
## Possible install-time or require-time problem

- [x] I have read the [documentation relating to installation](https://sharp.pixelplumbing.com/install).
- [x] I have ensured that the architecture and platform of Node.js used for `npm install` is the same as the architecture and platform of Node.js used at runtime.

### Are you using the latest version of sharp?

- [x] I am using the latest version of `sharp` as reported by `npm view sharp dist-tags.latest`.

If you cannot confirm this, please upgrade to the latest version and try again before opening an issue.

If you are using another package which depends on a version of `sharp` that is not the latest, please open an issue against that package instead.

### Is this a problem with filesystem permissions?

If you are using npm v6 or earlier and installing as a `root` or `sudo` user, have you tried with the `npm install --unsafe-perm` flag?

If you are using npm v7 or later, does the user running `npm install` own the directory it is run in?

If you are using the `ignore-scripts` feature of `npm`, have you tried with the `npm install --ignore-scripts=false` flag?

### What is the complete output of running `npm install --verbose --foreground-scripts sharp` in an empty directory?

<details>

```bash
npm verb cli /Users/user/.nvm/versions/node/v14.17.6/bin/node /Users/user/.nvm/versions/node/v14.17.6/bin/npm
npm info using npm@8.11.0
npm info using node@v14.17.6
npm timing npm:load:whichnode Completed in 0ms
npm timing config:load:defaults Completed in 2ms
npm timing config:load:file:/Users/user/.nvm/versions/node/v14.17.6/lib/node_modules/npm/npmrc Completed in 2ms
npm timing config:load:builtin Completed in 2ms
npm timing config:load:cli Completed in 3ms
npm timing config:load:env Completed in 0ms
npm timing config:load:file:/Users/user/dev/sharp-test/.npmrc Completed in 0ms
npm timing config:load:project Completed in 4ms
npm timing config:load:file:/Users/user/.npmrc Completed in 0ms
npm timing config:load:user Completed in 0ms
npm timing config:load:file:/Users/user/.nvm/versions/node/v14.17.6/etc/npmrc Completed in 0ms
npm timing config:load:global Completed in 0ms
npm timing config:load:validate Completed in 0ms
npm timing config:load:credentials Completed in 0ms
npm timing config:load:setEnvs Completed in 1ms
npm timing config:load Completed in 14ms
[...]

PR #3514, opened 2023-01-04
Get real architecture of M1 Mac regardless of Rosetta

Xheldon:
Get real architecture of M1 Mac regardless of Rosetta or node version, see #3239


lovell:
Logging added and docs updated via commit https://github.com/lovell/sharp/commit/5be36c2deb735577fc76fa52242836d40df276bd

PR #3541, opened 2023-01-29
Clarify supported bit depth for AVIF images with prebuilt libvips

paescuj:
Ref https://github.com/libvips/libvips/issues/3274#issuecomment-1406341947


lovell:
Thank you!

Issue #3534, opened 2023-01-17
Feature: extrude image


miltoncandelero:
### What are you trying to achieve?

currently `extend` uses a single color for all pixels, `extrude` would pick the closest pixel color

### What would you expect the API to look like?
the same as the `extend` api but without color
``` sharp.extrude(10)```
and
```
  sharp.extrude({
    top: 10,
    bottom: 20,
    left: 10,
    right: 10,
  })
```
### What alternatives have you considered?

extending with transparent and then going pixel by pixel replacing the transparent pixel. This defeats the idea of sharp.

### Please provide sample image(s) that help explain this feature

![image](https://user-images.githubusercontent.com/15253010/212920067-c4abfd71-16ad-4a85-81c8-dfc93a696335.png)

**Comment by lovell** [comment]
There's an existing `extend` option of `vips_embed` (as used by sharp's `extend()` method) that we could expose to support the scenario provided.

https://www.libvips.org/API/current/libvips-conversion.html#vips-embed

This would require `VIPS_EXTEND_COPY` (as `copy`), and I think we should also support at least `mirror` and `repeat`. A caller would need to set either the proposed `extend` option _or_ the existing `background` option, but not both.

```js
// PROPOSED API, NOT YET AVAILABLE
sharp.extend({
  ...
  extend: 'copy' | 'mirror' | 'repeat'
})
// PROPOSED API, NOT YET AVAILABLE
```

tquetano-netflix:
I am actually trying to accomplish this exact feat right now, and its quickly becoming a blocker for me to replace our old system that uses ImageMagick with sharp, so you can count on my vote.

Is there even a way to accomplish this in the current API?

PR #3556, opened 2023-02-10
Support different extend modes using "extend.extend" config item

janaz:
This is my attempt at implementing the feature asked for in #3534  

@lovell suggested the following API for it:
```js
// PROPOSED API, NOT YET AVAILABLE
sharp.extend({
  ...
  extend: 'copy' | 'mirror' | 'repeat'
})
// PROPOSED API, NOT YET AVAILABLE
```

This PR adds a new config item to `sharp.extend` as specified in the proposal. The supported values for `extend.extend` are:

* `background` (colour set from the _background_ property) - this is the current behaviour (default)
extend-equal-background

* `copy` (copy the image edges)
extend-equal-copy

* `repeat` (repeat the whole image)
extend-equal-repeat

* `mirror` (mirror the whole image)
extend-equal-mirror


lovell:
Hi Tomasz, good to hear from you, thank you very much for this.

It's interesting that you chose `ExtendMode` as the type name, perhaps we should

lib/index.d.ts
```diff
@@ -1243,6 +1243,8 @@ declare namespace sharp {
         sigma?: number | undefined;
     }
 
+    type ExtendMode = 'background' | 'copy' | 'repeat' | 'mirror';
```
lovell, line 1246: Sorry for the API naming confusion, which is my fault as I suggested it in the first place, but I think this type would be less confusing when named `ExtendWith`.

lovell (changes requested):
This is brilliant, thank you again Tomasz. I've left a couple of comments inline and I've given the public API some more thought too with one small change, but otherwise this looks great and is almost ready to go.

src/pipeline.cc
```diff
@@ -1499,6 +1518,7 @@ Napi::Value pipeline(const Napi::CallbackInfo& info) {
   baton->extendLeft = sharp::AttrAsInt32(options, "extendLeft");
   baton->extendRight = sharp::AttrAsInt32(options, "extendRight");
   baton->extendBackground = sharp::AttrAsVectorOfDouble(options, "extendBackground");
+  baton->extendMode = sharp::AttrAsInt32(options, "extendMode");
```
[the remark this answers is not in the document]
janaz: Thanks @lovell. I agree that using an enum type is better in this case.

lib/resize.js
```diff
@@ -393,6 +405,13 @@ function extend (extend) {
       }
     }
     this._setBackgroundColourOption('extendBackground', extend.background);
+    if (is.defined(extend.extend)) {
```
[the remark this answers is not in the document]
janaz: 👍

lib/index.d.ts
janaz: No worries. It was clearly stated that it was just a proposal. Names of public API attributes are hard to change and now is the best time to think about it.

janaz:
I pushed a commit with all the requested changes. Switching to `sharp::AttrAsEnum<VipsExtend>` made the code a lot simpler!

src/operations.cc
```diff
@@ -395,11 +395,11 @@ namespace sharp {
    * Split into frames, embed each frame, reassemble, and update pageHeight.
    */
   VImage EmbedMultiPage(VImage image, int left, int top, int width, int height,
-                        std::vector<double> background, int nPages, int *pageHeight) {
+                        VipsExtend extendWith, std::vector<double> background, int nPages, int *pageHeight) {
     if (top == 0 && height == *pageHeight) {
       // Fast path; no need to adjust the height of the multi-page image
       return image.embed(left, 0, width, image.height(), VImage::option()
-        ->set("extend", VIPS_EXTEND_BACKGROUND)
+        ->set("extendWith", extendWith)
```
lovell, line 402:
```suggestion
        ->set("extend", extendWith)
```

lovell (changes requested):
One small error but otherwise this is good to go, thank you.

lovell (approved):

lovell:
Thanks again, this will be in v0.32.0.

PR #3568, opened 2023-02-23
Update install.md

cychub:


docs/install.md
```diff
@@ -147,12 +147,12 @@ or the `npm_config_sharp_local_prebuilds` environment variable.
...
 if `sharp_binary_host` is set to `https://hostname/path`
 and the sharp version is `1.2.3` then the resultant URL will be
-`https://hostname/path/sharp-v1.2.3-napi-v5-platform-arch.tar.gz`.
+`https://hostname/path/v1.2.3/sharp-v1.2.3-napi-v5-platform-arch.tar.gz`.
 
 Filename example:
 if `sharp_local_prebuilds` is set to `/path`
 and the sharp version is `1.2.3` then the resultant filename will be
-`/path/sharp-v1.2.3-napi-v5-platform-arch.tar.gz`.
+`/path/v1.2.3/sharp-v1.2.3-napi-v5-platform-arch.tar.gz`.
```
lovell, line 155: The local prebuilds feature of prebuild-install doesn't add the version-as-a-path-component, so please can you remove the change on this line.

https://github.com/prebuild/prebuild-install#local-prebuilds

docs/install.md
```diff
@@ -147,12 +147,12 @@ or the `npm_config_sharp_local_prebuilds` environment variable.
 URL example:
 if `sharp_binary_host` is set to `https://hostname/path`
 and the sharp version is `1.2.3` then the resultant URL will be
-`https://hostname/path/sharp-v1.2.3-napi-v5-platform-arch.tar.gz`.
+`https://hostname/path/v1.2.3/sharp-v1.2.3-napi-v5-platform-arch.tar.gz`.
```
lovell, line 150: Good spot, thank you.

lovell (changes requested):

docs/install.md
cychub: updated

docs/install.md
lovell: Thanks for the update, it looks like this line change is still present so marking as unresolved.

lovell:
Thank you very much!

Issue #3574, opened 2023-02-28
Converting 16 Bit JPEG2000 greyscale image to 8 bit JPG causes loss of image quality


LachlanNewman:
## Possible bug

When trying to convert a 16-bit greyscale JPEG2000 image to an 8-bit JPG image there is a loss in the quality of the output. (see attached `actual_output.jpg`

Other Useful Information
- When printing the stats of the image when it is first read, the `max pixel` value is 4 and the `min pixel` value is 0. this doesn't make any sense because when you open the image in a JPG2000-compatible platform (see https://nroduit.github.io/en/) it shows a much larger range of shades of grey.
- also after normalizing the image the `max pixel` value is 239 and the `min pixel` value is 15. Shouldn't it stretch the luminance range so it's between 0 and 255?

### Is this a possible bug in a feature of sharp, unrelated to installation?

- [x] Running `npm install sharp` completes without error.
- [x] Running `node -e "require('sharp')"` completes without error.

### Are you using the latest version of sharp?

- [x] I am using the latest version of `sharp` as reported by `npm view sharp dist-tags.latest`.

### What is the output of running `npx envinfo --binaries --system --npmPackages=sharp --npmGlobalPackages=sharp`?

```
  System:
    OS: macOS 12.3.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 65.33 MB / 32.00 GB
    Shell: 5.8 - /bin/zsh
  Binaries:
    Node: 16.14.0 - ~/.nvm/versions/node/v16.14.0/bin/node
    Yarn: 1.22.17 - ~/.nvm/versions/node/v16.14.0/bin/yarn
    npm: 8.3.1 - ~/.nvm/versions/node/v16.14.0/bin/npm
  npmPackages:
    sharp: ^0.31.3 => 0.31.3 
```

### What are the steps to reproduce?

1. Download the attached file `input.jp2`
2. run the sample script below in Node.

### What is the expected behaviour?

should convert the image to 8 bits. negate the image and normalize the pixels from 0 to 255 and produce an image similar to the attachment `expected_output.jpg`

### Please provide a minimal, standalone code sample, without other dependencies, that demonstrates this problem

```
import sharp from 'sharp'
import fs from 'fs'

async function convertToJpeg() {
      const imageBytes = fs.readFileSync(`path/to/input.jp2`);

      const output = await sharp(imageBytes)
      .gamma()
      .grayscale()
      .negate({alpha: true})
      .normalise()
[...]

PR #3578, opened 2023-03-01
feat: set lower and upper bin levels for normalise

LachlanNewman:
See https://github.com/lovell/sharp/issues/3574#issuecomment-1449124779

### DIFF
--- a/lib/operation.js
+++ b/lib/operation.js
@@ -474,13 +474,23 @@ function negate (options) {
  * @param {Boolean} [normalise=true]
  * @returns {Sharp}
  */
-function normalise (lowerBin, upperBin) {
-  if(!is.inRange(lowerBin,0,100 ) || !is.inRange(upperBin,0,100)){
-    throw new Error('Range must be between 0 and 100 inclusively')
+function normalise (options) {
+  if(is.plainObject(options)){
+    const lowerBin = 'lowerBin' in options ? options.lowerBin : 1
+    const upperBin = 'upperBin' in options ? options.upperBin : 99
+
+    if(!is.number(lowerBin) || !is.number(upperBin)){
+      throw new Error('lowerBin and upperBin must be of type number')
+    }
+
+    if(!is.inRange(lowerBin,0,100 ) || !is.inRange(upperBin,0,100)){
+      throw new Error('Range must be between 0 and 100 inclusively')
+    }
+
+    this.options.normaliseLowerBin = lowerBin
+    this.options.normaliseUpperBin = upperBin
   }
-  this.options.normalise = true;
-  this.options.normaliseLowerBin = is.number(lowerBin) ? lowerBin : 1;
-  this.options.normaliseUpperBin = is.number(upperBin) ? upperBin : 99;
+  this.options.normalise = is.bool(options) ? options : true;
   return this;
 }
 
@@ -493,8 +503,8 @@ function normalise (lowerBin, upperBin) {
  * @param {Boolean} [normalize=true]
  * @returns {Sharp}
  */
-function normalize (normalize, lowerBin, upperBin) {
-  return this.normalise(normalize, lowerBin, upperBin);
+function normalize (options) {
+  return this.normalise(options);
 }
 
 /**
--- a/lib/operation.js
+++ b/lib/operation.js
@@ -474,8 +474,8 @@ function negate (options) {
  * @param {Boolean} [normalise=true]
  * @returns {Sharp}
  */
-function normalise (normalise, lowerBin, upperBin) {
-  this.options.normalise = is.bool(normalise) ? normalise : true;
+function normalise (lowerBin, upperBin) {
+  this.options.normalise = true;
   this.options.normaliseLowerBin = is.number(lowerBin) ? lowerBin : 1;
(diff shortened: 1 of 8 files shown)


lovell:
Hi, thanks for the PR, did you need this change to solve your original problem?

LachlanNewman:
> Hi, thanks for the PR, did you need this change to solve your original problem?

Hey @lovell which problem are you referring to ?

jonseppanen:
This in general is a very welcomed/needed addition to the library. Being able to adjust images with weird levels (e.g all at one end or the other) is sorely needed.

lovell:
@LachlanNewman I was referring to #3574, which I think you've been able to solve by increasing brightness. I was wondering if the change in this PR also allowed you to solve the same problem without the need to alter brightness?

lib/operation.js
```diff
@@ -474,8 +474,10 @@ function negate (options) {
  * @param {Boolean} [normalise=true]
  * @returns {Sharp}
  */
-function normalise (normalise) {
+function normalise (normalise, lowerBin, upperBin) {
```
lovell, line 477: How about having the API look a bit more like the following?
```js
.normalise({ lower: 1, upper: 99 })
```
(I'm happy to drop the `normalise` as-a-Boolean parameter as it's a little confusing.)

lib/operation.js
```diff
@@ -474,8 +474,10 @@ function negate (options) {
...
   this.options.normalise = is.bool(normalise) ? normalise : true;
+  this.options.normaliseLowerBin = is.number(lowerBin) ? lowerBin : 1;
```
lovell, line 479: Could validate these via `is.inRange()`.

src/operations.cc
```diff
@@ -68,8 +67,8 @@ namespace sharp {
     // Extract luminance
     VImage luminance = lab[0];
     // Find luminance range
-    int const min = luminance.percent(1);
-    int const max = luminance.percent(99);
+    int const min = luminance.percent(lowerBin);
```
lovell, line 70: Could optimise slightly here when the values are 0 or 100, which would be `min()` or `max()` respectively.

test/unit/normalize.js
```diff
@@ -20,10 +20,8 @@ describe('Normalization', function () {
   it('spreads rgb image values between 0 and 255', function (done) {
     sharp(fixtures.inputJpgWithLowContrast)
       .normalise()
-      .raw()
-      .toBuffer(function (err, data, info) {
+      .toFile('test2.jpg', function (err) {
```
lovell, line 23: Was this change intentional?

lovell (changes requested):
Thanks again, I've left a few questions/comments inline.

LachlanNewman:
> @LachlanNewman I was referring to #3574, which I think you've been able to solve by increasing brightness. I was wondering if the change in this PR also allowed you to solve the same problem without the need to alter brightness?

Nope, #3574 was for a specific image with a very low range of pixels. This one is to allow normalizing the images based on a particular range of pixels.

test/unit/normalize.js
LachlanNewman: no this was an accident i have fixed it.

lib/operation.js
LachlanNewman: [resolved](https://github.com/lovell/sharp/pull/3578/commits/8045250a2697591b40bef01db1c9c067587b1973)

lib/operation.js
LachlanNewman: [resolved](https://github.com/lovell/sharp/pull/3578/commits/2b16433d97e4cb95a07369b5e8fa68cc80d9b905)

src/operations.cc
LachlanNewman: [resolved](https://github.com/lovell/sharp/pull/3578/commits/f0cd634c86f5734d2a534b4becc594170e8195df)

lovell (changes requested):
Thank you for the updates, one further comment inline. I'll update the target branch for this PR to `flow` for inclusion in the next release, please can you rebase against that branch. This also means you will need to include a TypeScript definition update in `lib/index.d.ts`.

lib/operation.js
```diff
@@ -474,8 +474,13 @@ function negate (options) {
  * @param {Boolean} [normalise=true]
  * @returns {Sharp}
  */
-function normalise (normalise) {
-  this.options.normalise = is.bool(normalise) ? normalise : true;
+function normalise (lowerBin, upperBin) {
+  if(!is.inRange(lowerBin,0,100 ) || !is.inRange(upperBin,0,100)){
```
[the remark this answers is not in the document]
LachlanNewman: fixed

LachlanNewman:
@lovell fyi see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/64616

LachlanNewman:
@lovell see https://github.com/lovell/sharp/pull/3583

lovell:
Sorry I wasn't clear about the TypeScript definition changes. Now that you've rebased against `flow` you'll hopefully have seen what used to be published as `@types/sharp` is now part of *this* repo in the `lib/index.d.ts` file and will be published as part of `@sharp` directly in the next release - please see #3369. This means any API changes in the next release now need to be included in this repo instead of DefinitelyTyped.

PR #3586, opened 2023-03-10
feat: add min and max luminence to stats

LachlanNewman:
Return the min and maximum luminance levels based on the `lch` colourspace

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1012,6 +1012,10 @@ declare namespace sharp {
         sharpness: number;
         /** Object containing most dominant sRGB colour based on a 4096-bin 3D histogram (experimental) */
         dominant: { r: number; g: number; b: number };
+        /** Maximum luminence after conversion to CIELAB color space */
+        maxLuminance: number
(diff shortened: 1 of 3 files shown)


lovell:
Hi, thanks for the PR, were you aware that it's already possible to achieve this using something like the following code:
```js
const { data, info } = await sharp(input)
  .toColorspace("lch")
  .extractChannel(0)
  .raw()
  .toBuffer({ resolveWithObject: true });

const { channels } = await sharp(data, { raw: info }).stats();
const { min, max } = channels[0];
```

LachlanNewman:
> Hi, thanks for the PR, were you aware that it's already possible to achieve this using something like the following code:
> 
> ```js
> const { data, info } = await sharp(input)
>   .toColourspace("lch")
>   .extractChannel(0)
>   .raw()
>   .toBuffer({ resolveWithObject: true });
> 
> const { channels } = await sharp(data, { raw: info }).stats();
> const { min, max } = channels[0];
> ```

No i wasn't aware thank you ! i will close this PR

PR #3674, opened 2023-05-20
fixed method toFile(), param "fileOut" jpeg2000 file extension check

bianjunjie1981:
fixed method toFile(), param "fileOut" jpeg2000 file extension check
------------------------------------
The RE matches the file extension directly rather than the entire path，
The result will be much closer to correct.

### DIFF
--- a/lib/output.js
+++ b/lib/output.js
@@ -75,7 +75,7 @@ function toFile (fileOut, callback) {
     err = new Error('Missing output file path');
   } else if (is.string(this.options.input.file) && path.resolve(this.options.input.file) === path.resolve(fileOut)) {
     err = new Error('Cannot use same file for input and output');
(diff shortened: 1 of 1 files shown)


lovell:
Thank you very much for the PR.

Would it be worth altering the regex to ensure the `\.` is treated independently of the `|` alternate capture, perhaps a bit like the following (untested)?
```diff
- const jp2Regex = /\.jp[2x]|j2[kc]$/i;
+ const jp2Regex = /\.(jp[2x]|j2[kc])$/i;
```

A suitable filename for testing purposes might be `test.failj2c`, and if you're able to add a unit test to help prevent future regression that would be great.

lovell:
Landed as https://github.com/lovell/sharp/commit/7e6a70af4433847a7ef740489ab6bfad4b09861a - thank you very much for the PR - this will be in v0.32.2.

Issue #3164, opened 2022-03-29
ava test framework with worker threads - Module did not self-register


simplenotezy:
## Possible bug

I've been noticing this is quite a normal error message to receive, but most of the suggestions recommend deleting node_modules, or similar - which is not really relevant in my case, since the error only happens during a Github Actions deployment.

My github actions step looks something like this:

```yaml
        run: |
          npm ci
          npm run build --if-present
          npm run test
```

- I have tried adding `npm install --platform=linux --ignore-scripts=false --arch=x64 sharp` after `npm ci` but it will give me the same error.
- I have even tried pushing a build without the `package-lock.json` to see if that would make a difference during deployment, but to no avail.
- I verified node and npm version are same on my local machine as well as on the Github actions runner
- Output of `npm install --platform=linux --arch=x64 --verbose sharp` in the github actions runner: https://gist.github.com/simplenotezy/bb1dc5fa506c8629c03d80703247969e

I am not really sure how to proceed here.

```sh
❯ npm ls sharp
project@0.0.1 /Users/mf/Projects/project/backend
└── sharp@0.30.3

❯ npm why sharp
sharp@0.30.3
node_modules/sharp
  sharp@"^0.30.3" from the root project
```

Not sure if it is related to worker threads (https://sharp.pixelplumbing.com/install#worker-threads). The error does occur when running my tests, which I believe could be multi-threaded (Using [Ava 4](https://github.com/avajs/ava))

### Is this a possible bug in a feature of sharp, unrelated to installation?

- [x] Running `npm install sharp` completes without error.
- [x] Running `node -e "require('sharp')"` completes without error.

### Are you using the latest version of sharp?

- [x] I am using the latest version of `sharp` as reported by `npm view sharp dist-tags.latest`.

### What is the output of running `npx envinfo --binaries --system --npmPackages=sharp --npmGlobalPackages=sharp`?

```
System:
    OS: Linux 5.11 Ubuntu 20.04.4 LTS (Focal Fossa)
    CPU: (2) x64 Intel(R) Xeon(R) Platinum 8171M CPU @ 2.[60](https://github.com/DoubbleDK/backend/runs/5741770630?check_suite_focus=true#step:7:60)GHz
    Memory: 5.74 GB / 6.78 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
[...]

PR #3692, opened 2023-06-07
fix: updated worker threads docs (per #3164)

titanism:


lovell:
Hi, thanks for the PR. Given this will affect the use of `ava` with Node.js native modules in general, perhaps the [ava docs](https://github.com/avajs/ava/blob/main/docs/08-common-pitfalls.md) might be a more appropriate place for this?

lovell:
Commit https://github.com/lovell/sharp/commit/4a37a27cca34782a213e4ac7773299476a6c294b prevents `dlclose` from unloading the sharp binary on Linux and will therefore remove the Worker Thread problem entirely. This will be in v0.33.0. Thanks again for taking the time to submit a PR to improve sharp.

PR #3735, opened 2023-07-21
Docs: use SVG image for the resize fit property example

kleisauke:
The placeholder image was based on [this image](https://pixabay.com/vectors/mountain-flat-design-design-nature-5332139/), which is licensed under the [Pixabay License](https://pixabay.com/service/license/).

Screenshot:
Screenshot_2023-07-21_171853


lovell:
Great, thanks Kleis.

kleisauke:
Ah, I just noticed that this doesn't blend well with the dark mode (which is automatically utilized based on the `prefers-color-scheme` media feature). That's probably why the previous image had an opaque background and `#ED1C24` as the text color.

lovell:
Good spot, I've added a background via commit https://github.com/lovell/sharp/commit/9a66e25f53b033e150d9c547e02fd22aeec4ecf4

Issue #3747, opened 2023-08-02
TypeScript: webp options type missing preset


pilotso11:
### Is this a possible bug in a feature of sharp, unrelated to installation?

- [ x ] Running `npm install sharp` completes without error.
- [ x ] Running `node -e "require('sharp')"` completes without error.

If you cannot confirm both of these, please open an [installation issue](https://github.com/lovell/sharp/issues/new?labels=installation&template=installation.md) instead.

### Are you using the latest version of sharp?

- [ x] I am using the latest version of `sharp` as reported by `npm view sharp dist-tags.latest`.

If you cannot confirm this, please upgrade to the latest version and try again before opening an issue.

If you are using another package which depends on a version of `sharp` that is not the latest, please open an issue against that package instead.

### What is the output of running `npx envinfo --binaries --system --npmPackages=sharp --npmGlobalPackages=sharp`?

  System:
    OS: macOS 13.4.1
    CPU: (12) arm64 Apple M2 Pro
    Memory: 38.89 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.13.0 - /usr/local/bin/node
    npm: 9.3.1 - /usr/local/bin/npm
    pnpm: 8.6.11 - /opt/homebrew/bin/pnpm

### What are the steps to reproduce?

```typescript
sharp(src).webp({preset: 'photo').output(dest)
```
Results in a warning 
```
TS2345: Argument of type '{ quality: number; preset: string; }' is not assignable to parameter of type 'WebpOptions'.   Object literal may only specify known properties, and 'preset' does not exist in type 'WebpOptions'.
```

### What is the expected behaviour?

`preset` should be accepted.  it is missing on the type

### Please provide a minimal, standalone code sample, without other dependencies, that demonstrates this problem

```typescript
sharp(src).webp({preset: 'photo').output(dest)
```

### Please provide sample image(s) that help explain this problem

Any image, this is just a problem with the types.

I will submit a pull request shortly.

lovell:
v0.32.5 now available, thanks for the PR.

PR #3748, opened 2023-08-02
Added webp preset types and test for same. Issue 3747

pilotso11:
For issue https://github.com/lovell/sharp/issues/3747.

The webp output options type is missing the preset values in the types.  They are present in the js library.
This PR adds:
* A new enum  (`PresetEnum`)
* The preset on `WebpOptions`
* The enum as `sharp.preset` on `sharp`.
* Tests for both string and enum based use of `webp({preset: ...})`

This follows the same implementation as other options, such as resize.fit.

npm test-types and npm test-lint both run cleanly.

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -146,6 +146,7 @@ declare namespace sharp {
     const kernel: KernelEnum;
     const fit: FitEnum;
     const bool: BoolEnum;
+    const preset: PresetEnum
 
     interface Sharp extends Duplex {
         //#region Channel functions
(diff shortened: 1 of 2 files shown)


lib/index.d.ts
```diff
@@ -146,6 +146,7 @@ declare namespace sharp {
     const kernel: KernelEnum;
     const fit: FitEnum;
     const bool: BoolEnum;
+    const preset: PresetEnum
```
lovell, line 149: How about `WebPPresetEnum` (or similar, the capitalisation looks a bit odd) to clarify this applies only to WebP images?

lovell (commented):
Thank you very much for the PR, I've left one question inline.

lib/index.d.ts
pilotso11: There don't seem to be any other format-specific enums that I can spot, but as a model yes that makes sense.  What if there were someday JpegXL-presets ... 

The WebP options are named `WebpOptions` so I'll rename this to `WebpPresetEnum` which also looks a little better.

pilotso11: This update has been pushed (the rename).  Ready for review again.

lovell:
> The enum as `sharp.preset` on `sharp`.

Did you intend to include this as part of your PR?

lovell (approved):

lovell:
Thank you Seth!

PR #3758, opened 2023-08-09
fix Type bug in WebpOptions.

sho-xizz:
While use sharp to convert to WebP with TypeScript, I found the Type Error of `WebpOptions.minSize`.
it param is boolean, but Type specify as number.
So we had going to use `@ts-tsgnore`  when use this option.

This PR corrects this bug.

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1124,7 +1124,7 @@ declare namespace sharp {
         /** Level of CPU effort to reduce file size, integer 0-6 (optional, default 4) */
         effort?: number | undefined;
         /** Prevent use of animation key frames to minimise file size (slow) (optional, default false) */
-        minSize?: number;
+        minSize?: boolean;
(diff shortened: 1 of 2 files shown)


lovell:
Thank you, this looks like a bug inherited from when the TypeScript definitions were imported from DefinitelyTyped. Are you able to add a [test](https://github.com/lovell/sharp/blob/main/test/types/sharp.test-d.ts) for this also, to help prevent regression?

sho-xizz:
Already there is minSize test as `Support `minSize` and `mixed` webp options`, but it looks like wrong.
So, I fix it param to boolean from number.

If this change is bad commit, could you tell me an example in one of your test cases I can base myself on to respect your guidelines, if any...?

lovell (approved):
Brilliant, thank you, the existing (inherited) test case was wrong so thanks for fixing this too!

Issue #3756, opened 2023-08-08
Pipeline chain order not respected when rotate is used after resize with fit mode "inside"


jozsefsallai:
### Is this a possible bug in a feature of sharp, unrelated to installation?

- [x] Running `npm install sharp` completes without error.
- [x] Running `node -e "require('sharp')"` completes without error.

If you cannot confirm both of these, please open an [installation issue](https://github.com/lovell/sharp/issues/new?labels=installation&template=installation.md) instead.

### Are you using the latest version of sharp?

- [x] I am using the latest version of `sharp` as reported by `npm view sharp dist-tags.latest`.

If you cannot confirm this, please upgrade to the latest version and try again before opening an issue.

If you are using another package which depends on a version of `sharp` that is not the latest, please open an issue against that package instead.

### What is the output of running `npx envinfo --binaries --system --npmPackages=sharp --npmGlobalPackages=sharp`?

```
  System:
    OS: Linux 5.15 Arch Linux
    CPU: (12) x64 AMD Ryzen 5 5600H with Radeon Graphics
    Memory: 13.00 GB / 15.48 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
    Node: 18.16.0 - ~/.asdf/installs/nodejs/18.16.0/bin/node
    Yarn: 1.22.19 - ~/.asdf/installs/nodejs/18.16.0/bin/yarn
    npm: 9.5.1 - ~/.asdf/plugins/nodejs/shims/npm
  npmPackages:
    sharp: ^0.32.4 => 0.32.4
```

### What are the steps to reproduce?

Creating an image manipulation pipeline which first resizes the image to a set width of 400 pixels and then rotates it by 90 degrees. The [example code](https://sharp.pixelplumbing.com/api-operation#rotate) in the documentation works correctly, but in this scenario, the `fit` property of the `resize` pipeline is set to `fill`. However, in my use case, this property needs to be set to `inside`.

### What is the expected behaviour?

`rotate -> resize` should work consistently across both `fill` and `inside` fit modes and output images that have a height of 400 pixels. With the `fit: "inside"` property, the image does not have a height of 400 pixels after rotating it by 90 degrees. Instead, it retains the original image dimensions that were resulted by the resize operation. In other words, the order of the chained operations doesn't matter if the fit property is set to `"inside"`.

[...]

PR #3762, opened 2023-08-14
Simplify 90/270 orient-before-resize logic

kleisauke:
See: #3756.


lovell:
So much better, thank you Kleis.

PR #3799, opened 2023-09-18
add block and unblock to typescript declaration

ldrick:
As I do not know all the possible `options.operations` I left it string[], if you know better, it would be cool to have the valid entries as List of Strings in the declaration.


lovell:
Thank you / danke schön!

Issue #3822, opened 2023-10-11
withMetadata TypeScript definition is missing boolean type


uhthomas:
### Is this a possible bug in a feature of sharp, unrelated to installation?

- [x] Running `npm install sharp` completes without error.
- [x] Running `node -e "require('sharp')"` completes without error.

If you cannot confirm both of these, please open an [installation issue](https://github.com/lovell/sharp/issues/new?labels=installation&template=installation.md) instead.

### Are you using the latest version of sharp?

- [x] I am using the latest version of `sharp` as reported by `npm view sharp dist-tags.latest`.

If you cannot confirm this, please upgrade to the latest version and try again before opening an issue.

If you are using another package which depends on a version of `sharp` that is not the latest, please open an issue against that package instead.

### What are the steps to reproduce?

The [withMetadata](https://github.com/lovell/sharp/blob/eefaa998725cf345227d94b40615e090495c6d09/lib/output.js#L212) implementation checks for a boolean and therefore implies it is valid to provide one as an argument. The [TypeScript definition](https://github.com/lovell/sharp/blob/eefaa998725cf345227d94b40615e090495c6d09/lib/index.d.ts#L649) however does not indicate this.

```sh

src/infra/repositories/media.repository.ts:35:21 - error TS2559: Type 'false' has no properties in common with type 'WriteableMetadata'.
```

### What is the expected behaviour?

```ts
withMetadata(withMetadata?: boolean | WriteableMetadata): Sharp;
```

lovell:
It looks like the work for https://github.com/lovell/sharp/issues/3824 may supersede this fix as parts of the `withMetadata()` API are being deprecated and most scenarios should switch to use the (proposed) `keepX()` methods instead.

lovell:
v0.33.0 is now available.

PR #3823, opened 2023-10-11
fix: allow boolean withMetadata argument

uhthomas:
Fixes: #3822


lovell (changes requested):
Thank you very much for reporting and fixing this. Please can you [add a test](https://github.com/lovell/sharp/blob/main/test/types/sharp.test-d.ts) to help avoid regression.

uhthomas:
@lovell Done.

lovell:
Thank you, I'm not planning any v0.32.x patch releases so I've merged this for inclusion in v0.33.0.

PR #3848, opened 2023-11-10
Correctly check for when sharp is unavailable

pi0:
Find while investigating issues with https://github.com/nuxt/image/pull/1102

`module.exports` defaults to an empty object `{}`, if no sources are available, we were not detecting it. 

PS: Let me know if you prefer another style for changes.


lovell:
Great spot, thanks Pooya, the curse of "istanbul ignore next" strikes again ;)

PR #3853, opened 2023-11-14
Mention JPEG-XL in readme

sandstrom:
https://jpegxl.info/


lovell:
Hi, thanks for the PR, this change makes sense only if prebuilt binaries contain support for JPEG-XL. I'm unsure if/when this will be, so I'll close for now.

Issue #3824, opened 2023-10-11
Enhancement: take advantage of libvips 8.15.0 keep metadata feature


uhthomas:
### Is this a possible bug in a feature of sharp, unrelated to installation?

- [x] Running `npm install sharp` completes without error.
- [x] Running `node -e "require('sharp')"` completes without error.

If you cannot confirm both of these, please open an [installation issue](https://github.com/lovell/sharp/issues/new?labels=installation&template=installation.md) instead.

### Are you using the latest version of sharp?

- [x] I am using the latest version of `sharp` as reported by `npm view sharp dist-tags.latest`.

If you cannot confirm this, please upgrade to the latest version and try again before opening an issue.

If you are using another package which depends on a version of `sharp` that is not the latest, please open an issue against that package instead.

### What are the steps to reproduce?

We're generating thumbnails without metadata for security, but we want to apply an ICC profile to avoid dull colours. It does not seem possible to do this without first creating a thumbnail without any metadata, and then applying the ICC profile.

### What is the expected behaviour?

Apply an ICC profile and strip all other metadata in one operation.

### Please provide a minimal, standalone code sample, without other dependencies, that demonstrates this problem

https://github.com/immich-app/immich/pull/4438

lovell:
The forthcoming libvips 8.15.x will invert the remove vs keep metadata logic via https://github.com/libvips/libvips/pull/3585 and, as a result, I'm planning to revisit and rework the current `withMetadata()` API in sharp as part of integrating it.

I've not had the chance to create an issue for this yet, so if it's OK with you, let's use this to track it.

Note to self: see also https://github.com/lovell/sharp/issues/2910

etnoy:
Seems like libvips 8.15 is getting closer, rc1 was just released

adriaanmeuris:
I'm sharing my use case as it may be relevant to upcoming changes in `withMetadata`:

1. we convert 2 jpg files to CMYK colorspace with `sharp`, using `withMetadata` in combination with `icc: 'U.S. Web Coated (SWOP) v2'`. The resulting files each contain the ICC profile in the metadata, adding about 500kb to each file:
    ```js
    await sharp('input.jpg')
      .toColourspace('cmyk')
      .withMetadata({ icc: 'U.S. Web Coated (SWOP) v2.icc' })
      .toFile('output.jpg');
    ```
[...]

Issue #2910, opened 2021-09-30
Option to replace EXIF metadata


stackcoder:
**What are you trying to achieve?**
Discard all existing metadata of an image and only write entries specified within `withMetadata` to the file. The current behavior merges existing entries with the new ones.

**Have you searched for similar questions?**
Sure at least: #650, #2724, #2812, #2638, #2507

The `strip` [tag of libvips](https://github.com/libvips/libvips/blob/608a7cee9b2d00e903d5767915e8bfb3fa75e90c/cplusplus/include/vips/VImage8.h#L3601) is `!baton->withMetadata` [controlled](https://github.com/lovell/sharp/blob/f37ca8249afcd26777b7e3e1d0192640572341ef/src/pipeline.cc#L956) by `!baton->withMetadata`. So might be difficult to combine `strip` and updating metadata together within libvips. Is there any other option instead encoding the image twice?

```javascript
// My current workaround:
sharp.prototype.discardMetadata = async function () {
  const {data, info} = await this
    .raw()
    .toBuffer({resolveWithObject: true});
  return sharp(data, {raw: info});
};
```

**Are you able to provide a minimal, standalone code sample that demonstrates this question?**
Already exists, there you go: https://sharp.pixelplumbing.com/api-output#withmetadata

**Are you able to provide a sample image that helps explain the question?**
Not required.

lovell:
This is being covered as part of the work integrating the new "keep" metadata feature of libvips - see https://github.com/lovell/sharp/issues/3824#issuecomment-1809148065

lovell:
Please subscribe to #3824 for updates on the provision of a new `withExif()` function to replace all EXIF metadata.

PR #3856, opened 2023-11-16
Take advantage of libvips new "keep" feature to increase control over output metadata

lovell:
libvips v8.15.0 deprecated the "strip" option for removing all metadata, instead replacing it with a set of more fine-grained "keep" flags.

This PR adds new functions to take advantage of this:

- keepExif
- withExif
- withExifMerge
- keepIccProfile
- withIccProfile
- keepMetadata

EXIF can now be either replaced or merged, plus there are new TypeScript definitions for EXIF metadata structure.

The input ICC profile can now be attached to the output image.

The `exif` and `icc` properties of `withMetadata` are now deprecated. The ability to provide a boolean to `withMetadata` has been removed.

Relates to #3824 and #2910


adriaanmeuris:
I'm thrilled to see the progress on this! After reviewing it, I wanted to share some thoughts.

Using `withIccProfile` is quite powerful, handling both the transformation of image data and embedding that profile into the output file. Reflecting on our previous conversation in https://github.com/lovell/sharp/issues/3824#issuecomment-1809148065, I understand that the default behaviour, when no functions are used, aligns with `libvips`' `keep=none` approach.

Here's where I see a potential enhancement. With `libvips` we can convert the color space and strip all metadata, including ICC profiles. For instance:
```sh
vips icc_transform input_cmyk_fogra39_with_icc.jpg "output_cmyk_swop_stripped.jpg[keep=none]" "U.S. Web Coated (SWOP) v2.icc"
```

In the current implementation of this PR, achieving this seems not possible because using any of the new functions inadvertently adds an ICC profile to the output file, so its behaviour does not entirely match what `keep=none` does in the example above.

Recognizing that the current approach might be tailored to more common use cases, I wonder if it's possible to introduce an optional boolean flag to `keepMetadata`? This addition could offer a way to replicate `libvips` behaviour.  

Alternatively, integrating the same metadata flags as in `libvips` [as discussed here](https://github.com/libvips/libvips/pull/3585#issuecomment-1742025733) might provide an even finer level of control. Of course, this could mean a bit more work:
```
 * @VIPS_FOREIGN_KEEP_NONE: don't attach metadata
 * @VIPS_FOREIGN_KEEP_EXIF: keep Exif metadata
 * @VIPS_FOREIGN_KEEP_XMP: keep XMP metadata
 * @VIPS_FOREIGN_KEEP_IPTC: keep IPTC metadata
 * @VIPS_FOREIGN_KEEP_ICC: keep ICC metadata
 * @VIPS_FOREIGN_KEEP_OTHER: keep other metadata (e.g. PNG comments and some TIFF tags)
 * @VIPS_FOREIGN_KEEP_ALL: keep all metadata
```

Would love to hear your thoughts on this.

lovell:
@adriaanmeuris I'd like to avoid an API that allows people to create invalid or incorrectly-rendered images. An image that uses a profile should attach that profile (the exception perhaps being device independent profiles such as sRGB). I'm aware your use case is for PDFs, but sharp isn't really designed for PDF manipulation.

adriaanmeuris:
@lovell I completely share your concern about preventing the creation of incorrectly rendered images with sharp, and it's not my intention to compromise this aspect of the library.

I do want to point out that my use of sharp is not for PDF manipulation, but rather for preprocessing images for a target that is not a browser. Unlike browsers, where stripping out the sRGB profile by default is standard due to browsers assuming untagged images are in sRGB, other targets can offer more precise control over color consistency.

For ICC-based images in PDFs, the format refers to a separate ICC buffer (the one available in sharp's `metadata` function) rather than embedding the profile in the image. I completely agree it's not the most common scenario, but it does mirror the rationale behind stripping out the sRGB profile by default, while recognizing that not all use cases target a browser environment.

Considering this, would you be open to a separate function such as `removeIccProfile`? Being a distinct function ensures intentional use and allows for clear documentation of its purpose. This should address the concern about accidentally creating incorrectly-rendered images.

I'd certainly be open to creating a PR to make this possible.

lovell:
@adriaanmeuris The new `withIccProfile()` function now accepts an optional `attach` parameter, defaulting to `true`, so you could use:
```js
.withIccProfile('profile.icc', { attach: false })
```

adriaanmeuris:
Thanks, @lovell! I've tested the new `withIccProfile()` with `attach` parameter and it works exactly as expected. Really appreciate you considering and implementing this.

Issue #3338, opened 2022-08-18
ImageMagick-style tint


shahkashani:
## Question about an existing feature

I'm pretty sure this is just me being stupid, but is there a way to use `tint` to achieve something like the below?

### ImageMagick

```convert "input.png" -fill "#c31306" -tint 100 "output.png"```

![output-im](https://user-images.githubusercontent.com/1395893/185453564-fcc08fa8-0443-4428-bc27-a56754324b56.jpg)

### Sharp

```await sharp(image).tint('#c31306')```

![output-sharp](https://user-images.githubusercontent.com/1395893/185453697-33d27e75-6e12-4557-a12c-988356250021.png)

Note how the highs and lows are also tinted in the Sharp case, but not with ImageMagick. 

The [ImageMagick source code](https://github.com/ImageMagick/ImageMagick/blob/master/MagickCore/visual-effects.c#L2972) indicates that they use a weighting function for their tinting:

```
%  TintImage() applies a color vector to each pixel in the image.  The length
%  of the vector is 0 for black and white and at its maximum for the midtones.
%  The vector weighting function is f(x)=(1-(4.0*((x-0.5)*(x-0.5))))
```

But I am not sure how to convert this into Sharp terms. Perhaps I could use `recomb` somehow?

I have even tried to use `composite`, but none of the blend modes seem to be able to achieve what I want. `multiply` gets close, but I think ultimately what I need is something like Cairo's `color` blend mode, which doesn't exist in Sharp or vips.

Any help would be highly appreciated!

### What are you trying to achieve?

A tint that preserves hights and lows.

### When you searched for similar issues, what did you find that might be related?

I found some tickets regarding tint not behaving as expected, but those issues have been resolved.

### Please provide a minimal, standalone code sample, without other dependencies, that demonstrates this question

```
const image = await sharp('input.png').toBuffer();
const outputBuffer = await sharp(image).tint('#c31306');
await outputBuffer.toFile('output.png');
```

### Please provide sample image(s) that help explain this question

Raw image:

![sandwich](https://user-images.githubusercontent.com/1395893/185454088-9c627d29-c4b4-4d3d-9ce5-0fa4666bb566.png)

lovell:
[...]

PR #3859, opened 2023-11-18
Improve tint luminance with weighting function

lovell:
This PR applies ~~the same~~ a similar [weighting function](https://github.com/ImageMagick/ImageMagick/blob/6d5984e1aa9839b9cac3818ed99e24e1a76f6461/MagickCore/visual-effects.c#L2985) as ImageMagick to help preserve dark and light areas when using the `tint` operation.

See https://github.com/lovell/sharp/issues/3338 for context.


jcupitt:
Nice! Though it's not quite the same weighting function as IM, it should be a little better

PR #3864, opened 2023-11-21
use $NODE rather than node for install script

koush:
when the runtime used to npm install is not node (ie, electron), the `node` binary will not be in path or available. use the $NODE env instead, as that is provided by npm/node/electron.


koush:
this is a minimal change as my use case is only for installing the prebuilt binaries. You may want to adjust all invocations of `node` accordingly in the package.json scripts.

Issue #3893, opened 2023-12-11
Some images are only displayed in black & white, old Intel (before 2009) and QEMU-emulated CPUs


Zharkan:
## Possible install-time or require-time problem

- [x] I have read the [documentation relating to installation](https://sharp.pixelplumbing.com/install).

### Are you using the latest version of sharp?

- [x] I am using the latest version of `sharp` as reported by `npm view sharp dist-tags.latest`.

### Are you using a supported runtime?

- [x] I am using Node.js with a version that satisfies `^18.17.0 || ^20.3.0 || >=21.0.0`

### Are you using a supported package manager and installing optional dependencies?

- [x] I am using npm >= 9.6.5 with `--include=optional`

### What is the complete output of running `npm install --verbose --foreground-scripts sharp` in an empty directory?

<details>

```
npm info using npm@10.2.5
npm info using node@v20.10.0
npm verb title npm install sharp
npm verb argv "install" "--loglevel" "verbose" "--foreground-scripts" "sharp"
npm verb logfile logs-max:10 dir:/home/media_gestion/.npm/_logs/2023-12-11T23_18_28_144Z-
npm verb logfile /home/media_gestion/.npm/_logs/2023-12-11T23_18_28_144Z-debug-0.log
npm http fetch GET 200 https://registry.npmjs.org/sharp 353ms (cache revalidated)
npm verb reify failed optional dependency /path/to/project/node_modules/fsevents
npm verb reify failed optional dependency /path/to/project/node_modules/@next/swc-win32-x64-msvc
npm verb reify failed optional dependency /path/to/project/node_modules/@next/swc-win32-ia32-msvc
npm verb reify failed optional dependency /path/to/project/node_modules/@next/swc-win32-arm64-msvc
npm verb reify failed optional dependency /path/to/project/node_modules/@next/swc-linux-arm64-musl
npm verb reify failed optional dependency /path/to/project/node_modules/@next/swc-linux-arm64-gnu
npm verb reify failed optional dependency /path/to/project/node_modules/@next/swc-darwin-x64
npm verb reify failed optional dependency /path/to/project/node_modules/@next/swc-darwin-arm64
npm verb reify failed optional dependency /path/to/project/node_modules/@img/sharp-win32-x64
npm verb reify failed optional dependency /path/to/project/node_modules/@img/sharp-win32-ia32
npm verb reify failed optional dependency /path/to/project/node_modules/@img/sharp-wasm32
[...]

PR #3897, opened 2023-12-15
Fixed VIPS_NOVECTOR correct work

icetee:
Based on the #3893 issue, there was a setting problem, which is fixed with this modification. If the `process.env.VIPS_NOVECTOR=0`, then simd set `false` in default.

### DIFF
--- a/lib/utility.js
+++ b/lib/utility.js
@@ -201,7 +201,6 @@ function counters () {
 function simd (simd) {
   return sharp.simd(is.bool(simd) ? simd : null);
 }
-simd([0, "0", "false"].includes(process.env.VIPS_NOVECTOR) ? false : true);
 
 /**
  * Block libvips operations at runtime.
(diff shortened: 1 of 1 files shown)


lib/utility.js
```diff
@@ -201,7 +201,7 @@ function counters () {
 function simd (simd) {
   return sharp.simd(is.bool(simd) ? simd : null);
 }
-simd(true);
+simd([0, "0", "false"].includes(process.env.VIPS_NOVECTOR) ? false : true);
```
lovell, line 204: libvips checks for the presence rather than value of this env var, so we should probably do likewise (could use `is.defined()` to check?).

https://github.com/libvips/libvips/blob/e95a169a3c66e36c5062736f3c622aa06f924d02/libvips/iofuncs/vector.cpp#L89

lovell (commented):
Thank you for the PR, I've left one comment inline.

kleisauke:
I would suggest to simply remove this line, given that libvips already utilizes SIMD by default (when compiled against either Highway or liborc).

icetee:
I deleted the line, utilities.cc unlocks it and checks it. Its work.

lovell:
Kösz Tamás!

Issue #3908, opened 2023-12-20
`Index.d.ts` missing prompt for `keepmetadata`


rambo-panda:
![image](https://github.com/lovell/sharp/assets/2776050/d5bdc4b3-9037-4eb0-b71e-aca7a6913cff)

https://github.com/lovell/sharp/blob/main/lib/index.d.ts#L680C63-L680C63

lovell:
Thanks for reporting, looks like this was missed as part of https://github.com/lovell/sharp/pull/3856 - happy to accept a PR if you're able.

PR #3914, opened 2023-12-23
Added typescript prompt for keepmetadata

abhi0498:
This PR adds keepMetaData type to Sharp instance.

See #3908 for more context.


lovell:
Thank you for the PR, please can you add a test case to [sharp.test-d.ts](https://github.com/lovell/sharp/blob/main/test/types/sharp.test-d.ts) for this to help prevent regression.

lovell:
Thank you!

PR #3937, opened 2024-01-10
docs(install): Write full supported versions

Princesseuh:
Hello!

I'm not too picky on the exact format, so feel free to reformat in a way you prefer, but I found it confusing how the "semver" version mention in the docs wasn't accurate, `>= 18.17.0` implies that any versions after `18.17.0` works, but that is not the case. For instance, someone on `20.2.x` will fail to install Sharp, as will someone on `19.x.x`. 

This caused a few downstream issues in repos I maintain that have Sharp as an `optionalDependencies` because users didn't get an error on install (since it's an optionalDep), but as far as they could tell from the Sharp docs, they had a supported version of Node.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -35,7 +35,7 @@ deno run --allow-ffi ...
 
 ## Prerequisites
 
-- Node-API v9 compatible runtime e.g. Node.js ^18.17.0 or >=20.3.0.
+* Node-API v9 compatible runtime e.g. Node.js ^18.17.0 or >=20.3.0.
 
 ## Prebuilt binaries
 
--- a/docs/install.md
+++ b/docs/install.md
@@ -35,7 +35,7 @@ deno run --allow-ffi ...
 
 ## Prerequisites
 
-* Node-API v9 compatible runtime e.g. Node.js >= 18.17.0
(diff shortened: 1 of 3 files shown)


docs/install.md
```diff
@@ -35,7 +35,7 @@ deno run --allow-ffi ...
 
 ## Prerequisites
 
-* Node-API v9 compatible runtime e.g. Node.js >= 18.17.0
+- Node-API v9 compatible runtime e.g. Node.js ^18.17.0 or >=20.3.0.
```
Princesseuh, line 38:
```suggestion
* Node-API v9 compatible runtime e.g. Node.js ^18.17.0 or >=20.3.0.
```

lovell:
Brilliant, thank you very much for this Erika.

lovell:
I also added a runtime check via commit https://github.com/lovell/sharp/commit/45e8071599772f320d0d8314fedf0b13b0592d28

Princesseuh:
Awesome, thank you! I unfortunately had to pin the Sharp version used on my projects to `0.32.6` for now, since I still support Node `>=18.14.1`, but will be updating to Sharp 0.33.0 for sure once we make our semver checks more strict.

Thank you very much for your work on Sharp, you deserve way more help from the ecosystem.

Issue #3936, opened 2024-01-10
Missing Pre-Built win32-arm64


Stoyvo:
### Is this a possible bug in a feature of sharp, unrelated to installation?

- [ x ] Running `npm install sharp` completes without error.
- [ x ] Running `node -e "require('sharp')"` completes without error.

### Are you using the latest version of sharp?

- [ x ] I am using the latest version of `sharp` as reported by `npm view sharp dist-tags.latest`.

### What is the output of running `npx envinfo --binaries --system --npmPackages=sharp --npmGlobalPackages=sharp`?

Not Relevant.

### What are the steps to reproduce?

Look at GitHub Release assets for this repository. If win32-arm64 is missing on latest release.

### What is the expected behaviour?

A pre-built win32-arm64 asset.

PR #3938, opened 2024-01-10
Arm64 support

Stoyvo:
Add support for MacOS Arm64 and Windows Arm64

Fixes: https://github.com/lovell/sharp/issues/3936

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,17 @@ jobs:
             nodejs_version: "^20.3.0"
             nodejs_version_major: 20
             platform: darwin-x64
+          - os: macos-13-arm
+            nodejs_arch: arn64
+            nodejs_version: "^18.17.0"
+            nodejs_version_major: 18
+            platform: darwin-arn64
+            prebuild: true
+          - os: macos-13-arm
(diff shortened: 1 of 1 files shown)


lovell:
Hi, thanks for the PR. Prebuilt ARM64 binaries are already provided for macOS and Linux. Unfortunately there are no CI providers that currently provide support for ARM64 on Windows. If you know of one, please let me know.

I'll close this PR as it won't work on the existing CI setup.

tstoyanovski-humi:
@lovell are you using github runners?
https://github.com/actions/runner/pull/2022

lovell:
@tstoyanovski-humi The issue you've linked to relates to software support. What we need is someone offering hosted hardware, for free, ideally GitHub Actions itself as that's what we currently use for Windows on amd64.

jhholm:
Not sure if this is the correct place, but it seems that ARM hosted runners are finally available in GitHub

https://blogs.windows.com/windowsdeveloper/2025/04/14/github-actions-now-supports-windows-on-arm-runners-for-all-public-repos/

lovell:
For Windows ARM64 support please see https://github.com/lovell/sharp-libvips/issues/238

PR #3962, opened 2024-01-19
chore: Fix typo s/introducted/introduced/

hugo-syn:
Fix typo s/introducted/introduced/

(Forgot to add it on my previous PR)


lovell:
Thanks, this typo was inherited from DefinitelyTyped so I can't take credit for it ;)

PR #3982, opened 2024-02-06
[doc] Added information about using sharp with electron forge + vite.

idchlife:
+ Slight titles and format changes in electron section.

I stumbled upon common issue with using sharp + electron. In my case it was electron forge + vite.
After searching for an answer for some time I discovered the resolution.

I think it would be nice for electron users to know how they can avoid common errors via new option configuration.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -233,6 +233,8 @@ custom:
 
 ### electron
 
+#### electron-builder
+
 Ensure `sharp` is unpacked from the ASAR archive file using the
 [asarUnpack](https://www.electron.build/configuration/configuration.html)
 option.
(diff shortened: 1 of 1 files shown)


guanzhongxu:
I recently encountered the same problem. I tried many methods but ended up having to downgrade the version of Sharp and modify the path of the .node file introduced in the Sharp source code. Only then could I get the program to run properly. Later, I was surprised to find your submission. I tried to modify it according to your method, but it still didn't work. It's possible that there's an issue with my approach. Below is my configuration code.
package.json:
image
electron-builder.json5:
image
vite.config.ts
image
error message：
image

idchlife:
@guanzhongxu my addition is to electron forge + vite. You are using electron builder.

In your situation (I checked your repro repo), please add vite rollup config into the electron.main.vite config, like this:

```ts
export default defineConfig({
	optimizeDeps: {
		exclude: ["sharp"]
	},
	plugins: [
		common({}),
		vue(),
		Components({
			resolvers: [NaiveUiResolver()]
		}),
		AutoImport({
			imports: [
				"vue",
				"vue-router",
				{
					"naive-ui": ["useDialog", "useMessage", "useNotification", "useLoadingBar"]
				}
			],
			dts: "./src/autoImport.d.ts"
		}),
		electron({
			main: {
				entry: "electron/main.ts",
                                // BELOW THE PART YOU NEED
				vite: {
					build: {
						rollupOptions: {
							external: ["sharp"]
						}
					}
				}
			},
```



Now your project starts with no problems.

I used `npm` to install dependencies, though. Don't know how it will work with `pnpm`

guanzhongxu:
I have indeed run it according to your way, thank you very much for your guidance.

lovell:
Thank you. I've moved this to a generic Vite section as this information will help all of its users, not just those using Electron. Landed as https://github.com/lovell/sharp/commit/fc439bedf1d4bdfc9462031ccf22231008d33e75

PR #4036, opened 2024-03-20
feat: bitdepth option for avif/heif

### LINKED ISSUE #4031: Support outputting higher bit-depth AVIF

As a photo library, we convert and sanitize images to WebP and JPEG for performance. But these formats can compromise on color fidelity since they're limited to 8 bits. As a result, it's not uncommon for users to want to avoid using the converted images altogether.

AVIF, unlike these formats, allows for higher bit depths of 10 and 12. Combined with its higher compression, this helps bridge the gap between formats like WebP and the often very large originals.

AVIF also supports HDR, but this is out of scope for this request since it likely requires upstream changes in libvips.

#3358 changed AVIF to always have 8-bit output for compatibility with AOM builds. Defaulting to 8 bits would still maintain this behavior, but allow for higher bit depths when the API consumer knows their build supports this.

Similar to the TIFF output options, a `bitdepth` setting that defaults to 8 and can be configured to 10 or 12.

P3 helps, but ideally it'd be complementary to this rather than an alternative.

Ultra HDR is a way to embed gain maps into JPEG. However, it's still limited in color fidelity and doesn't benefit from AVIF's better compression. libvips also doesn't currently support Ultra HDR, so it'd be quite a bit more work.

mertalev:
Following #4031, this PR exposes a `bitdepth` setting to allow outputting 10-bit or 12-bit AVIF/HEIF files. The default remains 8, so this doesn't change the current behavior.

Fixes #4031

### DIFF
--- a/lib/constructor.js
+++ b/lib/constructor.js
@@ -325,6 +325,7 @@ const Sharp = function (input, options) {
     heifCompression: 'av1',
     heifEffort: 4,
     heifChromaSubsampling: '4:4:4',
+    heifBitdepth: 8,
     jxlDistance: 1,
     jxlDecodingTier: 0,
     jxlEffort: 7,
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1244,6 +1244,8 @@ declare namespace sharp {
         effort?: number | undefined;
         /** set to '4:2:0' to use chroma subsampling, requires libvips v8.11.0 (optional, default '4:4:4') */
         chromaSubsampling?: string | undefined;
+        /** Set bitdepth to 8, 10 or 12 bit (optional, default 8) */
+        bitdepth?: 8 | 10 | 12 | undefined;
     }
 
     interface HeifOptions extends OutputOptions {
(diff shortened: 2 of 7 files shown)


lovell:
Thank you for the PR. The prebuilt binaries only include support for a bitdepth of 8, so you won't be able to add a unit test for this path, but you can check valid values are accepted.

It looks like lots of the tests are failing as the new `heifBitdepth` property needs adding somewhere around here in [constructor.js](https://github.com/lovell/sharp/blob/dc07fd4e9c293a9c62dc84d8997aaf8d2833cf3e/lib/constructor.js#L327).

mertalev:
Oh, right! I forgot the default build only supports 8-bit output.

mertalev:
I got the tests working locally after adding `--build-from-source` and fixing a few bugs. Bit depth as a boolean didn't work too well haha

lovell (approved):
Thanks for the updates, all the tests are passing now, this is good to merge.

(I'll also add someone to the docs to highlight that the prebuilt binaries only support a bitdepth of 8.)

lovell:
I added a guard against this property when using the prebuilt binaries via commit https://github.com/lovell/sharp/commit/aa1bbcb5c1b871c5d862c66f4cdf1c8dd985c1ed as the failure condition could otherwise lead to a segfault.

mertalev:
Yeah, that makes sense.

adityapatadia:
Just a question, is there a harm if we pass `bitdepth: 10` for all images? Will it end up causing bigger output size?

PR #4048, opened 2024-04-01
types: added missing types for OverlayOptions

ike-gg:
`animated` property was missing from declaration types, just added it


ike-gg:
added also: `density`, `failOn`, `limitInputPixels`

lovell (approved):

lovell:
Thank you very much for reporting and fixing this.

PR #4074, opened 2024-04-19
Simplify StaySequential operation

kleisauke:


lovell:
Thank you Kleis, I totally missed `vips_image_is_sequential`, that's exactly what this feature needs.

PR #4088, opened 2024-04-26
Add missing angle bracket

skyf0l:


skyf0l:
Indeed, I accidentally saw this mistake on a big repo. Searching a bit on github, I realized it was quite common. So I decided to fix it on a few large repos.

Btw I agree with you that these e-mails are sensitive data. Although they're already in the Git history and the maintainers wrote them themselves. It's also possible that these emails are simply not currently being used by the authors. That's a question worth exploring.

[eslint-plugin-package-json](https://github.com/JoshuaKGoldberg/eslint-plugin-package-json) is a good catch, I'm going to dig it up and add a rule regarding the format of contributors' names/emails.

PR #4089, opened 2024-04-26
doc: fix ensuareAlpha alpha argument

gutenye:


lovell:
I'm unsure I understand this change as the current documentation is correct. Adding an alpha channel using a value of 0 produces a fully-transparent image.

The alpha channel relates to the ability to see whatever is behind the image. Perhaps you are thinking of a mask?

gutenye:
@lovell you're right, the output file must be `png`.

If I use `jpg` as output, like below, it outputs a black image, what caused it? This gets me confused.

```
await sharp('rgb.jpg')
  .ensureAlpha(0)
  .toFile('output.jpg')
```

lovell:
JPEG images do not support an alpha channel.

Issue #4045, opened 2024-03-29
Output of profile conversion in combination with `negate` seems off


adriaanmeuris:
### Is this a possible bug in a feature of sharp, unrelated to installation?

- [x] Running `npm install sharp` completes without error.
- [x] Running `node -e "require('sharp')"` completes without error.

If you cannot confirm both of these, please open an [installation issue](https://github.com/lovell/sharp/issues/new?labels=installation&template=installation.md) instead.

### Are you using the latest version of sharp?

- [x] I am using the latest version of `sharp` as reported by `npm view sharp dist-tags.latest`.

If you cannot confirm this, please upgrade to the latest version and try again before opening an issue.

If you are using another package which depends on a version of `sharp` that is not the latest, please open an issue against that package instead.

### What is the output of running `npx envinfo --binaries --system --npmPackages=sharp --npmGlobalPackages=sharp`?

```
  System:
    OS: macOS 14.3.1
    CPU: (12) arm64 Apple M2 Pro
    Memory: 88.27 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 18.18.2 - ~/.nvm/versions/node/v18.18.2/bin/node
    Yarn: 1.22.21 - ~/.nvm/versions/node/v18.18.2/bin/yarn
    npm: 10.3.0 - ~/.nvm/versions/node/v18.18.2/bin/npm
    pnpm: 7.28.0 - ~/Library/pnpm/pnpm
  npmPackages:
    sharp: ^0.33.3 => 0.33.3 
```

### What are the steps to reproduce?

I need to process a CMYK TIFF file ([tiger-cmyk-fogra.tif](https://github.com/lovell/sharp/files/14801857/tiger-cmyk-fogra.tif.zip)) with an embedded Coated FOGRA39 profile. My goal is to convert this image to the `U.S. Web Coated (SWOP) v2` profile and invert the colors, ideally in a single operation. The following code performs these tasks in two steps and achieves the desired result:
```js
// Convert from FOGRA to SWOP and then invert in two steps
await sharp('tiger-cmyk-fogra.tif')
  .withIccProfile('U.S. Web Coated (SWOP) v2.icc')
  .toFile('tiger-cmyk-swop.tif');

await sharp('tiger-cmyk-swop.tif')
  .keepIccProfile()
  .negate()
  .toFile('tiger-cmyk-swop-inverted.tif');
```

[...]

PR #4096, opened 2024-05-08
Add tests for validating output colors when doing cmyk to cmyk conversion with embedded profiles

adriaanmeuris:
As discussed in https://github.com/lovell/sharp/issues/4045, this PR:
1. Adds a test to validate output colors when doing CMYK to CMYK colourspace transformations with embedded profiles using perceptual intent (the current default). This test might be updated in the future if the default intent changes to `relative` instead of `perceptual`, or if an optional parameter is provided to select the preferred intent for CMYK transformations.

2. Moves the `negate` operation to the bottom of the pipeline, preventing colourspace transformations to be applied to already negated image data resulting in output colors being incorrect.

3. Adds a test that validates (2) by asserting similarity with expected color values when negating in combination with a CMYK colourspace transformation.

Currently, all tests pass except for `negate (png, trans)`. This seems related to premultiplication of the alpha channel as commenting out the `resize` operation of the test results in the test passing.

@lovell I'm happy to implement any changes to ensure the test passes. Can you give me some pointers on how to tackle negating in combination with premultiplication of the alpha channel if present?


lovell:
Thanks for working on this Adriaan, always good to have more testing around CMYK logic.

I took a look at the failing test and I think the current expectation is wrong, and that your change produces the correct output, so please go ahead and update the `negate-trans.png` file as part of this PR.

Do you know the source of the `U.S. Web Coated (SWOP) v2.icc` file? I ask as it might be subject to copyright. If the ownership/licensing is unclear, perhaps one of the profiles in the Debian [icc-profiles](https://packages.debian.org/bookworm/all/icc-profiles/filelist) package could be used?

lovell (approved):
Thanks for the updates.

PR #4111, opened 2024-05-22
fix: return only boolean when useGlobalLibvips is used by gyp

project0:
The debug log is compromising the gyp variable use_global_libvips and therefore breaks compilation when the detection adds additional output.

```
/app/node_modules/sharp # SHARP_FORCE_GLOBAL_LIBVIPS=true node -p "Boolean(require('./lib/libvips').useGlobalLibvips()).toString()" 
sharp: Detected SHARP_FORCE_GLOBAL_LIBVIPS, skipping search for globally-installed libvips
true
```


lovell:
Thanks for the PR. Given the "skipping search for globally-installed libvips" log message is somewhat confusing when `SHARP_FORCE_GLOBAL_LIBVIPS` is set, could this be simplified to something like the following?

```diff
+++ b/lib/libvips.js
@@ -172,7 +172,7 @@ const useGlobalLibvips = () => {
     return skipSearch(false, 'SHARP_IGNORE_GLOBAL_LIBVIPS');
   }
   if (Boolean(process.env.SHARP_FORCE_GLOBAL_LIBVIPS) === true) {
-    return skipSearch(true, 'SHARP_FORCE_GLOBAL_LIBVIPS');
+    return true;
   }
   /* istanbul ignore next */
   if (isRosetta()) {
```

project0:
i guess it could, but at the same time it is a helpful information that the detection is enforced/influenced by a the env variable.

project0:
Also, giving the output is used in a script we already got lucky as we never checked for the `false` value.

PR #4123, opened 2024-06-04
Fix flip and flop test

happycollision:
It was only testing flop.

-----

I don't know if such a small change would make me a contributor in the package file, so I did that as a separate commit which can easily be dropped.


lovell:
Thank you for finding and fixing this test, which has been incorrect for over 9 years!

happycollision:
Heh. Hopefully I've earned enough goodwill for you to read [my lengthy proposal for a new feature](https://github.com/lovell/sharp/issues/4144).

Issue #4145, opened 2024-07-02
Enhancement: add support to recomb for 4x4 recombination matrices


ton11797:
### What are you trying to achieve?

I want to multiply the Alpha channel with an opacity value. This is what I came up with in my code

```
    async multiplyAlphaChannel(layerImage: Sharp, multiply: number) {
        const layerMetadata = await layerImage.metadata();
        const channels: Array<{ data: Buffer }> = await Promise.all([
            layerImage.extractChannel(0).raw().toBuffer({ resolveWithObject: true }),
            layerImage.extractChannel(1).raw().toBuffer({ resolveWithObject: true }),
            layerImage.extractChannel(2).raw().toBuffer({ resolveWithObject: true }),
            layerImage.extractChannel(3).raw().toBuffer({ resolveWithObject: true })
        ]);

        const pixelArray = new Uint8ClampedArray(channels[0].data.length * 4);

        for (let i = 0; i < channels[0].data.length; i++) {
            pixelArray[i * 4] = channels[0].data[i]; // Red
            pixelArray[i * 4 + 1] = channels[1].data[i]; // Green
            pixelArray[i * 4 + 2] = channels[2].data[i]; // Blue
            pixelArray[i * 4 + 3] = channels[3].data[i] * multiply; // Alpha
        }

        return await sharp(pixelArray, {
            raw: {
                width: layerMetadata.width,
                height: layerMetadata.height,
                channels: 4
            }
        }).png();
    }
```

### Question

Are there any more efficient or simplified methods to achieve this using the Sharp library?

lovell:
This is a scenario that matrix multiplication via the [recomb](https://sharp.pixelplumbing.com/api-operation#recomb) operation should allow but currently it only supports a 3x3 matrix for non-alpha channels.

Let's tag this as an enhancement for 4x4 matrix support.

```js
// Proposed API to support 4x4 matrix - NOT YET AVAILABLE
const alphaMultiplication = 1.1;
const output = await sharp(input)
  .recomb([
    1, 0, 0, 0,
    0, 1, 0, 0,
    0, 0, 1, 0,
    0, 0, 0, alphaMultiplication
  ])
  .toBuffer();
```

lovell:
v0.33.5 now available, thank you for implementing this feature.

PR #4147, opened 2024-07-02
Enhance recombination algorithm to support 4x4 matrices (#4145)

ton11797:
This pull request enhances the recombination algorithm to support 4x4 matrices. 

**Changes:**
- Modified the recombination function to handle 4x4 matrices.
- Added test cases to verify the correctness of the implementation for 4x4 matrices.

**Related Issues:**
- Issue #4145 

This is my first contribution to public project, so any feedback or suggestions for improvement would be greatly appreciated.


lib/operation.js
```diff
@@ -787,24 +787,28 @@ function linear (a, b) {
...
-    inputMatrix[2][0], inputMatrix[2][1], inputMatrix[2][2]
-  ].map(Number);
+
+  for (const row of inputMatrix) {
+    if (row.length !== length) {
+      throw new Error('Invalid recombination matrix');
+    }
+  }
+
+  this.options.recombMatrix = inputMatrix.flat().map(Number);
```
lovell, line 810: Excellent use of `flat()` here, I suspect it wasn't fully-available when this feature was first added. Perhaps this also provides a simpler way to validate the matrix length, i.e. must be either 9 or 16?

lib/index.d.ts
```diff
@@ -1730,6 +1730,7 @@ declare namespace sharp {
 
     type Matrix2x2 = [[number, number], [number, number]];
```
lovell, line 1731: Not sure `Matrix2x2` is used anywhere, I suspect this was inherited via DefinitelyTyped and could be removed.

lovell (commented):
Fantastic, thank you very much for working on this, I've left a couple of questions inline.

lib/index.d.ts
ton11797: this type are used at affine() //2x2 affine transformation matrix

lib/operation.js
ton11797: fixed

src/pipeline.cc
```diff
@@ -1613,10 +1613,16 @@ Napi::Value pipeline(const Napi::CallbackInfo& info) {
     }
   }
   if (options.Has("recombMatrix")) {
-    baton->recombMatrix = std::unique_ptr<double[]>(new double[9]);
+    baton->recombMatrixSize = sharp::AttrAsInt32(options, "recombMatrixSize");
```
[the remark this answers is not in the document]
ton11797: fixed

lovell:
Thanks again for adding this feature, I made a couple of small changes and it's now on the `main` branch as commit https://github.com/lovell/sharp/commit/60c5c5083dd12fae04db2640bb5c40564bb0728d for inclusion in the v0.33.5 release.

Issue #4144, opened 2024-07-01
Proposal: opt-in mode to pre-orient images


happycollision:
> Please forgive me for not using your feature request template. I still hit all the points that you are intending me to supply, but I have done so using a different narrative structure.

# Proposal: add an opt-in mode to conceptualize all commands starting from display orientation instead of actual orientation

> I have already done 80% (?) of the work (assuming you like my approach) for this proposal in a fork. You can [look at the tests](https://github.com/happycollision/sharp/commit/758c7bf5d867b248c7e1b82e4c6017c19bcce4ff#diff-5be3a10f0879c68466425e29c02b2d68c64533848182ad9866e8f411ada64b11) if you want to see the changes I made to achieve the API I am proposing. It is still a work in progress, but the new mode already works as intended.

## Summary

Sharp manipulates images based on actual orientation instead of display orientation. This creates surprising results that many developers can fix with `.rotate()`. But sometimes `.rotate()` is not viable or unavailable. (Read more below for why that might be.)

I'd like to propose an **opt-in mode** that allows users to issue commands to Sharp based on starting from the _display orientation_ of an image instead of the _actual orientation_. In other words, Sharp should effectively load an image, auto-orient it and then let users continue to issue `.rotate` and `.flip`/`.flop` commands.

See the ["This solution"](#this-solution) section below for the finer details.

## Terminology

If there is agreed-upon language for these concepts, I am unaware, so I will use the following:

- **Actual orientation**: the way the image's pixels are arranged in the file on disk before any EXIF orientation (if present) is applied.  
- **EXIF orientation**: The metadata which indicates how an image should display, relative to its actual orientation. The image above had its EXIF orientation removed so that it would display with its actual orientation.
- **Display orientation**: The orientation an image appears to have after EXIF orientation is applied to the actual orientation. In images without EXIF data, this is the same as the actual orientation.

|Actual Orientation|EXIF Orientation|Display Orientation|
|:-:|:-:|:-:|
|<img src="https://github.com/happycollision/md_drafts/assets/3663628/d5e96864-314f-4858-8eab-0864fde4342c" width="225">|`8`|<img src="https://github.com/happycollision/md_drafts/assets/3663628/63e90fbb-c629-4289-a0fe-eccc81901178" width="300">|
[...]

PR #4151, opened 2024-07-03
Rotation sugar and more

happycollision:
Working branch for [ideas discussed in this proposal](https://github.com/lovell/sharp/issues/4144).

- [x] add tests
- [x] fix flip/flop after auto-rotate
- [x] add `.autoOrient()` (alias for `.rotate()`)
- [x] ensure you can chain an additional `.rotate(angle)` after `.rotate()`
- [x] maybe add an option to always auto orient
- [x] Drop TEMP commit
- [x] Add constructor option `{ autoOrient: boolean }`
- [x] Add tests for composite option
- [x] Add composite option `{ autoOrient: boolean }`
- [ ] Documentation overhaul
- [ ] Add metadata adjustment


test/unit/rotate.js
```diff
@@ -50,6 +50,57 @@ describe('Rotation', function () {
...
+      [90, 180, 270, 45].forEach(function (angle) {
+        const [inputWidth, inputHeight] = orientation === 'Landscape' ? [600, 450] : [450, 600];
+        const expectedOutput = fixtures.expected(`${orientation}_${exifTag}_rotate${angle}-out.jpg`);
+        it(`${orientation} image with EXIF Orientation ${exifTag}: Auto-rotate then rotate ${angle}`, function (done) {
+          const [width, height] = angle === 45 ? [742, 742] : [inputWidth, inputHeight];
+          const [expectedWidth, expectedHeight] = angle % 180 === 0 ? [width, height] : [height, width];
+
+          sharp(input)
+            .rotate()
+            .rotate(angle)
```
happycollision, line 75: Some level-setting. I want to make sure we are on the same page, @lovell.

If I read your comments in the proposal correctly, you expect that this method of auto-orienting and then adding an additional rotation should already  work.

If you run these tests, you'll see it only works for the 45 degree angle.

Whether you implied it should already work today or not, I think your preference is that this _actually is_ the correct way to achieve the goals of my proposal.

happycollision (commented):
Okay, @lovell, I think these changes are ready for a review.

There is a failing test, but it seems to be related to the recent change to recomb.

lovell:
Thank you very much for working on this.

Thinking a bit more about the public API for this, I prefer your original suggestion to add a constructor property as it can then be used consistently for additional inputs e.g. when compositing.

```js
// PROPOSED API, NOT YET AVAILABLE
await sharp(input, { autoOrient: true })...
```
which would then also allow:
```js
// PROPOSED API, NOT YET AVAILABLE
await sharp(base, { autoOrient: true })
  .composite([{
    input: respectExifOrientationOverlay,
    autoOrient: true
  }, {
    input: ignoreExifOrientationOverlay
  }])
  ...
```
This should involve moving the existing (internal) `useExifOrientation` option into `inputDescriptor` object and its various instances.

happycollision:
I’ll probably make that change on or before Friday. I assume you still want to keep `.rotate()` working as is, since removing that would be a breaking change. But how about `.autoOrient()` (the method call)? I know I would personally like it to stay, because of other tools that do not expose a way to alter options for an instance, but do expose all methods.

lovell:
> I assume you still want to keep .rotate() working as is,

Yes please; it can be deprecated at some point in the future, but not yet.

> because of other tools that do not expose a way to alter options for an instance

This feels like a missing feature of the tools. Do you have any examples? Let's try to improve the tools rather than create duplicate API.

happycollision:
### Summary

I think having both is valid for a couple reasons. But this is not my library and I certainly have no idea what goes through your head as the maintainer. I am thrilled that any of this is happening at all, and I'll gladly show the final, merged PR to the maintainers of these aforementioned libraries and encourage them to allow downstream users to have access to the new feature.

### Explanation

> This feels like a missing feature of the tools. Do you have any examples?

I agree with that sentiment, and it makes sense as the ideal. Let's take one case to illustrate a general problem that I assume arises naturally, though.

The whole reason I got interested in creating my proposal for these changes was because I'd like to use a great tool called [svelte/enhanced-img](https://github.com/sveltejs/kit/tree/main/packages/enhanced-img). It uses another tool called [imagetools](https://github.com/JonasKruckenberg/imagetools) which is what ultimately calls into Sharp. I described the problems there in a couple issues and briefly in my proposal.

Imagetools [won't allow a rotation of either `undefined` or 0](https://github.com/JonasKruckenberg/imagetools/blob/2da03b253eba04210f55f3abd08934105646bc0d/packages/core/src/transforms/rotate.ts#L12), and there is no way for me to specify options for Sharp instances. Ideally, that library author would have noted the importance of `.rotate()`, but it is understandable to have missed it because of the name.

That tool seems to ignore the current options you can specify in Sharp. This is likely because the kind of things you do in the options for Sharp are out of scope. Reading raw data, creating new images from scratch, adding text on top of an image, etc. These are not what you might expect as common cases for a tool like this. Its main job is to take an image and allow the developer to created tailored versions of that image on the fly when importing them into their various components for use in a website.

All this to say, I can see where a developer of a tool might breeze right past options and go straight to all the methods documentation to see "how do I manipulate an image I've already loaded into a Sharp instance?"

Whether right or wrong, I understand why this might happen.

Adding an explicit method called `autoOrient` would help prevent other authors from missing the subtlety inherent to the `.rotate` method. They'd just see it in the docs and realize, "Oh I need to support this because that sounds like a real issue".

> Let's try to improve the tools rather than create duplicate API.

In writing up the response above, I read the docs about Sharp's options object a little more intentionally than I have in the past. I'd argue that `.autoOrient` as a method fits the current ethos of Sharp, whereas adding it _only_ as an option does not.

The options are concerned with how to read in an image, how to create a new image from scratch, or how to create new text from scratch. The methods are concerned with what you do now that you have an image loaded. Adjusting the orientation feels very much like the latter and not like the former.

I wholeheartedly agree that making auto orientation available as a composite option makes sense. That is a feature I previously knew nothing about, and it would be a shame if it wasn't supported there.

Ultimately, I have convinced myself that both `sharp(img).autoOrient()` and `sharp(base).composite([{ input: img, autoOrient: true }])` are valid uses. Which, based on the inner workings, might mean that `sharp(img, { autoOrient: true })` will naturally fall out of that as well.

lib/constructor.js
```diff
@@ -357,6 +357,7 @@ const Sharp = function (input, options) {
     }
   };
   this.options.input = this._createInputDescriptor(input, options, { allowStream: true });
+  this.options.useExifOrientation = !!this.options.input.autoOrient;
```
happycollision, line 360: Not sure if you'd rather do this or rename useExifOrientation. The main part of the pipeline is looking for `useExifOrientation` to do auto orientation, while the composite part of the pipeline looks for `autoOrient`.

happycollision:
I've added the options object API and gotten it working with composite images as well.

I assume that the docs need a little love at this point.

lib/constructor.js
happycollision: Update: I went ahead and replaced `options.useExifOrientation` with `options.input.autoOrient` everywhere, and let that cascade into the C++ stuff as well.

lib/operation.js
```diff
@@ -57,11 +43,11 @@ const is = require('./is');
  * @throws {Error} Invalid parameters
  */
 function rotate (angle, options) {
-  if (this.options.useExifOrientation || this.options.angle || this.options.rotationAngle) {
+  if (this.options.angle || this.options.rotationAngle) {
```
happycollision, line 46: Dropping the check for auto-orientation here seems to be what the rotate function actually should do if it is going to warn. Calling `.rotate().rotate(90)` is a legit operation—even if we now prefer `.autoOrient().rotate(90)`. What we ought to warn about is an operation where commands are effectively dropped: `.rotate(90).rotate(180)`. I changed the test to reflect this as well.

lib/operation.js
```diff
@@ -57,12 +43,12 @@ const is = require('./is');
  * @throws {Error} Invalid parameters
  */
 function rotate (angle, options) {
-  if (this.options.useExifOrientation || this.options.angle || this.options.rotationAngle) {
+  if (!is.defined(angle)) { return this.autoOrient(); }
+
+  if (this.options.angle || this.options.rotationAngle) {
     this.options.debuglog('ignoring previous rotate options');
```
happycollision, line 49: This warning _says_ that previous rotations will be ignored, but that isn't actually true unless you reset `options.angle` and `options.rotationAngle` to `0` here. Unless you do, these two sets of commands will produce a different output:

```
sharp(img)
  .rotate(90) // cardinal angle
  .rotate(5)
// output will be rotated 95 degrees

sharp(img)
  .rotate(91) // non-cardinal angle
  .rotate(5)
// output will be rotated 5 degrees
```

Want me to throw in a fix for that? Or alter the warning? Or just leave as is?

```suggestion
    this.options.debuglog('ignoring previous rotate options');
    this.options.angle = 0;
    this.options.rotationAngle = 0;
```

src/pipeline.cc
happycollision, line 1: I am commenting on this file only so this conversation can be in a single thread. These thoughts are about the entire PR.

Everything in this PR is backwards-compatible based on documentation, but not in actual effect.

The changes in behavior for `.rotate()` are evidenced by the failing tests that I added in one of my first commits. I am guessing there will be some people out there who are depending on the old behavior.

in a separate branch (the one I originally created to prove my ideas before making this proposal), I kept the behavior of `.rotate()` untouched while also introducing a new was of invoking auto-orientation.

I could rewrite a few things in this PR and do the same here. The idea being that if you use `.rotate()`, the re-orientation and subsequent calls to `rotate` would produce the old output, while using `auto-orient` would always invoke the new behavior.

The documentation and runtime warnings could deprecate the use of `.rotate()`, and encourage `autoOrient`. Then the code that preserves the old behavior could be removed at the next major.

Between the tests on my old branch which I could pull into this one, and the tests on this branch, I think it’d be fairly easy to juggle these behaviors.

docs/api-operation.md
```diff
@@ -60,6 +45,34 @@ const resizeThenRotate = await sharp(input)
 ```
 
 
+## autoOrient
+> autoOrient() ⇒ <code>Sharp</code>
+
+Auto-orient based on the EXIF `Orientation` tag, then remove the tag.
```
happycollision, line 51: It occurs to me now that I am trying to test this new version of Sharp with `@sveltejs/enhanced-img` that the metadata is not adjusted in time for some useful operations to take place. For example:

```
sharp(img, {autoOrient: true]}).metadata()
```

The above will show the metadata before any auto-orientation takes place, which seems like a headache from the perspective of someone using the tool. If I am asking sharp to auto-orient the image, why would I want the old width and height?

I see several ways to work this issue:

1. No change. All metadata is from the input state of the image.
2. If `autoOrient` is true, drop "orientation" and swap width and height as necessary from metadata.
3. Add an option to `.metadata({ autoOrient: boolean })` that opts you in to view the adjusted metadata.
4. Export a function `autoOrientMetadata(metadata)` that devs can use as needed on metadata objects.

I lean toward option 3, with an eye toward changing the default behavior at the next major.

No matter which path you'd like to take, I think any transforms on the metadata ought to also include the original stuff as well. Maybe something like the following:

```diff
{
  format: 'jpeg',
- width: 4032,
+ width: 3024,
- height: 3024,
+ height: 4032
  space: 'srgb',
  channels: 3,
  depth: 'uchar',
  density: 72,
  chromaSubsampling: '4:2:0',
  isProgressive: false,
  resolutionUnit: 'inch',
  hasProfile: false,
  hasAlpha: false,
- orientation: 6,
+ preAutoOrientation: {
+   orientation: 6,
+   width: 4032,
+   height: 3024
}
```

lovell:
Thanks for working on this, I'm very busy at the moment and haven't had a chance to review this. Given there might be slightly breaking changes this change is unlikely to make it into the next patch release, so I'll revisit once that is published.

happycollision:
Take your time. I’d like to be sure to get this right than rush something out.

labsforge:
wow can't wait to see this merged 🙏

lovell:
@happycollision Hi Don, I'm starting to prepare the next big release and your PR is on the list of things I'd like to include. Is this something you're still interested in? I realise it's been a while so no worries if not. The work-in-progress branch is at https://github.com/lovell/sharp/tree/hat

happycollision:
Absolutely! What’s your timeline? There were lots of questions I had as I was developing that I’ll need to refresh myself with and I want to make sure I move at an appropriate clip for your schedule.

lib/operation.js
lovell: Yes please to fixing this at the same time.

docs/api-operation.md
lovell: I like the idea of adding extra fields to the metadata that include the dimensions after applying orientation, which is closest to option 5 here. There have been a few people ask questions along the lines of "why are the metadata dimensions not what I expect?" and this would help answer it.

lovell:
Great, glad to hear it, I've tried to answer some of your questions inline. This is open source so there are no deadlines or ETAs.

It's looking like the next sharp release will be dependent on an as-yet-unreleased libvips v8.16.1 so this PR isn't blocking anything right now.

happycollision:
I've rebased my branch on the current state of `hat`. I'll probably have time this week to address feedback.

happycollision:
@lovell, the kids gave me the gift of all falling asleep on time tonight, so I think I've finally addressed all your feedback. I am guessing that you'll still have some more thoughts on this PR. So let me clarify what has happened to this point:

- `.rotate(undefined)` is still supported, but is not documented. It calls `.autoOrient()` under the hood.
- multiple calls to `.rotate(number)` will only respect the final call. (This is not new to expected behavior, but is new to actual behavior
- Adds `instance.autoOrient()`, `sharp(img, {autoOrient: true})`, and `sharp(base).composite([{ input: img, autoOrient: true }])` to perform an initial rotation on the image based on exif data if present. Any calls to `.rotate()` thereafter are (effectively) applied after the initial auto orientation is applied.
- fixes `flip` and `flop` behaviors so that they can be applied after an exif auto rotation has been applied.

All of this is at the current tip of the `hat` branch.

The docs at this point might need a touch more work that I have done so far. But that is where this PR stands at this moment.

lib/index.d.ts
```diff
@@ -364,24 +364,72 @@ declare namespace sharp {
...
-         * @returns A sharp instance that can be used to chain operations
          */
         rotate(angle?: number, options?: RotateOptions): Sharp;
 
+        /**
+         * Alias for calling `rotate()` with no arguments, which orients the image based
+         * on EXIF orientsion.
+         *
+         * This operation is aliased to emphasize its purpose, helping to remove any
+         * confusion between rotation and orientation.
```
happycollision, line 424: I need to update this documentation and also the docs on rotate to indicate that this is the preferred method.

happycollision:
Oh no those workflow runs. I’ll have a look when I can.

lovell:
I've fixed up the tests and docs, plus changed the metadata property to match the name of the option that would apply it, namely `autoOrient`.

This has landed as commit https://github.com/lovell/sharp/commit/14c83e1f4c02feb3d86ccde3004651fa1b07c384 and will be part of v0.34.0.

If you spot anything else in the docs that still needs updating please let me know.

Thanks again for all your time and work on this Don, it is going to be a very welcome addition.

styfle:
Is there any benefit to changing from the legacy `.rotate()` to the newer `autoRotate: true`?

Or is this just a syntax change with the same underlying logic? I ask because I'm wondering if we should update Next.js here:

https://github.com/vercel/next.js/blob/f1312ef98d9c1fb65f2677d061a6e9f14211d9c5/packages/next/src/server/image-optimizer.ts#L557-L564

Also I discovered a bug in the types and submitted a fix here: https://github.com/lovell/sharp/pull/4362

Issue #3549, opened 2023-02-04
read/write 'tEXT' chunk


vqt123:
With the MASSIVE generative AI movement, automatic1111 has become a staple tool for image generation. This tool creates images that include lots of data regarding the configuration of any image in the 'tEXT' chunk of an image's metadata. It does not appear that sharp has a way to read or write to this section of the images data.

vqt123:
here is an image that includes the 'tEXT' data section

https://imagecache.civitai.com/xG1nkqKTMzGDvpLrqFT7WA/60d49930-c321-4012-e4d9-aee47bc49c00/width=1280

the decoded chunk is as follows:

keyword: paramaters
text:
highly detailed robot head, clockwork, transparent, filled with colorful liquid, expressive, abstract, trending on artstation, by ryohei hase Negative prompt: (bad-image-v2:0.8) Steps: 20, Sampler: DPM++ 2M, CFG scale: 7, Seed: 756648331, Size: 512x512, Model hash: c35782bad8, Model: general_realisticVisionV13, Denoising strength: 0.4, ENSD: 31337, Hires upscale: 2.5, Hires steps: 8, Hires upscaler: 4x-UltraSharp, Discard penultimate sigma: True

lovell:
"Artificial Intelligence" hyperbole aside, there are existing metadata standards such as EXIF that might be more suitable for such image generators to use, however `tEXt` chunk handling is a useful feature for PNG images in its own right.

The `tEXt` chunks are currently parsed by libspng/libvips into metadata properties with names of the form `png-comment-{index}-{keyword}`, for example `png-comment-0-parameters`. The `index` exists only within libvips to create unique names, as the PNG spec allows multiple `tEXt` chunks with the same `keyword`.

I'd suggest something like the following data structure to expose this, both when reading via [metadata()](https://sharp.pixelplumbing.com/api-input#metadata) or writing via a new property of [png()](https://sharp.pixelplumbing.com/api-output#png).
```js
// PROPOSED API, NOT YET AVAILABLE
comments: [
  { keyword: 'parameters', text: 'highly detailed...' },
  { keyword: 'Copyright', text: 'legal minefield' },
  ...
]
```

A PR to expose this in sharp would be welcome, if you're able.

mactive:
I'd like to wait this.

rvion:
[...]

PR #4157, opened 2024-07-12
Expose input PNG comments in metadata

nkeynes:
Read-side implementation of PNG text following the proposed API in #3549, ie

```
comments: [
  { keyword: 'parameters', text: 'highly detailed...' },
  { keyword: 'Copyright', text: 'legal minefield' },
  ...
]
```

Notes:
* Returns the comments in the order that they're stored in the VipsImage metadata, rather than sorting on the embedded index. It looks like this should give the same ordering under normal circumstances.
* Reuses the existing test fixture `testJoinChannel.png` that happens to already have a comment, in order to keep change size down.

### DIFF
--- a/src/metadata.cc
+++ b/src/metadata.cc
@@ -10,6 +10,8 @@
 #include "common.h"
 #include "metadata.h"
 
+static void* readPNGComment(VipsImage *image, const char *field, GValue *value, void *p);
+
 class MetadataWorker : public Napi::AsyncWorker {
  public:
   MetadataWorker(Napi::Function callback, MetadataBaton *baton, Napi::Function debuglog) :
@@ -131,6 +133,8 @@ class MetadataWorker : public Napi::AsyncWorker {
         memcpy(baton->tifftagPhotoshop, tifftagPhotoshop, tifftagPhotoshopLength);
         baton->tifftagPhotoshopLength = tifftagPhotoshopLength;
       }
+      // PNG comments
+      vips_image_map(image.get_image(), readPNGComment, &baton->comments);
     }
 
     // Clean up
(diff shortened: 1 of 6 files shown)


src/metadata.cc
```diff
@@ -246,6 +250,17 @@ class MetadataWorker : public Napi::AsyncWorker {
           Napi::Buffer<char>::NewOrCopy(env, baton->tifftagPhotoshop,
             baton->tifftagPhotoshopLength, sharp::FreeCallback));
       }
+      if (baton->comments.size() > 0) {
+        int i = 0;
+        Napi::Array comments = Napi::Array::New(env);
```
lovell, line 255: It looks like we know the number of comments here so could pre-size the array, something like:
```suggestion
        Napi::Array comments = Napi::Array::New(env, baton->comments.size());
```

src/metadata.cc
```diff
@@ -285,3 +300,21 @@ Napi::Value metadata(const Napi::CallbackInfo& info) {
 
   return info.Env().Undefined();
 }
+
+const char *PNG_COMMENT_START = "png-comment-";
+const int PNG_COMMENT_START_LEN = strlen(PNG_COMMENT_START);
+
+static void* readPNGComment(VipsImage *image, const char *field, GValue *value, void *p) {
+  MetadataComments *comments = reinterpret_cast<MetadataComments *>(p);
```
lovell, line 308: Excellent use of `vips_image_map()` to read out the values here. Is is possible to switch to the slightly safer `static_cast`?

lovell (commented):
Hello, thank you very much for this PR, I've left a couple of questions inline.

The failing CI jobs are unrelated to this change. It looks like the latest version of `node-gyp` no longer supports EOL versions of Python, and the failing environments are all using either 3.6 or 3.7 (latest is 3.12). I'll investigate this separately.

src/metadata.cc
nkeynes: Yes, static_cast works here, updated as suggested :+1:

lovell:
Thank you for the updates, please can you rebase against the latest commit on `main` to pick up the CI fixes.

hugh-bita:
look good to me ?

lovell (approved):
:+1:

PR #4168, opened 2024-07-18
Add blur precision

marcosc90:
This PR adds `precision` option to `.blur()`. 

Currently, using blur on large images is quite slow. In my case, setting [precision](https://www.libvips.org/API/current/libvips-basic.html#VipsPrecision) to `approximate` significantly improves performance, reducing the processing time from over a minute to just a few seconds.


lib/constructor.js
```diff
@@ -226,6 +226,7 @@ const Sharp = function (input, options) {
     negateAlpha: true,
     medianSize: 0,
     blurSigma: 0,
+    precision: 'integer',
```
marcosc90, line 229: If I'm not mistaken this is the current default if not precision is passed to vips gaussblur

marcosc90:
> FWIW, I'm uncertain about exposing `VIPS_PRECISION_APPROXIMATE` because it's not production-quality and remains incomplete, see e.g. issue https://github.com/libvips/libvips/issues/1162.

That's unfortunate, maybe I could remove `approximate` from the docs and expose it only for advanced users?

lovell:
Thank you very much for the PR Marcos.

It looks like https://github.com/libvips/libvips/issues/1162 relates to (u)short images processed by `vips_conv` /`vips_conva`, but Gaussian blur uses a (faster) separable matrix via `vips_convsep` / `vips_convasep`, so might be unaffected.

I guess as a compromise we could limit the use of the `approximate` value to uchar images, or force-cast to uchar?

lib/operation.js
```diff
@@ -371,7 +383,7 @@ function median (size) {
  * @returns {Sharp}
  * @throws {Error} Invalid parameters
  */
-function blur (sigma) {
+function blur (sigma, precision) {
```
lovell, line 386: Given `precision` is optional (and too, I suppose, is `sigma`), now might be the time to change the method signature to also accept an Object of optional properties here.

lovell (commented):
Once we're happy with the API for this, please can you update the TypeScript definitions at https://github.com/lovell/sharp/blob/main/lib/index.d.ts

lib/operation.js
marcosc90: Sounds good to me, to confirm you want to do a breaking change to: `blur(options)`: `{ sigma, precision }`

marcosc90:
> Thank you very much for the PR Marcos.
> 
> It looks like [libvips/libvips#1162](https://github.com/libvips/libvips/issues/1162) relates to (u)short images processed by `vips_conv` /`vips_conva`, but Gaussian blur uses a (faster) separable matrix via `vips_convsep` / `vips_convasep`, so might be unaffected.
> 
> I guess as a compromise we could limit the use of the `approximate` value to uchar images, or force-cast to uchar?

You're welcome! Would you prefer to fall back to `integer` in case `approximate` is passed for ushort images, or throw an error?

lib/operation.js
lovell: Non-breaking please, if possible. If the first parameter is a number (rather than an Object) then we should be able to assume it is the value for `sigma`.

lib/operation.js
```diff
@@ -367,23 +378,43 @@ function median (size) {
...
  * @throws {Error} Invalid parameters
  */
-function blur (sigma) {
-  if (!is.defined(sigma)) {
+function blur (options) {
+  let sigma;
+  if (is.number(options)) {
+    sigma = options;
+  } else if (is.plainObject(options)) {
+    if(!is.number(options.sigma)) {
```
marcosc90, line 392: @lovell made `options.sigma` required if options object is passed, given that `precision` is only used in `gaussblur`

kleisauke:
> It looks like https://github.com/libvips/libvips/issues/1162 relates to (u)short images processed by `vips_conv` /`vips_conva`, but Gaussian blur uses a (faster) separable matrix via `vips_convsep` / `vips_convasep`, so might be unaffected.
> 
> I guess as a compromise we could limit the use of the `approximate` value to uchar images, or force-cast to uchar?

Good catch, `vips_convasep()` is indeed not affected. 

I opened PR https://github.com/libvips/libvips/pull/4055 to address the crash in `vips_conva()` that occurs with {u,}{short,int} images, so ignore my previous comment.

test/unit/blur.js
```diff
@@ -91,4 +104,33 @@ describe('Blur', function () {
...
+      sharp(fixtures.inputJpg).blur({ sigma: 1, precision: 'invalid' });
+    });
+  });
+
+  it('specific radius 10 and precision approximate', async () => {
+    const approximate = await sharp(fixtures.inputJpg)
+      .resize(320, 240)
+      .blur({ sigma: 10, precision: 'approximate' })
+      .toBuffer();
+    const integer = await await sharp(fixtures.inputJpg)
```
lovell, line 119:
```suggestion
    const integer = await sharp(fixtures.inputJpg)
```

test/unit/blur.js
```diff
@@ -91,4 +104,33 @@ describe('Blur', function () {
           });
       });
   });
+
+  it('invalid precision', function () {
+    assert.throws(function () {
+      sharp(fixtures.inputJpg).blur({ sigma: 1, precision: 'invalid' });
+    });
```
lovell, line 111: Could add an assertion for the error message here.
```suggestion
    }, /Expected one of: integer, float, approximate for precision but received invalid of type string/);
```

test/unit/blur.js
```diff
@@ -91,4 +104,33 @@ describe('Blur', function () {
...
+
+    await new Promise(resolve => {
+      fixtures.assertSimilar(fixtures.expected('blur-10.jpg'), approximate, resolve);
+    });
+  });
+
+  it('options.sigma is required if options object is passed', function () {
+    assert.throws(function () {
+      sharp(fixtures.inputJpg).blur({ precision: 'invalid' });
+    });
```
lovell, line 134: Ditto possible error message assertion.

test/unit/blur.js
```diff
@@ -91,4 +104,33 @@ describe('Blur', function () {
...
+      .toBuffer();
+    const integer = await await sharp(fixtures.inputJpg)
+      .resize(320, 240)
+      .blur(10)
+      .toBuffer();
+
+    assert.notDeepEqual(approximate, integer);
+
+    await new Promise(resolve => {
+      fixtures.assertSimilar(fixtures.expected('blur-10.jpg'), approximate, resolve);
```
lovell, line 127: Could use `util.promisify()` here, although I should probably deal with this properly and modernise `assertSimilar` to return a Promise.

lovell (commented):
Thanks for the updates, I've left a couple of comments inline otherwise this is good to go.

test/unit/blur.js
marcosc90: Thought about refactoring `.assertSimilar` but it seemed out of scope of this PR, I probably can get to it in another PR if you'd like.

test/unit/blur.js
marcosc90: My bad, missed it. Solved.

test/unit/blur.js
marcosc90: Done, both.

test/unit/blur.js
lovell: No worries, I can fix that afterwards, no need to change anything in this PR.

lovell:
Thank you very much for working on this Marcos.

PR #4172, opened 2024-07-21
Expose optional minAmpl parameter of blur operation

marcosc90:
following https://github.com/lovell/sharp/pull/4168, this PR adds `minAmpl` option to `.blur()`, completing the exposure of all options of `vips_gaussblur()`.


lovell:
Thank you for the PR Marcos.

I usually prefer for the parameter names in sharp to match libvips for consistency as you've done here but I feel "ampl" might be better expanded in the external API to "amplitude" to remove any ambiguity, perhaps as `minAmplitude`.

It might also be good to include the libvips description of this property in the JSDocs, something like "A smaller value will generate a larger, more accurate mask".

https://github.com/libvips/libvips/blob/81b6b783f09fd7acd0e0ede1a28fef906ce810e9/libvips/convolution/gaussblur.c#L186

marcosc90:
> I usually prefer for the parameter names in sharp to match libvips for consistency as you've done here but I feel "ampl" might be better expanded in the external API to "amplitude" to remove any ambiguity, perhaps as `minAmplitude`.

I had the same thought, I preferred `minAmplitude` but went with `minAmpl` to mimic libvips. I'll push the changes in a bit.

marcosc90:
@lovell I left internal API with `minAmpl`, let me know if you want this changed as well.

lovell:
Thank you very much!

PR #4191, opened 2024-08-14
fix(types): chroma-subsampling can be undefined

DavidVaness:
change chromaSubsampling type to be a union with undefined


lovell:
Good spot, thanks for the PR.

PR #4197, opened 2024-08-23
fix(types): update Channels type

DavidVaness:
Hey @lovell, improving on the types again

This PR adds 1 and 2 as possible values for the Channels type:

1: The image has only one channel representing grayscale values. Each pixel is stored as a single intensity value, ranging from black to white.

2:  The image could have one channel for grayscale intensity and another channel for alpha (transparency). This is similar to an RGBA image but without the color information, only grayscale (intensity) and alpha.

Possibly there are other reasons for a 2 channel Image but this is the most common one in my experience

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1441,8 +1441,8 @@ declare namespace sharp {
         depth?: 'char' | 'uchar' | 'short' | 'ushort' | 'int' | 'uint' | 'float' | 'complex' | 'double' | 'dpcomplex';
     }
 
-    /** 3 for sRGB, 4 for CMYK */
-    type Channels = 3 | 4;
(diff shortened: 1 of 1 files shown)


lovell:
Thanks for the PR. It looks like `Channels` is also used to define the interface when creating an image, and that should continue to only allow a value of 3 or 4.

https://github.com/lovell/sharp/blob/fc32e0bd3f9111b80cf078df7b0cfc355695674e/lib/index.d.ts#L985-L986

Perhaps we could split this into something like `CreateChannels` vs `Channels`?

lovell:
Landed via commit https://github.com/lovell/sharp/commit/1533bf995acda779313fc178d2b9d46791349961, thank you for fixing this.

Issue #3321, opened 2022-08-04
Expose background option in constructor for PDF input


omegayal:
## Question about an existing feature
When install a custom libvips with PDF support, we can convert a PDF file to an PNG image with sharp.
However, when the PDF has transparent background, the output image always shows white background.

### What are you trying to achieve?
https://github.com/libvips/libvips/issues/995
It seems that libvips provides the option "background" when converting PDF to PNG with transparent background:
`vips copy transparent.pdf[background=0] x.png`
Does sharp also have such option? Thanks.

lovell:
We'd need to expose the `background` option in the constructor API. Happy to accept a PR, if you're able.

https://sharp.pixelplumbing.com/api-constructor#parameters
https://www.libvips.org/API/current/VipsForeignSave.html#vips-pdfload

calebmer:
This is also something I'm interested in.

I explored writing a PR. Figured it would be easy since I could follow the code path to get the [`level` constructor option to vips](https://github.com/lovell/sharp/commit/7f142bddb3780312e12f66efa5a132b5a4a43531) but after ~10min I got stuck on:

1. According to the vips documentation `background` is a `VipsArrayDouble` ([source](https://www.libvips.org/API/current/VipsForeignSave.html#vips-pdfload)). Couldn't find an example of how to construct a `VipsArrayDouble` in this repository. I did find the code for `options.create.background` ([source](https://github.com/lovell/sharp/blob/f325dc3ec9ce2e711abf9fbe2b279a0b864dd8aa/src/common.cc#L431-L438)) which I expect is in the neighborhood of what one would want here.
2. It's unclear to me how to write a test for this given as a prerequisite you need a globally installed vips with PDFium (or Poppler).

lovell:
@calebmer Great, thanks for taking a look at this. Perhaps you could copy the approach taken for properties such as `resizeBackground` that use the `AttrAsVectorOfDouble` helper? As you say, we won't be able to fully test this feature, but hopefully can at least have a couple of sanity checks e.g. valid vs invalid values.

calebmer:
Ok, following the `resizeBackground` example I was able to attempt a fix: https://github.com/lovell/sharp/pull/4207

moudgalya1223:
hello  @lovell  is this stiil an issue ?

PR #4207, opened 2024-09-03
Add PDF background color option

calebmer:
Fixes https://github.com/lovell/sharp/issues/3321.

I followed existing code for `options.subifd` and `options.create.background` to pass a [`background` option to vips](https://www.libvips.org/API/current/VipsForeignSave.html#vips-pdfload) for rendering PDFs. I _haven't tested that this works_ with a vips build containing PDF support. Instead of using `sharp`'s global vips support I override the contents of the `@img/sharp-libvips-*` packages with my own custom builds derived from https://github.com/lovell/sharp-libvips which makes testing custom sharp C code a little more difficult.

If you’re not sure this is the right way to pass the `background` option to `vips_pdfload()` ([source](https://www.libvips.org/API/current/VipsForeignSave.html#vips-pdfload)) I can try to jump through some hoops to see if this change works in my local setup.


lovell (approved):
Brilliant, thank you very much for working on this Caleb.

lovell:
@GuM75 Yes, you are correct, thanks for reporting. It's relatively complex to test this kind of change as the prebuilt binaries do not support PDF.

Commit https://github.com/lovell/sharp/commit/5b5dfbad776a2d1aba0739eb52f23b7996821682 should fix this by refactoring the way in which background colours are created/set internally. This will be in v0.34.2.

PR #4240, opened 2024-10-21
updating package.json for the vulnerblite fix

moudgalya1223:


lovell:
Please can you explain this change. It appears to make sharp a recursive dependency of itself. Perhaps you meant to update a local lockfile in your own application?

Issue #4116, opened 2024-05-28
Cannot run a packaged electron forge app


neocrao:
## Possible install-time or require-time problem

- [X] I have read and understood all of the [documentation relating to installation](https://sharp.pixelplumbing.com/install).
- [X] I have searched for known bugs relating to this problem in my choice of package manager.

You must confirm both of these before continuing.

### Are you using the latest version of sharp?

- [X] I am using the latest version of `sharp` as reported by `npm view sharp dist-tags.latest`.

If you cannot confirm this, please upgrade to the latest version and try again before opening an issue.

If you are using another package which depends on a version of `sharp` that is not the latest,
please open an issue against that package instead.

### Are you using a supported runtime?

- [X] I am using Node.js with a version that satisfies `^18.17.0 || ^20.3.0 || >=21.0.0`
- [ ] I am using Deno
- [ ] I am using Bun

If you cannot confirm any of these,
please upgrade to the latest version
and try again before opening an issue.

### Are you using a supported package manager and installing optional dependencies?

- [X] I am using npm >= 9.6.5 with `--include=optional`
- [ ] I am using yarn >= 3.2.0
- [ ] I am using pnpm >= 7.1.0 with `--no-optional=false`
- [ ] I am using Deno
- [ ] I am using Bun

If you cannot confirm any of these, please upgrade to the latest version of your chosen package manager
and ensure you are allowing the installation of optional or multi-platform dependencies before opening an issue.

### What is the complete error message, including the full stack trace?

Below are some of the things I have tried as I scoured through the the reported issues in this repository. None of the solutions worked for me. I am hoping that its something simple that I have missed.

I'd appreciate any help.

<details>

Steps to reproduce with different attempts to solve the issue:

1. Setup an electron forge project (https://github.com/neocrao/electron-forge-sharp/commit/cb18b5d86809cf9b9c7ad6f1958cb9081bf91d5e)

    ```
    npm init electron-app@latest my-app -- --template=webpack-typescript
    ```

    `npm run start` and `npm run make` work without any issues

1. Install `sharp` (https://github.com/neocrao/electron-forge-sharp/commit/f6571bddced3a1a695e70ca2633c483b52381f8e)

    ```
    npm install -S --include=optional sharp
    ```

1. Use `sharp` in the project (https://github.com/neocrao/electron-forge-sharp/commit/fae3202d9b341c3139209e04456619c71ce7b097)

    `npm run start` works.
    
[...]

PR #4261, opened 2024-11-11
Update install doc for Electron-builder and Electron-forge

Ymihere03:
Update install documentation for Electron

Renamed the previous Electron section to 'electron-builder' and fixed the broken external link

Added a new section for 'electron-forge' with its own external link to electron-forge documentation and included sharp ASAR configuration. The config for 'electron-forge' is based on part of the resolution in https://github.com/lovell/sharp/issues/4116

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -233,11 +233,10 @@ custom:
       scripts:
         - npm install --os=linux --cpu=x64 sharp
 ```
-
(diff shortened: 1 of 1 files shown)


lovell:
Landed as eba82a8, thank you very much for the PR.

PR #4282, opened 2024-11-23
chore(benchmark): remove simd(true)

styfle:
Since `simd(true)` is the default behavior, these can be removed.

https://sharp.pixelplumbing.com/api-utility#simd


lovell:
Thank you Steven!

PR #4305, opened 2024-12-23
Docs: prebuilt binaries support tile-based output

kleisauke:
Except when targeting WebAssembly.

Context: https://github.com/kleisauke/wasm-vips/issues/87#issuecomment-2559951532.


lovell:
Thanks Kleis!

PR #4347, opened 2025-03-16
Ensure Yarn PnP hash calculation works with pre-releases

kleisauke:
Before:
```console
$ ls -1 .yarn/unplugged | grep sharp-libvips-linux-x64
@img-sharp-libvips-linux-x64-npm-1.1.0-rc4-5841416c64
$ yarn node -p "require('sharp/lib/libvips').yarnLocator()"
6e8b62218c
```

After:
```console
$ ls -1 .yarn/unplugged | grep sharp-libvips-linux-x64
@img-sharp-libvips-linux-x64-npm-1.1.0-rc4-5841416c64
$ yarn node -p "require('sharp/lib/libvips').yarnLocator()"
5841416c64
```

### DIFF
--- a/lib/libvips.js
+++ b/lib/libvips.js
@@ -113,7 +113,9 @@ const sha512 = (s) => createHash('sha512').update(s).digest('hex');
 const yarnLocator = () => {
   try {
     const identHash = sha512(`imgsharp-libvips-${buildPlatformArch()}`);
-    const npmVersion = semverCoerce(optionalDependencies[`@img/sharp-libvips-${buildPlatformArch()}`]).version;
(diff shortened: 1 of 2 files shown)


kleisauke:
An alternative approach could be to simply remove the `semverCoerce()` call, as Yarn Plug'n'Play likely(?) uses the raw tag value here (e.g. `2.0` remains `2.0` rather than being converted to `2.0.0` first). However, this would cause issues when referencing these dependencies with version ranges (e.g. `^1.1.0`, `=1.1.0`, etc.).

lovell:
Brilliant, thank you for investigating and fixing this Kleis.

PR #4362, opened 2025-04-04
fix(types): update `autoOrient` type to include undefined

styfle:
This PR fixes the following error:

```
node_modules/sharp/lib/index.d.ts(1590,15): error TS2430: Interface 'OverlayOptions' incorrectly extends interface 'SharpOptions'.
  Types of property 'autoOrient' are incompatible.
    Type 'boolean | undefined' is not assignable to type 'boolean'.
      Type 'undefined' is not assignable to type 'boolean'.
```

The bug was introduced in:

- https://github.com/lovell/sharp/issues/4144
- https://github.com/lovell/sharp/pull/4151

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -971,7 +971,7 @@ declare namespace sharp {
          *
          * Using this option will remove the EXIF `Orientation` tag, if any.
          */
-        autoOrient?: boolean;
+        autoOrient?: boolean | undefined;
         /**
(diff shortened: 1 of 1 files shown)


styfle:
I discovered this when updating `vercel/next.js` here: 

- https://github.com/vercel/next.js/pull/77839

lovell:
Thank you Steven!

PR #4368, opened 2025-04-08
Docs: update path to `api-resize-fit.svg`

kleisauke:


lovell:
Thanks Kleis, missed that one with the switch to the new docs site. I'll take a look to see if we can get Astro to manage this file as part of the published site now rather than go via jsdelivr.

PR #4369, opened 2025-04-11
fix(types): Add missing properties for animated images

sdonfro:
Hello! This PR is to add two missing properties to `sharp.OutputInfo` that are present for animated images. Comments are copied from the same properties in `sharp.Metadata`.

The test `Animated GIF round trip` in `test/unit/gif.js` references them ([link](https://github.com/lovell/sharp/blob/5b5dfbad776a2d1aba0739eb52f23b7996821682/test/unit/gif.js#L67-L68)), so it does seem intentional for them to be part of the OutputInfo.

This is my first PR for the project, so please let me know if you need more information from me. 
Thanks for all the hard work on this library!


lovell:
Thank you very much Steven!

PR #4387, opened 2025-05-04
Fix(WebP): Add smartDeblock definition to WebpOptions

Stephen-X:
This PR proposes a small fix to add the missing typescript definition for the `smartDeblock` option introduced in https://github.com/lovell/sharp/commit/8afec170ed81bc840157ecf5d3ee07ae200ffe33.

I'm not entirely sure yet if I should update [`sharp.test-d.ts`](https://github.com/lovell/sharp/blob/701143afb364c1b618422c59ae955f4084651930/test/types/sharp.test-d.ts). Testing for `WebpOptions` seems scattered and there's no test for all options e.g. `smartSubsample`. But I'm happy to add one if needed.


lovell:
Thank you!

PR #4395, opened 2025-05-13
Minor simplifications

kleisauke:
See the individual commits for more details.


lovell:
Merged as separate commits, dankjewel Kleis.

PR #4397, opened 2025-05-16
Fix pnpm supportedArchitectures link in install.md

harshalbhakta:


lovell:
Thanks for the PR. There are a couple of other links to `pnpm.io` in this same file - do they need updating also?

PR #4402, opened 2025-05-28
Regenerate expected fixtures for libvips v8.17.0

kleisauke:
Regenerate `extract.tiff` to avoid an test failure due to PR https://github.com/libvips/libvips/pull/4313.
```console
$ curl -LO https://github.com/lovell/sharp/raw/v0.34.2/test/fixtures/expected/extract.tiff
$ vips copy extract.tiff[fail-on=warning] x.jpg 

(vips:22733): tiff2vips-WARNING **: 14:54:38.286: The JPEG strip/tile is encoded with progressive mode, which is normally not legal for JPEG-in-TIFF.
libtiff should be able to decode it, but it might cause compatibility issues with other readers

(vips:22733): VIPS-WARNING **: 14:54:38.287: error in tile 0 x 0
```


lovell:
Thanks Kleis, your suggested change for the TIFF extraction test looks good. For the composite tests, does forcing a higher PNG bitdepth for both the expected and actual fixtures  via something like `.toColourspace('rgb16')` help?

lovell:
Thank you Kleis.

PR #4408, opened 2025-06-05
Fixes CVE-2025-48387

aashutoshrathi:
Bumps `tar-fs` to `^3.0.9`
Ref: https://nvd.nist.gov/vuln/detail/CVE-2025-48387

### DIFF
--- a/package.json
+++ b/package.json
@@ -184,7 +184,7 @@
     "nyc": "^17.1.0",
     "prebuild": "^13.0.1",
     "semistandard": "^17.0.0",
-    "tar-fs": "^3.0.8",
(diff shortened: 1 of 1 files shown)


lovell:
Thanks for the PR. As you probably saw, `tar-fs` is part of `devDependencies` and is only used by the unit tests on trusted input. It is *not* a runtime dependency of sharp so the vulnerability mentioned does not apply.

In addition, the `^3.0.8` semver range includes the latest version so running `npm upgrade` locally will update any lockfiles.

aashutoshrathi:
noted

Issue #3650, opened 2023-04-27
Compositing a cmyk image onto another cmyk image without switching to rgb


Aminelahlou:
## Feature request
I would like to composite a cmyk image onto another cmyk image 
### What are you trying to achieve?
I would like to composite cmyk images onto one cmyk image without converting to RGB the composited images

### When you searched for similar feature requests, what did you find that might be related?

I did not find any because we are able to open cmyk images only since 0.32.1 so people were not able to composite images using cmyk images

### What would you expect the API to look like?

Same api as composite method

### What alternatives have you considered?

I considered fusing cmyk image buffers but it would be equivalent to rewriting composite function.

### Please provide sample image(s) that help explain this feature

I have made a github repo with the input image (cmyk image with a Cyan value of 100 of 1000 pixel by 1000 pixel).
What the repo does : 
1. creating a label
2. Modulating an image
3. Compositing the label onto the modulated CMYK image (the label is just for information purpose so we don't care if it uses RGB and then is converted to cmyk)
4. creating a white image which would include all the images
5. compositing the modulated cmyk image with the label, onto the white cmyk image

When I execute this code, I have this error :  ```
Input buffer contains unsupported image format
```
which means i cannot composite cmyk image onto another without losing information about the composited cmyk image since it needs to be converted into RGB unless there is a way I did not see.

I have made a github repo available here: https://github.com/Aminelahlou/sharpCmykTest (same as issue #3620)
The image is input/test-sharp-input.tif
The script to execute is test-sharp-composite.js 

```
const sharp = require("sharp");
async function compositeFunc() {
  const labelInput = await sharp({
    text: {
      text: "random text",
      font: "Arial",
      width: 500,
      height: 400,
      channels: 4,
      background: { r: 255, g: 255, b: 255, alpha: 1 },
    },
  })
    .tiff()
    .toBuffer();
  const modulatedImage = sharp("./input/test-sharp-input.tif")
    .pipelineColourspace("cmyk")
    .modulate({ hue: 0, saturation: 1, brightness: 1 })
    .composite([
      {
        input: labelInput,
        width: 500,
        height: 400,
        channels: 4,
        top: 0,
        left: 0,
      },
    ])
    .toColourspace("cmyk");

  const whiteImage = await sharp({
    create: {
      width: 15000,
      height: 15000,
      channels: 4,
      background: { r: 255, g: 255, b: 255 },
      limitInputPixels: false,
    },
  })
    .toColorspace("cmyk")
[...]

PR #4412, opened 2025-06-08
Support composite op with non-sRGB pipeline colourspace

kleisauke:
Split out from #4402.


lovell:
Excellent, thank you Kleis, this implements the enhancement previously discussed at https://github.com/lovell/sharp/issues/3650 too.

lovell:
This isn't really a breaking change, and nor is the libvips upgrade to 8.17.0, so I've re-tagged for inclusion in v0.34.3.

PR #4415, opened 2025-06-16
Ensure SVG scale-on-load optimisation uses newly exposed params

kleisauke:
Follow-up to commit c4b1d80.


lovell:
Thanks Kleis, I'm going to refactor this to remove the duplicate code as it has been the source of a number of bugs.

lovell:
Commit https://github.com/lovell/sharp/commit/e286e2bff9ac7142cc32c6a476fe5d9e45f3d23b moves this logic into a single option-building function, which should hopefully reduce the chance of future mistakes.

PR #4422, opened 2025-07-10
Remove redundant TIFF warning handler setup

kleisauke:
This became redundant after commit libvips/libvips@22994d9.

### DIFF
--- a/src/sharp.cc
+++ b/src/sharp.cc
@@ -18,10 +18,8 @@ Napi::Object init(Napi::Env env, Napi::Object exports) {
     vips_init("sharp");
   });
 
-  for (auto domain : { "VIPS", "vips2tiff" }) {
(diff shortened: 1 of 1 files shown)


lovell:
I was planning to leave this logic in place for a bit longer as some OS and package managers for which prebuilt binaries are not provided e.g. FreeBSD still provide libvips v8.17.0.

kleisauke:
Isn't that an invalid use case? See for example:
https://github.com/lovell/sharp/blob/276ac2df8d83f357f2d6825a9f0fc53ceb07d6b0/src/common.h#L17-L21

In other words, you can't compile sharp v0.34.3 with libvips v8.17.0.

lovell:
Ha ha, oh yeah, I try to be an angel to FreeBSD's demon but look what happens! :sweat_smile:

PR #4441, opened 2025-08-21
Types: add "linear" to `KernelEnum`

BayanBennett:
This matches the JS code

https://github.com/lovell/sharp/blob/8607ff2f4a5d0f523f57c589378256a626b62c97/lib/resize.js#L69-L78


lovell:
Thank you!

PR #4444, opened 2025-08-26
Improve Emscripten linker flags

kleisauke:
See the individual commits for more details.


lovell:
Thanks Kleis, I've retained the individual commits here during the merge as these provide very useful history.

PR #4449, opened 2025-09-08
Update to color ^5.0.0

mrginglymus:
Color had transitive dependencies on is-arrayish and simple-swizzle, two recently compromised packages.

As simple-swizzle is no longer recommended (and never really was) this seems like a good time to update to the latest version of color which eliminates those dependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,7 +137,7 @@
     "vips"
   ],
   "dependencies": {
-    "color": "^4.2.3",
+    "color": "^5.0.1",
     "detect-libc": "^2.0.4",
     "semver": "^7.7.2"
   },


lovell:
Thanks for the PR, have you tested this? I ask as I had previously investigated this upgrade only to discover v5 of the `color` package had switched to ESM-only.

mrginglymus:
Still working on getting tests running - having some "being a windows developer" issues.

43081j:
@lovell one thing worth noting is that all LTS versions of node now support `require(esm)` (i.e. you can `require` an esm-only module)

so this may be opportunity to bump your `engine` constraint just the same (and as a prerequisite to doing this upgrade)

mrginglymus:
@lovell WSL to the rescue - full test suite pass on node `v22.15.1`.

But yes as @43081j, only on LTS nodes

kleisauke:
Both [^1] and [^2] indicate that version 5.0.1 of color is _also_ compromised.

[^1]: https://github.com/debug-js/debug/issues/1005#issuecomment-3266868187
[^2]: https://bsky.app/profile/bad-at-computer.bsky.social/post/3lydje4zqis2y

lovell:
sharp currently needs to support versions of Node.js that do not provide ESM and this will be the case for a while yet.

It's all moot anyway as the latest v5.0.1 of the `color` package appears to *also* be impacted.

The correct action that we all need right now is for npm (owned by Microsoft, annual income of US$245 billion) to unpublish *all* of the affected dependencies.

ascorbic:
This isn't about the compromise of color and the other related packages: the affected version has already been unpublished. The issue is the dependency on simple-swizzle. I don't think sharp can upgrade to an ESM-only version while still supporting Node 18.

ascorbic:
It looks like simple-swizzle@0.2.3 has finally been removed from npm

43081j:
yes, to be clear, the focus in this issue shouldn't be that the packages were compromised

the issue is that the swizzle package has long since been deprecated by the maintainer and advised against being used. so its good clean up to move off it

we can't do that here by upgrading to `5.x` since it means we need to bump sharp's minimum node to `20`

but we could look into alternatives to `color` (which support `>=18` and are as lightweight or even better)

it may be sensible to close this PR and track it in an issue instead for that reason

mrginglymus:
I agree with @43081j that given the requirement to support old node, perhaps an alternative library would help clean up the dep tree.

On the vulnerability side - 5.0.1 is _not_ currently vulnerable. A vulnerable version was published with a version of 5.0.1; this was removed and a new clean version was published over the top. Not quite sure _why_, but it is safe.

PR #4455, opened 2025-09-20
Docs: libvips manages its own thread pool

kleisauke:
Since version 8.14, libvips no longer uses [`GThreadPool`](https://docs.gtk.org/glib/struct.ThreadPool.html).
https://www.libvips.org/2022/12/22/What's-new-in-8.14.html#faster-threading


lovell:
Dankjewel Kleis!

Issue #4421, opened 2025-07-09
Enhancement: add support for bigtiff output


chenhongtu:
### What are you trying to achieve?

I want to export tiff files larger than 4GB and need bigtiff support

### What would you expect the API to look like?

```
sharp(input)
  .tiff({ bigtiff: true})
  .toFile('output_large.tif');
```

### What alternatives have you considered?

I have considered using libvips directly to implement

lovell:
This should be relatively easy, as `bigtiff` is an existing property of libvips' `tiffsave` and it should follow a similar pattern to the way `miniswhite` is handled. Happy to accept a PR, if you're able.

https://www.libvips.org/API/current/method.Image.tiffsave.html

arwidt:
Im a bit confused about this. I started doing a PR for this but then when I started testing I managed to create files in BigTiff format even without my changes. This seems to happen automatically.

This is with sharp version `0.34.3`
```
await sharp(veryLargeSourceImage, { limitInputPixels: false })
        .resize(40000, 40000)
        .tiff({
          compression: 'none'
        })
        .toFile(veryLargeOutput);
```

When running this with my testing image I get these images.
```
3.4G Sep  9 20:12 lossless-35000x35000.tiff
3.8G Sep  9 20:13 lossless-37000x37000.tiff
4.0G Sep  9 20:14 lossless-38000x38000.tiff
4.5G Sep  9 20:15 lossless-40000x40000.tiff
```
And when I test these with `file` I get.
```
file *.tiff
lossless-35000x35000.tiff: TIFF image data, little-endian
lossless-37000x37000.tiff: TIFF image data, little-endian
lossless-38000x38000.tiff: Big TIFF image data, little-endian
lossless-40000x40000.tiff: Big TIFF image data, little-endian
```

Im really confused since I hit the limit in my other project and thats why I found this issue. So what is going on? When is a .tiff file to large if memory is not a problem?

lovell:
@arwidt Thanks for working on this. Your experience will be due to `compression: 'none'` and https://github.com/libvips/libvips/issues/591 - we should probably update the docs to reflect this.

arwidt:
@lovell Thanks for answer. I will investigate some more. I guess we can hit the normal tiff roof then when using compression. Even though that is a ridiculously big file.

arwidt:
[...]

PR #4459, opened 2025-09-28
BigTIFF support

throwbi:
Solves #4421 

I cannot provide full unit test coverage, but I tested locally with a 8 GB image passing the commented out tests, so do with those what you will.

I obviously cannot upload a 8 GB image to the repository nor elsewhere, since the file is confidential, and I haven't found a good way to create a large enough image in test to cross 4 GB when using compression.


lovell:
Thank you, this will be part of v0.34.5

PR #4471, opened 2025-11-07
Fix invalid escape sequence

inukshuk:
Trying to get rid of this syntax warning:

https://github.com/lovell/sharp/actions/runs/19138347662/job/54696283442#step:9:48


lovell:
Thank you / danke !

PR #4474, opened 2025-11-10
Use tags for deprecations

Jiralite:
This actually tells IDEs that properties are deprecated (it will strikethrough the property and deprioritise it in autocomplete). It's already used elsewhere in the file too.

Also used `@link` references.


lovell:
Thank you!

PR #4481, opened 2025-11-29
UHDR image support

### LINKED ISSUE #4314: Please add support for HDR JPG gain maps

Add support for transcoding JPG gain maps. These images enable the ability to share images edited for modern displays that support HDR (high dynamic range). And compared to HDR-only encoding, using a gain map allows the image to be properly adapted to any display which does not support the full HDR encoding. That adaptation is critical to being able to safely share HDR images on any display - including SDR ("standard dynamic range") and monitors with limited HDR capabilities.

These HDR displays are already more common than generally known. Roughly 80% of viewers on Instagram already have HDR support and you can see examples of HDR vs standard images in my [Instagram feed](https://www.instagram.com/gregbenzphotography/) in the IG phone app (or using Chrome with an M1+ MacBook Pro).

There is a request for JXL support (https://github.com/lovell/sharp/issues/4263). While that is a great format, JPG (and AVIF) with a gain map is more practical. The JPG format is 100% backwards compatible and already widely supported. The AVIF gain map encoding will be very important and already has support in Adobe software and Chrome. We're probably a year away from having enough critical support to use AVIF, but there is clear and steady progress towards that point. There is little support for JXL and it is unclear if it will gain acceptance on the web (especially as Chrome removed support).

Should support simple transcoding of JPG (crop, resize, compress) while retaining the HDR. That means the base image and auxiliary gain map should be edited in a consistent manner, and the HDR metadata should be retained. 

The Google-backed open-source [libultrahdr](https://github.com/google/libultrahdr/issues) library already supports all of this. All that is needed is some integration with libvips.

There is a request for libvips to add support: https://github.com/libvips/libvips/issues/3799

### Please provide sample image(s) that help explain this feature

More info on [HDR](https://gregbenzphotography.com/hdr/) and [gain maps](https://gregbenzphotography.com/hdr-photos/jpg-hdr-gain-maps-in-adobe-camera-raw/). 

Here is a sample image. The image will tell you if you are viewing just the SDR base image, or viewing as HDR (when using a supported browser and display, you should see blue text). The image is encoded with both the ISO and Android XMP standards (the ISO encoding is really all that is needed, the XMP encoding is redundant for backwards compatibility, but won't be important for too long).
Image

Developer resources for gain maps: https://gregbenzphotography.com/hdr/#developers

ri-char:
This PR adds UHDR image support, leveraging the newly introduced UHDR functionality in libvips (see libvips/libvips#3799). Closed #4314.

# New Public Functions

Use these UHDR options for output image.

| Param | Type | Default | Description |
| --- | --- | --- | --- |
| [options] | <code>Object</code> |  | output options |
| [options.quality] | <code>number</code> | <code>80</code> | quality, integer 1-100 |

Keep Gainmap for uhdr image from the input image in the output image. This is the defualt behaviour for uhdr input.

Remove gainmap from the input image in the output image.


lovell:
Hello, thank you very much for taking the time to submit a PR - it provides a great start to help people manage images with gain maps.

I intend to take the integration with Ultra HDR much deeper, including it in prebuilt binaries, and supporting gain map scaling/manipulation when resizing/cropping etc. This also means I should probably think more about how this is exposed/implied in the API. I'll create a separate tracking issue with thoughts.

PR #4499, opened 2026-02-18
Fix #4475: Correct FormatEnum type keys jp2→jp2k and v→vips

### LINKED ISSUE #4475: 'jp2' and 'v' does not exist in sharp.format object

- [x] Running `npm install sharp` completes without error.
- [x] Running `node -e "require('sharp')"` completes without error.

If you cannot confirm both of these, please open an [installation issue](https://github.com/lovell/sharp/issues/new?labels=installation&template=installation.md) instead.

- [x] I am using the latest version of `sharp` as reported by `npm view sharp dist-tags.latest`.

If you cannot confirm this, please upgrade to the latest version and try again before opening an issue.

If you are using another package which depends on a version of `sharp` that is not the latest, please open an issue against that package instead.

### What is the output of running `npx envinfo --binaries --system --npmPackages=sharp --npmGlobalPackages=sharp`?

```
  System:
    OS: Linux 6.17 Arch Linux
    CPU: (16) x64 AMD Ryzen 7 PRO 4750U with Radeon Graphics
    Memory: 16.33 GB / 22.84 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Binaries:
    Node: 24.10.0 - /home/jun/.nvm/versions/node/v24.10.0/bin/node
    npm: 11.6.1 - /home/jun/.nvm/versions/node/v24.10.0/bin/npm
```

### Does this problem relate to file caching?

The default behaviour of libvips is to cache input files, which can lead to `EBUSY` or `EPERM` errors on Windows.
Use [`sharp.cache(false)`](https://sharp.pixelplumbing.com/api-utility#cache) to switch this feature off.

- [x] Adding `sharp.cache(false)` does not fix this problem.

### Does this problem relate to images appearing to have been rotated by 90 degrees?

Images that contain EXIF Orientation metadata are not auto-oriented. By default, EXIF metadata is removed.

- To auto-orient pixel values use the parameter-less [`rotate()`](https://sharp.pixelplumbing.com/api-operation#rotate) operation.
- To retain EXIF Orientation use [`keepExif()`](https://sharp.pixelplumbing.com/api-output#keepexif).

- [x] Using `rotate()` or `keepExif()` does not fix this problem.

1. Copy the provided minimal reproducible code below into `main.ts` file
2. Run `node main.ts` (Node v22.18.0 or later required for [native TypeScript support](https://nodejs.org/en/learn/typescript/run-natively))

According to the TypeScript definitions, `sharp.format` should return a `FormatEnum` object which is expected to contain `jp2` and `v` properties.

https://github.com/lovell/sharp/blob/dc6820b49f4bc5943ca7c6b1f7e379621b81f6a6/lib/index.d.ts#L47-L49

https://github.com/lovell/sharp/blob/dc6820b49f4bc5943ca7c6b1f7e379621b81f6a6/lib/index.d.ts#L1915-L1940

**Expected**

```
jp2 {
  id: 'jp2k',
  input: { file: false, buffer: false, stream: false },
  output: {
    file: false,
    buffer: false,
    stream: false,
    alias: [ 'j2c', 'j2k', 'jp2', 'jpx' ]
  }
}
jp2k undefined
v {
  id: 'vips',
  input: {
    file: true,
    buffer: false,
    stream: false,
    fileSuffix: [ '.v', '.vips' ]
  },
  output: { file: true, buffer: false, stream: false }
}
vips undefined
```

**Actual**

However, `sharp.format` returns an Object that contains `jp2k` and `vips` instead of `jp2` and `v`. This matches the test cases that I found in this repository:

https://github.com/lovell/sharp/blob/dc6820b49f4bc5943ca7c6b1f7e379621b81f6a6/test/unit/jp2.js#L14

https://github.com/lovell/sharp/blob/dc6820b49f4bc5943ca7c6b1f7e379621b81f6a6/test/unit/util.js#L134-L136

```
jp2 undefined
jp2k {
  id: 'jp2k',
  input: { file: false, buffer: false, stream: false },
  output: {
    file: false,
    buffer: false,
    stream: false,
    alias: [ 'j2c', 'j2k', 'jp2', 'jpx' ]
  }
}
v undefined
vips {
  id: 'vips',
  input: {
    file: true,
    buffer: false,
    stream: false,
    fileSuffix: [ '.v', '.vips' ]
  },
  output: { file: true, buffer: false, stream: false }
}
```

### Please provide a minimal, standalone code sample, without other dependencies, that demonstrates this problem

```typescript
import sharp from "sharp";

console.log("jp2", sharp.format.jp2);
console.log("jp2k", sharp.format.jp2k);

console.log("v", sharp.format.v);
console.log("vips", sharp.format.vips);
```

rodionsteshenko:
The `FormatEnum` TypeScript interface declared `jp2` and `v` as keys, but the runtime `sharp.format` object uses `jp2k` and `vips` (matching libvips internal format identifiers).

## Changes

- `lib/index.d.ts`: Renamed `jp2` → `jp2k` and `v` → `vips` in `FormatEnum` interface
- `test/unit/util.js`: Added two test cases verifying the correct keys exist and the old incorrect keys are undefined

## Test

Added `uses jp2k as format key, not jp2 (#4475)` and `uses vips as format key, not v (#4475)` tests that:
1. Verify `sharp.format.jp2k` exists with correct id
2. Verify `sharp.format.vips` exists with correct id  
3. Verify `sharp.format.jp2` and `sharp.format.v` are undefined

Full test suite passes on macOS ARM (Apple Silicon).

### DIFF
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -1923,7 +1923,7 @@ declare namespace sharp {
         input: AvailableFormatInfo;
         jpeg: AvailableFormatInfo;
         jpg: AvailableFormatInfo;
-        jp2: AvailableFormatInfo;
+        jp2k: AvailableFormatInfo;
         jxl: AvailableFormatInfo;
         magick: AvailableFormatInfo;
         openslide: AvailableFormatInfo;
@@ -1935,7 +1935,7 @@ declare namespace sharp {
         svg: AvailableFormatInfo;
         tiff: AvailableFormatInfo;
         tif: AvailableFormatInfo;
-        v: AvailableFormatInfo;
+        vips: AvailableFormatInfo;
         webp: AvailableFormatInfo;
     }
 
--- a/test/unit/util.js
+++ b/test/unit/util.js
@@ -142,6 +142,16 @@ describe('Utilities', () => {
     it('output alias', () => {
       assert.deepStrictEqual(['jpe', 'jpg'], sharp.format.jpeg.output.alias);
     });
+    it('uses jp2k as format key, not jp2 (#4475)', () => {
+      assert.strictEqual('object', typeof sharp.format.jp2k);
+      assert.strictEqual('jp2k', sharp.format.jp2k.id);
+      assert.strictEqual(undefined, sharp.format.jp2);
+    });
+    it('uses vips as format key, not v (#4475)', () => {
+      assert.strictEqual('object', typeof sharp.format.vips);
+      assert.strictEqual('vips', sharp.format.vips.id);
+      assert.strictEqual(undefined, sharp.format.v);
+    });
   });
 
   describe('Versions', () => {


lovell:
This has already been fixed as the linked issue #4475 states - see commit https://github.com/lovell/sharp/commit/d161e45e06b983ab6383ae102821a57872138573

It looks like you're using an "AI" agent to open a lot of identikit PRs in many other repos. If this is the case, please take a few seconds of your time to assess its output before letting it open PRs without human review.

Issue #4490, opened 2026-01-19
Please check the value instead of simply the presence


ardelio:
https://github.com/lovell/sharp/blob/7b4c4762432b14c62676e860c8034b5cd326f464/install/check.js#L8

Other packages, such as gdal-async, use this variable as a mechanism to indicate to the package that it must build from source. Said packages, as a convention, look for the presence of their package name within the value which is a more robust implementation that plays nice within the ecosystem, than simply looking at the presence of the env var.

It would be most excellent if Sharp could do the same, e.g.:

```
const buildFromSource = process.env.npm_config_build_from_source;

if (useGlobalLibvips() || (buildFromSource && buildFromSource.includes('sharp'))) {
    process.exit(1);
  }
``` 

This would improve the install experience and prevent consumers from having to install and rebuild in multiple steps to overcome Sharp's implementation.

For example, we have to do the equivalent of the following:

```
// Installs sharp as we would like, but gdal-async is not bound to our versions of gdal
npm install

// Rebuild Gdal Async bound to our versions of GDAL.
npm_config_build_from_source=gdal-async npm_config_shared_gdal=true npm rebuild gdal-async
```

Thank you!

lovell:
Hello, this logic will be going away in the next release (via commit https://github.com/lovell/sharp/commit/e1bad5470e43ea1f8895ed4694d0be7ae1cbe8e7), in part due to [npm removing custom config](https://github.com/npm/cli/issues/8153), as well as a wider move in the JS package manager ecosystem to prevent `install` scripts running by default.

ardelio:
Thank you, I'll close the issue then, cheers!

PR #4505, opened 2026-03-12
Use `brew --prefix` to get brew package config path for install script

dsherret:
We noticed the Bun package manager is excluding the install script for your package because it is slow.

https://github.com/oven-sh/bun/blob/a305e99af5d9dbe74ffabfa5a3c0cfdf815ac71a/src/install/postinstall_optimizer.zig#L14-L19

In Deno we were looking into making a similar optimization, but we don't believe it would be good to override your preferences just so we can do better on an install benchmark.

I was looking at the code and I believe that it should actually get the value from `brew --prefix` then `./lib/pkgconfig`. I'm not sure though. It is much faster to do this on my machine.

```
% hyperfine --warmup 5 "brew --prefix" "brew environment --plain"
Benchmark 1: brew --prefix
  Time (mean ± σ):      22.9 ms ±   1.6 ms    [User: 8.3 ms, System: 8.8 ms]
  Range (min … max):    18.8 ms …  29.1 ms    120 runs
 
Benchmark 2: brew environment --plain
  Time (mean ± σ):     789.2 ms ±  65.3 ms    [User: 209.1 ms, System: 139.0 ms]
  Range (min … max):   714.9 ms … 884.1 ms    10 runs
 
Summary
  brew --prefix ran
   34.41 ± 3.68 times faster than brew environment --plain
```

### DIFF
--- a/lib/libvips.js
+++ b/lib/libvips.js
@@ -144,17 +144,24 @@ const globalLibvipsVersion = () => {
   }
 };
 
+const getBrewPkgConfigPath = () => {
+  try {
+    const brewPrefix = (spawnSync('brew', ['--prefix'], { encoding: 'utf8' }).stdout || '').trim();
+    if (brewPrefix) {
+      return `${brewPrefix}/lib/pkgconfig`;
+    }
+  } catch (_err) {
+    // ignore
+  }
+  return undefined;
+};
+
 /* node:coverage enable */
 
 const pkgConfigPath = () => {
   if (process.platform !== 'win32') {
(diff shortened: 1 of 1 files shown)


lovell:
Hi David, thank you for taking the time to investigate the problem and provide a solution, this will definitely improve things for those using Homebrew and building from source .

Luckily this class of problem should disappear in the next release as build from source becomes opt-in, please see https://github.com/lovell/sharp/issues/4490#issuecomment-3767165817 for more details.

PR #4508, opened 2026-03-18
feat: cicp

mertalev:
This integrates the libvips CICP changes in [this](https://github.com/libvips/libvips/pull/4928) branch, allowing for true HDR handling across input and output formats. It supports both linearization (necessary for color conversion, but less accurate) and passthrough via a `keepCicp` method and CICP metadata that can be provided in the constructor. It's marked as draft for now given the libvips branch is not merged. It serves as a way to test the branch and to see how it can be used by libraries.


lovell:
Thanks for this Mert, I'm looking forward to the upstream PR landing!

PR #4510, opened 2026-03-19
CI: Fix wasm32 build

kleisauke:
Bump emnapi to 1.9.1 for compatibility with Emscripten 5.0.3, see:
emscripten-core/emscripten@30517257796a28976afb5d2aceff79f07def2328

/cc @toyobayashi FYI.


kleisauke:
Marking as draft due to test failures:
https://github.com/kleisauke/sharp/actions/runs/23295199738/job/67740759490#step:10:24

kleisauke:
Only 1 test failure left after commit 99803d6bc7bc995f0f8fe7f3e0f03eaf4bdbaa70:
https://github.com/kleisauke/sharp/actions/runs/23297678823/job/67749563497#step:10:2086

toyobayashi:
fixed in emnapi 1.9.1

lovell:
@kleisauke Thanks for investigating this and @toyobayashi thanks for the quick fix.

The remaining sharp test failure appears to relate to the new `toUint8Array()` method, which returns a transferable `ArrayBuffer` rather than a non-transferable `SharedArrayBuffer`.

I think the WebAssembly equivalent is the `shared` option passed to `new WebAssembly.Memory({ shared: true })`, which maybe emnapi always uses? Perhaps `toUint8Array` should fall back to the `toBuffer` behaviour when run as WebAssembly?

kleisauke:
From a quick look, an `emnapi_sync_memory()` call may be missing, which could be the reason for the `toUint8Array()` test failure.
https://toyobayashi.github.io/emnapi-docs/reference/list.html#arraybuffer-related

I'll investigate this later, but in the meantime I'm marking this PR as ready.

lovell:
@kleisauke I fixed the test failure via commit https://github.com/lovell/sharp/commit/3879c9a7ef0d2eb19481d302db28ea2761444ca9 (this was also affecting Electron).

Issue #4355, opened 2025-03-28
Make sharp().toBuffer() return Buffers with transferrable backing stores


Rush:
### What are you trying to achieve?

I'm working in a multi-threaded Node.js application where `sharp().toBuffer()` output is passed to `worker_threads` via `postMessage`. To avoid copying overhead, I'd like to transfer the backing `ArrayBuffer` (i.e., use zero-copy transfers).

However, the current `Buffer` returned by `toBuffer()` may point to a shared or larger memory pool (via `node::Buffer::New(...)` with internal backing), making the `.buffer` potentially non-transferrable or containing unrelated memory slices.

Specifically you can test this by calling `require('worker_threads'). isMarkedAsUntransferable(buffer)` which today returns `true`.

### When you searched for similar feature requests, what did you find that might be related?

Didn't find anything relevant to transferrable ArrayBuffer as a backing store.

### What would you expect the API to look like?

Ideally, the existing `toBuffer()` call would already return a `Buffer` backed by a tightly sliced `ArrayBuffer` that is safe to transfer.

If not feasible as the default, a new option like:

```ts
sharp(...).toBuffer({ transferrable: true });

**Comment by lovell** [comment]
Images returned by sharp as `Buffer` objects do not come from the shared pool so should already be transferable.

https://github.com/nodejs/node/issues/55593 suggests this might be a recent Node.js change/feature/bug.

**Comment by Rush** [comment]
Can you point me to the place where those buffers are created in sharp? Also looked at this comment which may indicate that there is a pooling behavior: https://github.com/nodejs/node/issues/55593#issuecomment-2500316828

**Comment by lovell** [comment]
libvips (or one of its dependencies) allocates the memory, sharp uses [NewOrCopy](https://github.com/nodejs/node-addon-api/blob/main/doc/buffer.md#neworcopy) to wrap it as a `Buffer` instance.

https://github.com/lovell/sharp/blob/031c808aa5f8763e1d95ea0b7c1e462714501834/src/pipeline.cc#L1355

For Node.js this is a zero-copy operation. For recent versions of Electron that enable its memory cage feature, this involves a copy, which I guess might potentially involve an allocation from the pool.

**Comment by Rush** [comment]
Thanks for sharing the code sample. While it may be zero-copy for Node.js, the resulting buffer is still marked as untransferable.

To address this without introducing a copy, we could allocate the memory ourselves and wrap it with a `shared_ptr`:

```
[...]

PR #4520, opened 2026-04-10
types: return more precise `Buffer<ArrayBuffer>` from `toBuffer`

Andarist:


package.json
```diff
@@ -169,6 +169,11 @@
     "@img/sharp-win32-ia32": "0.34.5",
     "@img/sharp-win32-x64": "0.34.5"
   },
+  "peerDependenciesMeta": {
+    "@types/node": {
+      "optional": true
+    }
+  },
```
Andarist, line 176: I think this is kinda more compatible with strict package managers like pnpm. If you'd like to declare an explicit range for compatible `@types/node` then you'd have to declare the range within a regular `peerDependencies` declaration and, in here, the correct range would be something like:
```
^18.19.54 || ^20.16.10 || >=22.7.4
```

Those are the versions published on 2024-09-27 and they introduced this change: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/70694

lovell:
Hi Mateusz, thanks for the PR.

If you hadn't seen, the next 0.35.0 release of sharp will also allow output via a new `toUint8Array()` to complement the existing `toBuffer()` - see https://github.com/lovell/sharp/issues/4355

```ts
toUint8Array(): Promise<{ data: Uint8Array; info: OutputInfo }>;
```

My understanding is that Node.js `Buffer` instances allocated by native code are treated as being backed by a `SharedArrayBuffer` and are considered non-transferable whereas `Uint8Array` instances are always backed by an `ArrayBuffer` and are transferable (e.g. between worker threads).

I *think* this means that that existing `toBuffer()` will resolve with a `Buffer<SharedArrayBuffer>` - is this something you're able to check?

lovell:
@Andarist Hej Mateusz, is this something you're still interested in working on? No worries if not, I'm happy to take it on.

lovell:
Posting a possible diff here for my future self to use as this might be a breaking change:
```diff
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -667,7 +667,7 @@ declare namespace sharp {
          * @param options.resolveWithObject Resolve the Promise with an Object containing data and info properties instead of resolving only with data.
          * @returns A promise that resolves with the Buffer data.
          */
-        toBuffer(options?: { resolveWithObject: false }): Promise<Buffer>;
+        toBuffer(options?: { resolveWithObject: false }): Promise<Buffer<SharedArrayBuffer>>;
 
         /**
          * Write output to a Buffer. JPEG, PNG, WebP, AVIF, TIFF, GIF and RAW output are supported.
@@ -676,14 +676,14 @@ declare namespace sharp {
          * @param options.resolveWithObject Resolve the Promise with an Object containing data and info properties instead of resolving only with data.
          * @returns A promise that resolves with an object containing the Buffer data and an info object containing the output image format, size (bytes), width, height and channels
          */
-        toBuffer(options: { resolveWithObject: true }): Promise<{ data: Buffer; info: OutputInfo }>;
+        toBuffer(options: { resolveWithObject: true }): Promise<{ data: Buffer<SharedArrayBuffer>; info: OutputInfo }>;
 
         /**
          * Write output to a Uint8Array backed by a transferable ArrayBuffer. JPEG, PNG, WebP, AVIF, TIFF, GIF and RAW output are supported.
          * By default, the format will match the input image, except SVG input which becomes PNG output.
          * @returns A promise that resolves with an object containing the Uint8Array data and an info object containing the output image format, size (bytes), width, height and channels
          */
-        toUint8Array(): Promise<{ data: Uint8Array; info: OutputInfo }>;
+        toUint8Array(): Promise<{ data: Uint8Array<ArrayBuffer>; info: OutputInfo }>;
 
         /**
          * Set output density (DPI) in EXIF metadata.
```

Andarist:
I tested this with:
```js
const sharp = require("sharp");
const {
  MessageChannel,
  isMarkedAsUntransferable,
} = require("node:worker_threads");

(async () => {
  for (const mode of ["toBuffer", "toUint8Array"]) {
    const image = sharp({
      create: {
        width: 1,
        height: 1,
        channels: 3,
        background: "red",
      },
    }).png();

    const value =
      mode === "toBuffer"
        ? await image.toBuffer()
        : (await image.toUint8Array()).data;

    console.log(`\n${mode}`);
    console.log("sharp:", sharp.versions.sharp);
    console.log("node:", process.version);
    console.log("Buffer.isBuffer:", Buffer.isBuffer(value));
    console.log("value.constructor.name:", value.constructor.name);
    console.log(
      "value.buffer.constructor.name:",
      value.buffer.constructor.name,
    );
    console.log(
      "value.buffer instanceof ArrayBuffer:",
      value.buffer instanceof ArrayBuffer,
    );
    console.log(
      "value.buffer instanceof SharedArrayBuffer:",
      value.buffer instanceof SharedArrayBuffer,
    );
    console.log(
      "isMarkedAsUntransferable(value.buffer):",
      isMarkedAsUntransferable(value.buffer),
    );

    const { port2 } = new MessageChannel();

    try {
      port2.postMessage(value.buffer, [value.buffer]);
      console.log("transfer: ok");
    } catch (err) {
      console.log("transfer:", err.name, err.message);
    }
  }
})();

```

and got this output:
```
  toBuffer
  Buffer.isBuffer: true
  value.constructor.name: Buffer
  value.buffer.constructor.name: ArrayBuffer
  value.buffer instanceof ArrayBuffer: true
  value.buffer instanceof SharedArrayBuffer: false
  isMarkedAsUntransferable(value.buffer): true
  transfer: DataCloneError Cannot transfer object of unsupported type.

  toUint8Array
  Buffer.isBuffer: true
  value.constructor.name: Buffer
  value.buffer.constructor.name: ArrayBuffer
  value.buffer instanceof ArrayBuffer: true
  value.buffer instanceof SharedArrayBuffer: false
  isMarkedAsUntransferable(value.buffer): false
  transfer: ok
```

From what I can tell, none of those return `SharedArrayBuffer` here.

lovell:
Brilliant, thanks for the update Mateusz.

Although `ArrayBuffer` instances are supposed to be transferable ([MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer)), it looks like a Node.js `Buffer` instance wraps an `ArrayBuffer` that is marked as non-transferable. Given the outcome of your research this behaviour feels a bit like a bug, and there's some discussion about this upstream at https://github.com/nodejs/node/issues/55593 so we're not alone.

Node-API added an experimental [node_api_create_external_sharedarraybuffer](https://nodejs.org/docs/latest/api/n-api.html#node_api_create_external_sharedarraybuffer) method from Node.js 26.1.0 but we can't use that yet, and it's unclear if we need to anyway.

In summary I think this means that `Buffer<ArrayBuffer>` is correct in theory but a bit broken in practice.

lovell:
This feels like a documentation thing now so let's merge and publish. Thanks again for taking the time to make a PR.

PR #4527, opened 2026-04-27
ci: verify Node.js toolchain downloads in build-qemu

safeengineauto:
## Summary

This PR hardens the `build-qemu` workflow by verifying downloaded Node.js toolchain tarballs before extraction.

Currently the job downloads the Node.js tarball and pipes it directly into `tar`. This change downloads the tarball to disk, verifies it against the expected SHA-256 from upstream `SHASUMS256.txt`, and only extracts it after verification succeeds.

This affects the four QEMU-built platform packages:

- `linux-arm`
- `linux-s390x`
- `linux-ppc64`
- `linux-riscv64`

This is release-chain hardening, not a fix for a known vulnerability.

## Changes

- Add `nodejs_sha256` to each `build-qemu` matrix entry.
- Download the Node.js tarball to `/tmp`.
- Verify it with `sha256sum -c`.
- Extract only after verification succeeds.
- Use `curl --show-error --fail --location`.

## Notes

The SHA-256 values were taken from the corresponding upstream `SHASUMS256.txt` files for the exact Node.js versions already used by the workflow.

I kept the change limited to `build-qemu` and did not change dependency installation, action pinning, base images, or release logic.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -239,18 +239,21 @@ jobs:
             nodejs_arch: armv6l
             nodejs_hostname: unofficial-builds.nodejs.org
             nodejs_version: "18.17.0"
+            nodejs_sha256: "1a8ef740b5b7a1bc39591f562363537d3fb5595a451dfc6cb29797aa1412aeea"
             nodejs_version_major: 18
           - platform: linux-s390x
             base_image: "--platform=linux/s390x s390x/debian:bookworm"
             nodejs_arch: s390x
             nodejs_hostname: nodejs.org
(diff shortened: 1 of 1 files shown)


lovell:
Thanks for the PR, I'm unsure this added complexity is necessary.

I'm aware checksums are often used by OS-level package managers when dependencies might be distributed via non-secure HTTP, however we're always fetching Node.js binaries via HTTPS so the risk of in-flight tampering is much lower. Perhaps we can experiment with OS-provided Node.js packages when running under QEMU emulation?

PR #4529, opened 2026-05-07
Docs: bump astro to latest version 6.3.1

deining:
This PR bumps astro used to generate docuemntation to latest version 6.3.1.

### DIFF
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -74,7 +74,7 @@ export default defineConfig({
         {
           label: 'Changelog',
           collapsed: true,
-          autogenerate: { directory: 'changelog' }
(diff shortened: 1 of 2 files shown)


lovell:
Thank you for taking the time to make a PR for this. If you hadn't seen, the work for the next release is on the `0.35` branch and Astro was updated via commit https://github.com/lovell/sharp/commit/16725561d5895fe6bc0464ba3ac520b189d9be72, although I note there's a new version of `starlight-auto-sidebar` since then.

lovell:
I've now updated the Starlight package versions via commit https://github.com/lovell/sharp/commit/1e76faebad671f02a09db727b1570158c62affa2, thanks for the nudge.

PR #4532, opened 2026-05-21
Update minimum glibc and musl requirements

kleisauke:


lovell:
Thanks Kleis, I'll update the sharp-libvips packages too.

PR #4534, opened 2026-06-03
refactor: improve build error diagnostics in libvips.js - I would be glad to become your contributor

MatrixNeoKozak:
### Summary of changes
Modified the `spawnRebuild` function in `lib/libvips.js` to enhance error reporting when the `node-gyp` build process fails. Previously, the error output relied solely on `stdio: 'inherit'`. While functional, explicitly capturing stderr allows for better logging of build failures and improves the diagnostic capability during installation issues.

### Rationale
In scenarios where `node-gyp` fails, simply relying on inherited stdio can sometimes obscure the root cause, especially in CI environments or restricted terminal outputs. Explicitly capturing and logging the error makes it easier to troubleshoot native module compilation issues.

### Technical impact analysis
- The change improves the developer experience by providing clearer error messages during failed native module builds.
- There is no functional change to the build process logic itself, ensuring backward compatibility with existing build workflows.

### DIFF
--- a/lib/libvips.js
+++ b/lib/libvips.js
@@ -123,11 +123,16 @@ const yarnLocator = () => {
 
 /* node:coverage disable */
 
-const spawnRebuild = () =>
-  spawnSync(`node-gyp rebuild --directory=src ${isEmscripten() ? '--nodedir=emscripten' : ''}`, {
+const spawnRebuild = () => {
+  const result = spawnSync(`node-gyp rebuild --directory=src ${isEmscripten() ? '--nodedir=emscripten' : ''}`, {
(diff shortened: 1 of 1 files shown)


lovell:
Hi, thanks for the PR, however I'm unsure this change is necessary as the output should already be displayed. If you have a scenario where this isn't the case then please provide steps to reproduce.

In addition this PR is clearly LLM-generated, for example the `newman.tech` domain name does not exist.

https://github.com/MatrixNeoKozak/sharp/commit/5b6bafb3dad630db59b96d06f4a20be3bf20ba85.patch

PR #4535, opened 2026-06-07
Remove heif bitdepth property guard

kleisauke:
See https://github.com/lovell/sharp-libvips/issues/247

### DIFF
--- a/docs/src/content/docs/install.md
+++ b/docs/src/content/docs/install.md
@@ -55,7 +55,7 @@ Ready-compiled sharp and libvips binaries are provided for use on the most commo
 * FreeBSD (WebAssembly)
 
 This provides support for the
-JPEG, PNG, WebP, AVIF (limited to 8-bit depth), TIFF, GIF and SVG (input) image formats.
+JPEG, PNG, Ultra HDR, WebP, AVIF, TIFF, GIF and SVG (input) image formats.
 
 ## Cross-platform
 
--- a/lib/output.mjs
+++ b/lib/output.mjs
@@ -1206,7 +1206,6 @@ function tiff (options) {
  * Use these AVIF options for output image.
  *
  * AVIF image sequences are not supported.
- * Prebuilt binaries support a bitdepth of 8 only.
  *
  * When using Windows ARM64, this feature requires a CPU with ARM64v8.4 or later.
  *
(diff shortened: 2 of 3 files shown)


lovell:
Dankjewel!

kleisauke:
As possible follow-up, we could default to 12-bit save for 16-bit images:
```diff
--- a/src/pipeline.cc
+++ b/src/pipeline.cc
@@ -1113,7 +1113,7 @@ class PipelineWorker : public Napi::AsyncWorker {
             ->set("Q", baton->heifQuality)
             ->set("compression", baton->heifCompression)
             ->set("effort", baton->heifEffort)
-            ->set("bitdepth", baton->heifBitdepth)
+            ->set("bitdepth", sharp::Is16Bit(image.interpretation()) ? 12 : baton->heifBitdepth)
             ->set("tune", baton->heifTune.c_str())
             ->set("subsample_mode", baton->heifChromaSubsampling == "4:4:4"
               ? VIPS_FOREIGN_SUBSAMPLE_OFF : VIPS_FOREIGN_SUBSAMPLE_ON)
@@ -1321,7 +1321,7 @@ class PipelineWorker : public Napi::AsyncWorker {
             ->set("Q", baton->heifQuality)
             ->set("compression", baton->heifCompression)
             ->set("effort", baton->heifEffort)
-            ->set("bitdepth", baton->heifBitdepth)
+            ->set("bitdepth", sharp::Is16Bit(image.interpretation()) ? 12 : baton->heifBitdepth)
             ->set("tune", baton->heifTune.c_str())
             ->set("subsample_mode", baton->heifChromaSubsampling == "4:4:4"
               ? VIPS_FOREIGN_SUBSAMPLE_OFF : VIPS_FOREIGN_SUBSAMPLE_ON)
```

i.e. similar to:
https://github.com/lovell/sharp/blob/e4ea2f35eb7194ec7ade46de64bf2a9904c07127/src/pipeline.cc#L1027
and:
https://github.com/libvips/libvips/blob/v8.18.3-rc1/libvips/foreign/heifsave.c#L534-L541
https://github.com/libvips/libvips/blob/v8.18.3-rc1/libvips/foreign/pngsave.c#L113-L120

That said, we may want to revisit this, since it effectively overrides the `bitdepth` option.

lovell:
Yeah, it's probably time to overhaul some of the colourspace-related internals and use one of sRGB or scRGB then combine/replace the slightly cumbersome `pipelineColourspace` and `toColourspace` functions into a single, easier to understand "use wide gamut" vs "use narrow gamut" setting.

Perhaps something as simple as `sharp(input, { hdr: true })...`, which would switch to scRGB, try to preserve gainmaps, save at high bit depths when format allows etc.

PR #4536, opened 2026-06-09
type-check density option before range validation

metsw24-max:
**Missing type guard on the density option**
Unlike every other numeric input option, `density` is range-checked with `is.inRange` but never type-checked, so a numeric string or single-element array (e.g. `'50'` or `[50]`) slips through to the native layer instead of being rejected. Added an `is.number` guard to match the sibling checks and the error message that already promises a number.


lovell:
Thank you!

PR #4542, opened 2026-06-11
ESM-only

### LINKED ISSUE #4539: ESM-only

Reduce the package size by switching to ESM-only.

Followup to #2981.

I think support for CJS projects is probably a must for a project like Sharp, so it can't just be migrated to ESM-only because of dynamic imports. So it would need a change in the API (breaking change), like `sharp = await loadSharp()`.

Keep ESM+CJS. Problem: increase in size, files are duplicated.

florian-lefebvre:
Closes #4539


florian-lefebvre:
Tests pass locally for me. @lovell can you run the workflows?

florian-lefebvre:
@lovell I think I fixed everything (coverage and node 20), should be good to go

florian-lefebvre:
I just tried running the failing tests locally on Windows 11 and can't replicate, so _I think_ that they should pass now that lint is fixed, but no guarantee :/

florian-lefebvre:
I think the issue was around the font file so I tried another approach of having the font file directly in the repo. I tried it locally on windows 11 and it worked but again, no guarantee. If that doesn't work, I'm out of ideas

florian-lefebvre:
It worked :tada:

lib/colour.js
```diff
@@ -4,7 +4,7 @@
 */
 
 import color from '@img/colour';
```
Norbiros, line 6: `@img/colour` is essentially the CommonJS version of the `color` package:

> This package converts the color package and its dependencies, all of which are MIT-licensed, to CommonJS.

Since `sharp` is now ESM-only, this seems like a good time to migrate to `color` directly.

florian-lefebvre: I'll let Lovell decide but I think it makes sense

Norbiros: For context I also opened https://github.com/e18e/module-replacements/issues/786 about it

Norbiros:
We could also replace `semver` with `verkit` - https://github.com/e18e/ecosystem-issues/issues/277

gameroman:
Looks like conflicts need to be resolved

florian-lefebvre:
Yeah I resolve conflicts from time to time, I don't do it too often since there's no ETA

PR #4544, opened 2026-06-13
type-check gif effort as integer before range validation

metsw24-max:
**Non-integer effort accepted by gif output**
The gif `effort` option is guarded with `is.number`, so a fractional value such as `7.5` passes the range check and is then silently truncated when read as a uint32 by the native layer, unlike the png, webp, heif and jxl effort options which all use `is.integer`. Swapped in `is.integer` so the guard matches the documented contract and the existing error message, which both promise an integer.

### DIFF
--- a/lib/output.mjs
+++ b/lib/output.mjs
@@ -929,7 +929,7 @@ function gif (options) {
       }
     }
     if (is.defined(options.effort)) {
-      if (is.number(options.effort) && is.inRange(options.effort, 1, 10)) {
+      if (is.integer(options.effort) && is.inRange(options.effort, 1, 10)) {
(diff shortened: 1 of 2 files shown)


lovell:
It looks like the biome linter has just added SVG support but is slightly broken so I've switched this bit off for now via commit https://github.com/lovell/sharp/commit/bda0007b97019f2c856ab926946971d812683cbf - please can you rebase.

lovell:
Landed via commit https://github.com/lovell/sharp/commit/fa2c9efe1dffcfada588a3ed0c30fec1ef6bfa68, thank you very much for fixing this.

PR #4545, opened 2026-06-15
validate recomb matrix entries are numbers

metsw24-max:
**Inconsistent type checking in recomb**
The recombination matrix is flattened and passed through `map(Number)`, so non-numeric entries are silently coerced rather than rejected: a stray string becomes `NaN` while `true` and `null` turn into `1` and `0` before they reach the native matrix. The sibling `affine` and `linear` operations already guard their input with `is.number`, so this should keep the behaviour consistent and surface the mistake to the caller instead of producing a quietly wrong result.


lovell:
Thank you!

Issue #4442, opened 2025-08-21
Please explain why you are rebuilding wasm in postinstall.


huuyafwww:
Hello!
Thank you for developing the wonderful Sharp library.  

I am currently using pnpm in my project, and while checking the lifecycle script,
I need some information to decide whether to allow Sharp's postinstall.
I have a question about one point.

I think sharp uses node-gyp during installation to rebuild the C# main body into wasm format using emscripten.

https://github.com/lovell/sharp/blob/8607ff2f4a5d0f523f57c589378256a626b62c97/package.json#L95
https://github.com/lovell/sharp/blob/8607ff2f4a5d0f523f57c589378256a626b62c97/install/check.js#L27
https://github.com/lovell/sharp/blob/8607ff2f4a5d0f523f57c589378256a626b62c97/lib/libvips.js#L126

This means rebuilding all WASM files for each cross-platform. Is that correct?
Strictly speaking, I understand that it regenerates wasm corresponding to all packages under the src directory.
Furthermore, if that understanding is correct, there is a question.

For example, let's look at @img/sharp-darwin-arm64.
If you check the npm code, you will see that it contains lib/sharp-darwin-arm64.node.
My understanding is that the wasm output is included in the package.
https://www.npmjs.com/package/@img/sharp-darwin-arm64?activeTab=code

However, I don't understand why it needs to be rebuilt again during installation. Could you please explain?🙇‍♂️
This is just my guess, but I think they are rebuilding to generate the optimal wasm for each execution environment's runtime. Is that correct?

lovell:
> This means rebuilding all WASM files for each cross-platform. Is that correct?

No. 

> ...using pnpm in my project, and while checking the lifecycle script...

You are unlikely to require the use of lifecycle scripts. The documentation suggests using `ignoredBuiltDependencies`, please see https://sharp.pixelplumbing.com/install/

The slightly longer explanation is that if and only if prebuilt binaries are not available for the current platform and if and only if an emscripten development environment is detected then sharp will attempt to build WebAssembly binaries. This is used by sharp's CI environment and occasionally for local development. No end users of sharp are ever expected to be in this situation.

huuyafwww:
@lovell 
Thank you for your prompt reply!
I understand that it was prepared for use by sharp developers.
In that case, we will add it to ignoredBuiltDependencies and respond accordingly.
Thank you very much!

tats-u:
[...]

PR #4547, opened 2026-06-16
docs: update `install` life cycle script removal

hyunbinseo:
Continuing from https://github.com/lovell/sharp/issues/4442#issuecomment-4717894434

## Assumptions

Users on `0.34` and below are facing a major update to `0.35` and above.

- Not all users know about life cycle scripts and how it works.
- Not all users know that libraries are shipped with prebuilt binaries.

## Goals

Make the changelog more actionable.

Guide users that building from source is possible but probably not needed. (as stated in the commit message)

Also, "now opt-in" is misleading because it gives an impression that compiling from source was automatic and mandatory.

If I understood correctly, it automatically compiled only if a compatible binary was not found:

> The slightly longer explanation is that if and only if prebuilt binaries are not available for the current platform and if and only if an emscripten development environment is detected then sharp will attempt to build WebAssembly binaries.

## Additional Suggestions

- Quote the actual error messages shown when package managers cannot resolve a prebuilt binary.
- Add a note on removing `sharp` from `onlyBuiltDependencies` or `allowBuilds`. This could help if a malicious update is released with life cycle scripts.

### DIFF
--- a/docs/src/content/docs/changelog/v0.35.0.md
+++ b/docs/src/content/docs/changelog/v0.35.0.md
@@ -5,8 +5,8 @@ slug: changelog/v0.35.0
 
 * Breaking: Drop support for Node.js 18, now requires Node.js >= 20.9.0.
 
-* Breaking: Remove `install` script from `package.json` file.
-  Compiling from source is now opt-in via the `build` script.
+* Breaking: Remove `install` script.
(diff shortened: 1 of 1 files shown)


lovell:
Landed via commit https://github.com/lovell/sharp/commit/188db4ea313d58ea6ec8606029c92ffdb55f7a82 (with additional mention of WebAssembly), thank you very much for this improvement.

PR #4548, opened 2026-06-17
bound dilate and erode width to avoid mask-size overflow

metsw24-max:
**Unbounded dilate/erode width overflows the mask size**

`dilate()` and `erode()` accept any positive integer, unlike the sibling `median()` which bounds its window to 1 to 1000, so an oversized width reaches the native `Dilate`/`Erode` where the structuring element is sized as `2 * width + 1` and passed to `new_matrix`; read through `AttrAsUint32` into an `int`, a width of `2^30` wraps that signed size negative and `2^32` narrows it to zero, silently dropping the operation. Bounded the width to match `median` so the value can no longer reach that calculation.

### DIFF
--- a/lib/operation.mjs
+++ b/lib/operation.mjs
@@ -424,10 +424,10 @@ function blur (options) {
 function dilate (width) {
   if (!is.defined(width)) {
     this.options.dilateWidth = 1;
-  } else if (is.integer(width) && is.inRange(width, 1, 1000)) {
+  } else if (is.integer(width) && is.inRange(width, 1, 65536)) {
     this.options.dilateWidth = width;
   } else {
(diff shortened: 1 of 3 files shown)


lovell:
Thanks for the PR to fix this. Is there a reason for the 1000 limit? I think it might need to be a bit higher than this, perhaps 65536 (1 << 16)?

metsw24-max:
No strong reason for 1000, I just matched the bound `median()` uses for its window. 65536 is fine here. The only value that needs guarding is the `2 * width + 1` mask size, and at 65536 that comes to 131073, still well within `int`. Pushed a change bounding both to 1..65536.

lovell:
Thank you

PR #4549, opened 2026-06-18
validate convolve kernel values are numbers

metsw24-max:
convolve validates the kernel width, height and that the flattened kernel array has the expected length, but never checks the entries themselves are numbers, unlike the sibling affine, linear and recomb operators which all guard their input with is.number. a NaN entry therefore passes straight through to the native convolution mask and produces a silently wrong (all-zero) result, whilst a string or null entry surfaces only the vague "A number was expected" from the native layer rather than the usual invalid-parameter error. the safer behaviour is to reject a malformed kernel up front, so this adds the same per-entry is.number check to the existing kernel validation, keeping the contract consistent with its siblings. the added tests cover the string and NaN cases, both of which slip through today.


lovell:
Thanks again!

PR #4550, opened 2026-06-19
feat: prepare esm only

florian-lefebvre:
In #4542 I had to do several things unrelated to the ESM migration:

- I discovered that some things are not actually covered by tests so I had to add some
- Fonts seem to have issues

In order to make reviewing said PR easier, I extracted these changes here to keep the other really focused.

I ran tests on Ubuntu and they worked


test/fixtures/index.js
```diff
@@ -15,7 +15,7 @@ const getPath = (filename) => path.join(__dirname, filename);
 async function fingerprint (image) {
   return sharp(image)
     .flatten('gray')
-    .greyscale()
+    .grayscale()
```
lovell, line 18: Is this change because there are no other tests for the `grayscale()` function and we needed path coverage? We should probably add something explicit to the colour-related tests. I can do this in a follow-up.

lovell (approved):
Thank you very much for doing this, the font-related tests on Windows have been a bit flaky for a while due to font discovery problems (timeouts, wrong font selected etc.).

test/fixtures/index.js
florian-lefebvre: Yes that's why :+1:

PR #4551, opened 2026-06-20
bound clahe width and height to avoid signed overflow

metsw24-max:
clahe() bounds width and height with only is.integer and a greater-than-zero check, unlike the sibling window operators median and dilate/erode which cap their size, so an oversized value reaches the native layer where claheWidth/claheHeight are read through AttrAsUint32 into an int and handed to hist_local. a width of 2^31 narrows that signed value negative and trips a GObject-CRITICAL warning, while 2^32 narrows to zero and the native guard then silently drops the equalisation that was requested. bound both to 1..65536 to match dilate and erode so the value can no longer reach that narrowing; the small search windows shown in the docs are unaffected.

### DIFF
--- a/lib/operation.mjs
+++ b/lib/operation.mjs
@@ -654,15 +654,15 @@ function normalize (options) {
  */
 function clahe (options) {
   if (is.plainObject(options)) {
-    if (is.integer(options.width) && options.width > 0) {
+    if (is.integer(options.width) && is.inRange(options.width, 1, 65536)) {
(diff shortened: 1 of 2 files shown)


lovell:
Thanks again. I've tightened up a few more options via commit https://github.com/lovell/sharp/commit/2bc2f3270ece597ae2fc5a2bc26b0af23f50d865 - keep the PRs coming if you find more.

PR #4552, opened 2026-06-22
bound trim margin to avoid signed overflow in extract size

metsw24-max:
**Unbounded trim margin overflows the extract size**

The `margin` option is only checked with `is.integer` and `>= 0`, so it lacks the upper bound the sibling pixel dimensions now carry. It is read via `AttrAsUint32` and used in `Trim()` as `width + 2 * margin`, so a margin around 2^30 overflows the signed int into a negative extract width (a GLib-GObject-CRITICAL plus an `extract_area` failure), while a margin around 2^31 narrows to a negative int and the native `margin > 0` guard then silently drops it. I have bounded it to 0..10000000 (VIPS_MAX_COORD), which keeps the existing 9999999 clamp test green and should make the limit easier to reason about alongside the other dimensions.


lovell:
Thank you

PR #4553, opened 2026-06-23
reject non-finite numbers in is.number

metsw24-max:
**Non-finite values treated as numbers**

`is.number` only excluded `NaN`, so `Infinity` and `-Infinity` passed as valid numbers and reached the native layer. The clearest symptom is `affine`: a non-finite `idx`/`idy`/`odx`/`ody` offset is set on the libvips `vips_affine` gdouble property, which rejects it with a `GLib-GObject-CRITICAL` and silently leaves the offset at 0, so the requested transform is quietly wrong rather than reported as invalid. Switching the check to `Number.isFinite` rejects these up front, which keeps the behaviour consistent for every numeric option that relies on the helper.

### DIFF
--- a/lib/is.mjs
+++ b/lib/is.mjs
@@ -78,7 +78,7 @@ const string = (val) => typeof val === 'string' && val.length > 0;
  * Is this value a real number?
  * @private
  */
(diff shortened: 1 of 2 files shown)


lovell:
Good spot, thank you.

PR #4555, opened 2026-06-23
bound extract region to libvips coordinate limit

metsw24-max:
extract() validates left, top, width and height with is.integer(value) && value >= 0 but never caps them, so a coordinate above libvips' VIPS_MAX_COORD (100000000) reaches vips_extract_area and is rejected by the gint gobject property with a GLib-GObject-CRITICAL, surfacing only a misleading 'extract_area: parameter width not set' instead of a parameter error; a value of 2^31 narrows to a negative int the same way. the sibling pixel-dimension options (extend, trim margin, clahe, dilate, erode) were already bounded for this reason, so this brings extract into line by clamping each of the four to 0..100000000 in the validation layer. values up to the limit still pass through to vips unchanged, so a genuine out-of-image request keeps its existing 'bad extract area' behaviour.

### DIFF
--- a/docs/src/content/docs/api-resize.md
+++ b/docs/src/content/docs/api-resize.md
@@ -236,10 +236,10 @@ Extract/crop a region of the image.
 | Param | Type | Description |
 | --- | --- | --- |
 | options | <code>Object</code> | describes the region to extract using integral pixel values |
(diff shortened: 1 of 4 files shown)


lovell:
Thank you

PR #4556, opened 2026-06-24
validate background colour values are numbers

metsw24-max:
**Non-numeric colour components reach libvips unchecked**

A background object with a non-numeric component such as `{ r: 'x', g: 0, b: 0 }` is parsed into an RGBA array where that channel becomes `NaN`, which `_getBackgroundColourOption` returns without checking, so it reaches the native layer and the result is silently wrong rather than rejected. The check sits in this shared helper because every background option (tint, flatten, extend, rotate, create, join, etc.) routes through it, matching the strictness the direct numeric options already apply.

### DIFF
--- a/lib/colour.mjs
+++ b/lib/colour.mjs
@@ -146,12 +146,16 @@ function _getBackgroundColourOption (value) {
     (is.string(value) && value.length >= 3 && value.length <= 200)
   ) {
     const colour = color(value);
-    return [
+    const rgba = [
       colour.red(),
       colour.green(),
       colour.blue(),
       Math.round(colour.alpha() * 255)
     ];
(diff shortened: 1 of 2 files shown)


lovell:
Hello again, thanks for the PR. If we're going to verify colour values then I think we should do so separately for each channel to improve error messaging as `received [object Object] of type object` doesn't provide enough information for someone to fix the problem.

metsw24-max:
Good point. I've switched it to validate each channel separately, so a bad component now names the channel, e.g. `Expected number for background.blue but received NaN of type number` instead of the unhelpful `[object Object]` message. Test updated to cover a couple of different channels.

lovell:
Thanks for the update

PR #4558, opened 2026-06-24
bound create and raw input dimensions to coordinate limit

metsw24-max:
**Unbounded create and raw input dimensions**

`create.width`/`height` and `raw.width`/`height` in `_createInputDescriptor` were validated only as `> 0`, so a dimension above the libvips coordinate limit reaches the `gint` width/height property and is silently dropped: 100000001 leaves a `GLib-GObject-CRITICAL` and a 1px fallback image, and 2147483648 narrows to a negative size. Bounded all four to 1-100000000 to match the `text.width`/`text.height` check already in the same function, so an oversized dimension now fails validation up front rather than producing a wrong-sized result. This also keeps the input-descriptor checks consistent with the other dimension options.

### DIFF
--- a/lib/input.mjs
+++ b/lib/input.mjs
@@ -181,8 +181,8 @@ function _createInputDescriptor (input, inputOptions, containerOptions) {
     if (is.defined(inputOptions.raw)) {
       if (
         is.object(inputOptions.raw) &&
-        is.integer(inputOptions.raw.width) && inputOptions.raw.width > 0 &&
-        is.integer(inputOptions.raw.height) && inputOptions.raw.height > 0 &&
+        is.integer(inputOptions.raw.width) && is.inRange(inputOptions.raw.width, 1, 100000000) &&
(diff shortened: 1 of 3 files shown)


lovell:
Thanks again for the PR. Happy to accept multiple fixes at once if known and it's easier.

Given it's starting to appear in a few places, do you think it might be worth defining the `100000000` coordinate value in a single location? This value is configurable upstream at start-up time in libvips so maybe do so in `lib/utility.mjs` so we can (in a separate change, in the future) make this a bit more dynamic.

lovell:
I'll need to think a bit more about how we expose getting/setting the upstream coordinate limit so let's merge and release this useful fix, thanks again.

metsw24-max:
Agreed, pulling the 100000000 into lib/utility.mjs makes sense, especially with a view to reading it back from libvips later so it stays in sync. Happy to put that follow-up together once you've settled on how you'd like the get/set exposed, so it can land in one go rather than piecemeal.

PR #4560, opened 2026-06-25
Reject ragged matrices in recomb and affine

chatman-media:
`recomb()` and `affine()` flatten the supplied matrix and validate only the *total* element count. A non-rectangular nested array whose values happen to flatten to a valid cardinality is therefore accepted and applied with the wrong shape instead of being rejected:

```js
sharp(input).recomb([[1, 2, 3, 4, 5], [6, 7], [8, 9]]);   // flat length 9 -> silently treated as 3x3
sharp(input).affine([[1, 2, 3], [4]]);                    // flat length 4 -> silently treated as 2x2
```

This validates that every row is an array of the expected length before flattening, so malformed matrices throw `Invalid parameter` as documented ("3x3 or 4x4" / "1x4 or 2x2"). The existing `recomb` "incorrect sub size" test already established the intent to reject non-rectangular matrices; this closes the gap where the row lengths cancel out.

Added unit tests for both that fail without the change. `npm run lint-js` (biome) clean.

### DIFF
--- a/lib/operation.mjs
+++ b/lib/operation.mjs
@@ -178,7 +178,13 @@ function flop (flop) {
  * @throws {Error} Invalid parameters
  */
 function affine (matrix, options) {
-  const flatMatrix = Array.isArray(matrix) ? [].concat(...matrix) : [];
+  const isValidShape = Array.isArray(matrix) && (
+    // 1x4 array of numbers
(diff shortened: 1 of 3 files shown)


lovell:
Thank you!

PR #4561, opened 2026-06-25
avoid signed overflow in cache memory limit

metsw24-max:
sharp.cache({ memory }) passes the megabyte value through to the native cache helper, which reads it as a 32-bit int and multiplies by 1048576 to convert to bytes before handing the result to vips_cache_set_max_mem. The multiply runs in 32-bit signed arithmetic and only widens to the size_t the function expects afterwards, so any value of 2048 or above overflows first. A 2048MB request wraps to a negative number that sign-extends to roughly 16 exabytes, which leaves the operation cache effectively unbounded, while exactly 4096MB wraps to zero and silently disables it. I came across this while confirming that larger cache limits were honoured and saw the reported memory.max had no relation to the value passed in. Casting the operand to size_t before the multiply keeps the conversion in 64-bit so the byte count stays correct across the documented range. I have limited the change to that one expression and added a regression test at 4096MB, which reported a maximum of zero beforehand.

### DIFF
--- a/lib/utility.mjs
+++ b/lib/utility.mjs
@@ -114,6 +114,12 @@ function cache (options) {
       return sharp.cache(0, 0, 0);
     }
   } else if (is.object(options)) {
+    for (const property of ['memory', 'files', 'items']) {
+      const value = options[property];
+      if (is.defined(value) && !(is.integer(value) && value >= 0)) {
+        throw is.invalidParameterError(property, 'a positive integer', value);
+      }
+    }
     return sharp.cache(options.memory, options.files, options.items);
   } else {
     return sharp.cache();
--- a/src/utilities.cc
+++ b/src/utilities.cc
@@ -23,15 +23,15 @@ Napi::Value cache(const Napi::CallbackInfo& info) {
 
   // Set memory limit
   if (info[size_t(0)].IsNumber()) {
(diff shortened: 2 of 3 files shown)


lovell:
Thanks for the PR. Whilst we're fixing this, I think we might also want to use `Uint32Value` instead of `Int32Value` for the cache values as they should all be positive integers. I guess we could also verify these values in the JavaScript layer.

metsw24-max:
Makes sense. Switched all three cache values to `Uint32Value`. I kept the `size_t` widening on memory because `Uint32Value` on its own still does the MB-to-bytes multiply in 32-bit and wraps from 4096MB up.

Also added a guard in `lib/utility.mjs` that rejects any cache value that isn't a non-negative integer, with a test for it. All squashed into the one commit.

lovell:
Thanks for the updates, it looks like we might have to skip the max memory test on 32-bit systems. The `buildPlatformArch()` function should provide a suitable value for this.

metsw24-max:
Good shout. The 4096MB byte count overflows a 32-bit `size_t` too, so I now skip that test when `buildPlatformArch()` reports a 32-bit arch (`arm`, `ia32`, `wasm32`). Runs as before on 64-bit and skips cleanly when I force `npm_config_arch=ia32`.

lovell:
Thank you. (The wasm32 test failure is unrelated, I'm fixing this now.)

PR #4564, opened 2026-07-01
bound composite left and top to coordinate limit

metsw24-max:
**Composite offsets truncate to 32-bit in native**

The per-overlay `left` and `top` are validated with `is.integer` only, so any safe integer up to 2^53 is accepted, but `AttrAsInt32` narrows them before they reach the `vips_composite` position arrays. An offset such as `left: 4294967301` silently becomes `5`, placing the overlay somewhere other than requested, while `2147483648` wraps negative and the overlay is dropped off-canvas with no error.

Bounded both to -100000000..100000000, the same coordinate limit the sibling `extract`/`create`/`raw` options already use. Negative offsets stay valid, so partly off-edge overlays are unaffected; only values that would truncate are now rejected up front. Regression cases added alongside the existing invalid `left`/`top` tests.


lovell:
Thank you

PR #4569, opened 2026-07-10
round palette bit depth up for png and gif colours

metsw24-max:
**Palette bit depth rounds down instead of up**

`bitdepthFromColourCount` maps the `colours` option to a PNG/GIF palette bit depth of 1, 2, 4 or 8, but it rounds the required bit count down to the nearest power of two, so a value needing 5 to 8 bits is capped at 4 and one needing 3 bits at 2. `png({ colours: 128 })` and `gif({ colours: 100 })` therefore emit a 16-entry palette, and `colours: 8` a 4-entry one, quietly discarding colours the caller asked to keep (visible via metadata `bitsPerSample`). Rounding up instead keeps the stored depth one of 1/2/4/8 while making it large enough to hold the requested count.

The reduced-colours GIF size test happened to depend on the old behaviour, as `colours: 128` only shrank the file because it collapsed to 16 entries, so it now uses `colours: 16` which still exercises a genuinely smaller palette. Added a PNG regression test asserting the written bit depth can represent the requested number of colours.

### DIFF
--- a/lib/output.mjs
+++ b/lib/output.mjs
@@ -33,7 +33,7 @@ const jp2Regex = /\.(jp[2x]|j2[kc])$/i;
 
 const errJp2Save = () => new Error('JP2 output requires libvips with support for OpenJPEG');
 
-const bitdepthFromColourCount = (colours) => 1 << 31 - Math.clz32(Math.ceil(Math.log2(colours)));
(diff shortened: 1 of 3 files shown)


lovell:
Thanks for fixing this

PR #4572, opened 2026-07-14
honour the tiff.subifd input option

metsw24-max:
**tiff.subifd is silently ignored**

The native input descriptor guards the sub-IFD read with `HasAttr(input, "subifd")`, but `lib/input.mjs` only ever sets the `tiffSubifd` property, so the guard never matches and the value keeps its `-1` default on the way to the libvips loader. Matched the guard to the read (which already uses `tiffSubifd`) and added a regression test.


lovell:
Thanks for fixing this too!

PR #4578, opened 2026-08-01
set n-pages in SetAnimationProperties to match pages written

metsw24-max:
**Stale n-pages on animated output**

SetAnimationProperties sets page-height on the output image but leaves n-pages at the value the loader recorded for the input file, so `info.pages` reports the source page count rather than the number actually written: `sharp(animated, { pages: 2 })` reports 30 alongside a height of 160 and a pageHeight of 80, and the same stale count is persisted to `v` output. Setting it beside page-height keeps the pair consistent, which is what the raw/create input paths and the animated join already do.

### DIFF
--- a/src/common.cc
+++ b/src/common.cc
@@ -711,9 +711,12 @@ namespace sharp {
     bool hasDelay = !delay.empty();
     VImage copy = image.copy();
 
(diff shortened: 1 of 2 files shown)


lovell:
Thanks for finding and fixing this.

PR #4580, opened 2026-08-03
Add AbortSignal support for image processing cancellation

### LINKED ISSUE #4579: Add support for `AbortSignal` to allow cancellation of in-flight image processing operations.

Add support for `AbortSignal` to allow cancellation of in-flight image processing operations.

Currently, once an image processing pipeline starts (e.g., `sharp(input).resize(...).toBuffer()`), there's no way to cancel it. This is problematic for:

- Long-running operations on large images or complex transformations
- HTTP servers that need to cancel processing when clients disconnect
- Interactive applications where users can cancel operations
- Resource-constrained environments where abandoned work should be stopped

The existing `timeout()` option provides time-based cancellation, but doesn't support event-driven or user-initiated cancellation patterns that are standard in modern Node.js APIs.

Searched for: "abort", "cancel", "AbortSignal", "AbortController"

Found that the `timeout()` feature exists but is limited to time-based cancellation. There's no support for the standard `AbortSignal` API that's now ubiquitous in Node.js (fetch, streams, fs.promises, etc.).

The timeout mechanism uses libvips progress callbacks to stop processing, which provides a foundation that could be extended to support abort signals.

```javascript
const controller = new AbortController();

// Option 1: Pass signal to constructor
const promise = sharp('input.jpg', { signal: controller.signal })
  .resize(4000, 4000)
  .toBuffer();

// Option 2: Pass signal to output methods
const promise = sharp('input.jpg')
  .resize(4000, 4000)
  .toBuffer({ signal: controller.signal });

// Option 3: Pass signal to toFile
const promise = sharp('input.jpg')
  .resize(4000, 4000)
  .toFile('output.jpg', { signal: controller.signal });

// Cancel the operation
controller.abort();

// Promise rejects with AbortError
// { name: 'AbortError', code: 'ABORT_ERR', message: 'The operation was aborted' }
```

The signal should be accepted in:
- `sharp()` constructor options
- `toBuffer()` options  
- `toFile()` options

When aborted, processing should actually stop (not just discard the result) and reject with a standard `AbortError`.

1. **External timeout wrappers**: Using `Promise.race()` with timeouts. This doesn't actually stop the processing, just ignores the result, wasting CPU and memory.

2. **Worker thread management**: Running sharp in worker threads and terminating them. This adds complexity, doesn't integrate with Node.js cancellation patterns, and may leave resources in inconsistent states.

3. **Custom event system**: Building a sharp-specific cancellation API. This would be non-standard and require developers to learn a new pattern.

The `AbortSignal` API is the standard across Node.js and web APIs, so supporting it aligns sharp with modern JavaScript patterns and provides a familiar interface for developers.

### Please provide sample image(s) that help explain this feature

Example use case - HTTP server with request cancellation:

```javascript
app.get('/image/:id', (req, res) => {
  const controller = new AbortController();
  
  // Cancel processing if client disconnects
  req.on('close', () => controller.abort());
  
  sharp(imagePath)
    .resize(800)
    .toBuffer({ signal: controller.signal })
    .then(buffer => res.send(buffer))
    .catch(err => {
      if (err.name === 'AbortError') {
        console.log('Processing cancelled');
        return;
      }
      res.status(500).send(err.message);
    });
});
```

This pattern is essential for building efficient HTTP services that don't waste resources on abandoned requests.

niwinz:
## Summary

This PR adds support for the standard `AbortSignal` API to allow cancellation of in-flight image processing operations. The signal can be provided via the `sharp()` constructor options or output method options (`toBuffer()`, `toFile()`).

Closes #4579 

## Motivation

Image processing operations can be long-running, especially for:
- Large images (multi-megapixel files)
- Complex transformations (multiple resize, rotate, composite operations)
- Slow encoders (AVIF, WebP with high quality settings)
- Batch processing scenarios

Currently, once processing starts, there's no way to cancel it. This leads to:

1. **Wasted resources**: CPU and memory are consumed even when the result is no longer needed
2. **Poor HTTP integration**: Servers can't stop processing when clients disconnect
3. **Bad UX**: Interactive applications can't respond to user cancellation requests
4. **Resource exhaustion**: In high-load scenarios, abandoned operations accumulate

The existing `timeout()` option provides time-based cancellation, but doesn't support event-driven or user-initiated cancellation patterns that are standard in modern Node.js APIs like `fetch()`, `fs.promises`, and streams.

### Cross-thread communication

The core challenge is that image processing happens on a libuv worker thread (via the C++ addon), while the abort signal is triggered on the main JavaScript thread. We need a way to communicate the cancellation across this boundary.

**Solution**: Use a `SharedArrayBuffer(1)` as a lock-free flag shared between threads.

```
Main Thread                    Worker Thread
    |                              |
    |-- abort() -----------------> |
    |   Atomics.store(flag, 1)     |
    |                              |-- progress callback
    |                              |-- check flag
    |                              |-- vips_image_set_kill()
    |                              |
```

**Why SharedArrayBuffer?**
- **True shared memory**: Unlike regular buffers, `SharedArrayBuffer` provides genuine cross-thread visibility without synchronization primitives
- **No new thread-safe function plumbing**: Avoids the complexity of `napi_threadsafe_function` for this simple signal
- **Zero overhead when unused**: No buffer allocated, no callback registered when no signal is provided
- **Actually stops CPU work**: Unlike approaches that just discard results, this stops the underlying libvips processing

### Integration with existing timeout mechanism

The abort mechanism reuses the same libvips progress callback infrastructure as the existing `timeout()` feature:

1. A progress callback is registered on the `VipsImage` when processing begins
2. The callback checks the shared abort flag on each progress event
3. When the flag is set (via `Atomics.store()` from the abort listener), it calls `vips_image_set_kill()` to stop processing
4. libvips propagates the cancellation through the operation pipeline and returns an error

**Why reuse the timeout mechanism?**
- **Minimal C++ surface area**: Leverages existing, tested infrastructure
- **Consistent behavior**: Works the same way as timeout, just triggered differently
- **Proper cleanup**: libvips handles resource cleanup when operations are killed
- **Battle-tested**: The timeout mechanism has been in sharp for years and is well-tested

### Error handling

When an operation is aborted, the error is converted to a standard `AbortError`:

```javascript
{
  name: 'AbortError',
  code: 'ABORT_ERR',
  message: 'The operation was aborted'
}
```

This matches the standard `DOMException` pattern used by `fetch()` and other async APIs, making it easy for developers to handle cancellation consistently across their application.

### Already-aborted signals

If a signal is already aborted when passed to sharp, the operation rejects immediately without invoking the native pipeline. This avoids unnecessary work and provides fast feedback.

## API Design

The signal can be provided in two places:

### 1. Constructor options

```javascript
sharp(input, { signal: controller.signal })
  .resize(800)
  .toBuffer();
```

Applies to all subsequent operations on this sharp instance. Convenient for simple cases where the entire pipeline should be cancellable.

### 2. Output method options

```javascript
sharp(input)
  .resize(800)
  .toBuffer({ signal: controller.signal });
```

Applies to that specific output operation. Useful when you have multiple outputs from the same input and want different cancellation logic for each.

**Why both?**
- **Flexibility**: Different use cases need different granularity
- **Backward compatibility**: Existing code continues to work unchanged
- **Consistency**: Matches patterns in other Node.js APIs

## Testing

Comprehensive tests cover:
- Abort during processing (slow blur operation)
- Already-aborted signal (immediate rejection)
- Normal completion without abort
- Invalid signal type validation
- Signal via constructor option
- Signal via `toFile()`
- Signal via stream output

All existing tests continue to pass with 100% code coverage maintained.

## Backward compatibility

This change is fully backward compatible:
- The `signal` option is optional
- When not provided, behavior is identical to before
- No changes to existing APIs or behavior
- No performance impact when signal is not used

### 1. HTTP request cancellation

```javascript
app.get('/image/:id', (req, res) => {
  const controller = new AbortController();
  req.on('close', () => controller.abort());
  
  sharp(imagePath)
    .resize(800)
    .toBuffer({ signal: controller.signal })
    .then(buffer => res.send(buffer))
    .catch(err => {
      if (err.name === 'AbortError') return;
      res.status(500).send(err.message);
    });
});
```

### 2. User-initiated cancellation

```javascript
const controller = new AbortController();

// Start processing
const promise = processImage(controller.signal);

// User clicks cancel
cancelButton.addEventListener('click', () => {
  controller.abort();
});

await promise.catch(err => {
  if (err.name === 'AbortError') {
    console.log('Processing cancelled by user');
  }
});
```

### 3. Batch processing with early termination

```javascript
const controller = new AbortController();

for (const file of files) {
  try {
    await sharp(file)
      .resize(800)
      .toBuffer({ signal: controller.signal });
  } catch (err) {
    if (err.name === 'AbortError') {
      console.log('Batch processing cancelled');
      break;
    }
    throw err;
  }
}
```

### 4. Timeout with custom logic

```javascript
const controller = new AbortController();

// Custom timeout logic
const timeoutId = setTimeout(() => {
  controller.abort();
}, 5000);

try {
  await sharp(input)
    .complexOperation()
    .toBuffer({ signal: controller.signal });
} finally {
  clearTimeout(timeoutId);
}
```

## Notes

This implementation provides a foundation for cancellation support that aligns with modern Node.js patterns while maintaining full backward compatibility.

### DIFF
--- a/lib/constructor.mjs
+++ b/lib/constructor.mjs
@@ -423,6 +423,14 @@ const Sharp = function (input, options) {
     // Function to notify of queue length changes
     queueListener
   };
+  if (is.defined(options) && is.defined(options.signal)) {
+    if (!(options.signal instanceof AbortSignal)) {
+      throw is.invalidParameterError('signal', 'AbortSignal', options.signal);
+    }
+    this.options.signal = options.signal;
+  } else {
+    this.options.signal = null;
+  }
   this.options.input = this._createInputDescriptor(input, options, { allowStream: true });
   return this;
 };
--- a/lib/index.d.ts
+++ b/lib/index.d.ts
@@ -650,35 +650,46 @@ declare namespace sharp {
          * @throws {Error} Invalid parameters
          * @returns A promise that fulfills with an object containing information on the resulting file
          */
-        toFile(fileOut: string): Promise<OutputInfo>;
+         toFile(fileOut: string): Promise<OutputInfo>;
 
-        /**
-         * Write output to a Buffer. JPEG, PNG, WebP, AVIF, TIFF, GIF and RAW output are supported.
+         /**
+          * Write output image data to a file.
+          * @param fileOut The path to write the image data to.
+          * @param options Options object with signal for cancellation.
+          * @throws {Error} Invalid parameters
+          * @returns A promise that fulfills with an object containing information on the resulting file
+          */
+         toFile(fileOut: string, options: { signal?: AbortSignal }): Promise<OutputInfo>;
+
+         /**
+          * Write output to a Buffer. JPEG, PNG, WebP, AVIF, TIFF, GIF and RAW output are supported.
          * By default, the format will match the input image, except SVG input which becomes PNG output.
          * @param callback Callback function called on completion with three arguments (err, buffer, info).
          * @returns A sharp instance that can be used to chain operations
          */
         toBuffer(callback: (err: Error, buffer: Buffer<ArrayBuffer>, info: OutputInfo) => void): Sharp;
 
-        /**
-         * Write output to a Buffer. JPEG, PNG, WebP, AVIF, TIFF, GIF and RAW output are supported.
-         * By default, the format will match the input image, except SVG input which becomes PNG output.
-         * The underlying `ArrayBuffer` may be marked as non-transferable by some JavaScript runtimes.
-         * @param options resolve options
-         * @param options.resolveWithObject Resolve the Promise with an Object containing data and info properties instead of resolving only with data.
-         * @returns A promise that resolves with the Buffer data.
-         */
-        toBuffer(options?: { resolveWithObject: false }): Promise<Buffer<ArrayBuffer>>;
+         /**
+          * Write output to a Buffer. JPEG, PNG, WebP, AVIF, TIFF, GIF and RAW output are supported.
+          * By default, the format will match the input image, except SVG input which becomes PNG output.
+          * The underlying `ArrayBuffer` may be marked as non-transferable by some JavaScript runtimes.
+          * @param options resolve options
+          * @param options.resolveWithObject Resolve the Promise with an Object containing data and info properties instead of resolving only with data.
+          * @param options.signal AbortSignal to cancel the operation.
+          * @returns A promise that resolves with the Buffer data.
+          */
+         toBuffer(options?: { resolveWithObject: false; signal?: AbortSignal }): Promise<Buffer<ArrayBuffer>>;
 
-        /**
-         * Write output to a Buffer. JPEG, PNG, WebP, AVIF, TIFF, GIF and RAW output are supported.
-         * By default, the format will match the input image, except SVG input which becomes PNG output.
-         * The underlying `ArrayBuffer` may be marked as non-transferable by some JavaScript runtimes.
-         * @param options resolve options
-         * @param options.resolveWithObject Resolve the Promise with an Object containing data and info properties instead of resolving only with data.
-         * @returns A promise that resolves with an object containing the Buffer data and an info object containing the output image format, size (bytes), width, height and channels
-         */
-        toBuffer(options: { resolveWithObject: true }): Promise<{ data: Buffer<ArrayBuffer>; info: OutputInfo }>;
+         /**
+          * Write output to a Buffer. JPEG, PNG, WebP, AVIF, TIFF, GIF and RAW output are supported.
+          * By default, the format will match the input image, except SVG input which becomes PNG output.
+          * The underlying `ArrayBuffer` may be marked as non-transferable by some JavaScript runtimes.
+          * @param options resolve options
+          * @param options.resolveWithObject Resolve the Promise with an Object containing data and info properties instead of resolving only with data.
+          * @param options.signal AbortSignal to cancel the operation.
+          * @returns A promise that resolves with an object containing the Buffer data and an info object containing the output image format, size (bytes), width, height and channels
+          */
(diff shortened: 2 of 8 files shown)


lovell:
Thanks for the PR, it looks like this is mostly LLM-generated.

At first glance:

- The new `signal` feature should replace the existing `timeout()` API, which will be deprecated and under the hood can use `AbortSignal.timeout()`.

 - The use of the `resolveWithObject` option on `toBuffer()` will soon be deprecated (and always set to `true`). If someone wants to set `signal` then they should not have control over `resolveWithObject`. This is to help protect consumers of this new API from upcoming breaking changes.

niwinz:
Yes, i'm using LLM for my work, but all of this is revewed by me in any case. I'm not very familiar with sharp codebase, so any feedback is more than welcome. We are starting to use [sharp on penpot](https://github.com/penpot/penpot/pull/10767) and cancellation was one thing i have missed on the API, this is the reason of this PR.

niwinz:
> Thanks for the PR, it looks like this is mostly LLM-generated.
> 
> At first glance:
> 
> * The new `signal` feature should replace the existing `timeout()` API, which will be deprecated and under the hood can use `AbortSignal.timeout()`.

I'm not in position on take this decision from the beginning, i mean, this is something that only you can know, if you provide me a guidance i will try to adapt the PR to the changes. Do you want me to do this also? (in my opinion this can be addressed in other PR)

> * The use of the `resolveWithObject` option on `toBuffer()` will soon be deprecated (and always set to `true`). If someone wants to set `signal` then they should not have control over `resolveWithObject`. This is to help protect consumers of this new API from upcoming breaking changes.

ACK, i will try to address this

niwinz:
I think i have addressed the second point of the feedback.

niwinz:
Any news on this?

lovell:
Hi, thanks again for the PR, and I am planning to revisit this soon.

Issue #3451, opened 2022-11-11
Possible race condition in metadata() when using Stream-based input


lucakcustom:
Considering the following code
```
const image_path = "./Images/128x128.png"
const test = async() => {
    const image1 = sharp(image_path).pipe(sharp())
    const image2 = sharp(image_path).pipe(sharp())
    const data1 = await image1.metadata()
    console.log(data1)
    const data2 = await image2.metadata()  // on this instruction the node interpreter will exit
    console.log(data2)
}
```
When we await for the metadata of data2 the node interpreter will immediately exit without any error. I tried to debug it but I could not find the point in which the node interpreter is closed (I might assume this happens in native, but I'm not sure).

This issue never happens if we work on different images so the following code is totally fine and works as expected

```
const image_path = "./Images/128x128.png"
const image_path_2 = "./Images/256x256.png"

const test = async() => {
    const image1 = sharp(image_path).pipe(sharp())
    const image2 = sharp(image_path_2).pipe(sharp())
    const data1 = await image1.metadata()
    console.log(data1)
    const data2 = await image2.metadata()
    console.log(data2)
}
```

lovell:
Thanks for reporting, there was a possible race condition when awaiting `metadata()` for Stream-based input, especially when the input Stream was very short i.e. smaller images.

Commit https://github.com/lovell/sharp/commit/df971207b88628c9f2cb4148732b85db2de7b1bd fixes this and adds a test that would previously have failed. This will be included as part of v0.31.3.

lucakcustom:
Thanks for your answer. Just out of curiosity, why was the interpreter exiting without returning any error? I would have expected some sort of error, even like a segfault, but I was getting nothing in all the environment I tried

lovell:
The "finish" event was already firing before the listener was attached, so the Promise returned by `metadata()` was never resolved or rejected.

I assume you invoked your sample by adding a call to `test()` at the top level rather than `await test()`. This means that once all other logic in the script had completed, the event loop would have been empty and with no callbacks remaining, the process ended.

lucakcustom:
Now I see. Thanks again, I appreciated your answer.

Issue #671, opened 2017-01-06
Streaming compatibility with archiver


danhaller:
Hi,

I'm trying to use _sharp_'s streaming capability in tandem with [_archiver_](https://github.com/archiverjs/node-archiver) to resize and zip multiple images and stream out to a consumer.  I can get each library to work on their own, but not together on any more than a single image.

Here's some example code:

```js
const fs = require('fs');
const archiver = require('archiver');
const sharp = require('sharp');
//const imagemagick = require('imagemagick-stream');
const request = require('request');

function resizeAndZip() {
  const imageStreams = [
    request('http://i.stack.imgur.com/tKsDb.png'),
    request('http://i.stack.imgur.com/EdUwb.png'),
    request('http://i.stack.imgur.com/5d55j.png')
  ];

  const zip = archiver('zip');

  zip.on('entry', (entry) => console.log(`${entry.name} successfully appended to zip`));

  imageStreams.map((stream) => {
    const resize = sharp().resize(100, 100);
    // const resize = imagemagick().resize('100x100');

    return stream.pipe(resize);
  })
  .forEach((resizedImageStream, i) => {
    zip.append(resizedImageStream, { name: `${i}.png` });
  });

  zip.finalize();

  return zip;
}

resizeAndZip().pipe(fs.createWriteStream(__dirname + '/output.zip'));
```

In the example you can see that I've also imported imagemagick-stream.  If this library is used as the resizer instead of sharp everything works fine.  I'm not sure whether this is a compatibility problem or if I'm using sharp incorrectly.

lovell:
Hello, I think the `resizedImageStream` instances have yet to enter flowing mode. This happens automagically when something else `pipe`s data out, but `archiver` doesn't appear to do this for you.

Something like the following should work:
```javascript
    zip.append(resizedImageStream.resume(), { name: `${i}.png` });
```

Perhaps sharp's [constructor jsdoc](https://github.com/lovell/sharp/blob/master/lib/constructor.js#L35) should include this information? I'd be happy to accept a PR if you'd like to make the docs clearer.

danhaller:
Thanks for the reply.  Adding ```resume``` is definitely an improvement, the zip file now completes with all the files.  It appears that most of the files are corrupt, however, when the archive is unzipped.  I'm trying not to get too stackoverflow here but is this likely to be a ```sharp``` issue or an ```archiver``` issue?

[...]

PR #4584, opened 2026-08-06
Prevent hang when output is requested after Stream input has finished

Jaybhade:
Every Stream-based input path waits for the Writable side to emit `finish` before flattening the accumulated chunks. `finish` is emitted only once, so any listener registered after it has already fired never runs — the Promise never settles and the callback is never called, with no error and no timeout.

```js
const pipeline = sharp().resize(320, 240);
createReadStream(input).pipe(pipeline);
await once(pipeline, 'finish');   // or await pipelineAsync(readable, pipeline)

await pipeline.toBuffer();        // hangs forever
```

Measured against `main`, waiting 2s for each call on an instance whose Writable side has already finished:

| call | before | after |
| --- | --- | --- |
| `toBuffer()` | hangs | resolves |
| `toBuffer(callback)` | hangs | resolves |
| `toFile(path)` | hangs | resolves |
| `metadata()` | resolves | resolves |
| `metadata(callback)` | hangs | resolves |
| `stats()` | hangs | resolves |
| `stats(callback)` | hangs | resolves |
| `clone()` then output | hangs | resolves |
| `.pipe(writable)` | resolves | resolves |

The two that already worked each carry their own workaround for this, added one entry point at a time: `metadata()` checks `writableFinished` (df97120, #3451) and Stream output re-emits `finish` when `streamInFinished` is set (d8df503, #671). Everything else was left waiting on an event that had gone.

## Fix

Both ad-hoc guards are replaced with a single `_whenStreamInFinished` helper, used everywhere the input stream is awaited:

```js
function _whenStreamInFinished (fn) {
  if (this.writableFinished) {
    fn();
  } else {
    this.once('finish', fn);
  }
}
```

Calling the handler directly means Stream output no longer has to re-emit `finish`, which also stops that event being delivered twice to any listener the caller has attached. With nothing left reading `streamInFinished`, its bookkeeping in `_write` goes too — and with it the `finish` listener that was attached to every Stream-fed instance.

An input stream that finishes without producing any data now rejects with the usual unsupported-format error rather than hanging, since `writableFinished` does not depend on a chunk having arrived.

No public API or type definitions change.

## Tests

Eight cases in `test/unit/io.js`, one per entry point above, each requesting output only after `finish` has been awaited. Seven of them hang on `main` and are only reported as failures by the 60s suite timeout; the eighth (`metadata()` via Promise) passes both before and after, and is included so the path that had the original workaround stays covered.

`npm test` is otherwise unchanged: 1821 tests, 100% branch coverage retained.

### DIFF
--- a/lib/constructor.mjs
+++ b/lib/constructor.mjs
@@ -492,7 +492,7 @@ function clone () {
   clone.options.queueListener = queueListener;
   // Pass 'finish' event to clone for Stream-based input
   if (this._isStreamInput()) {
-    this.on('finish', () => {
+    this._whenStreamInFinished(() => {
       // Clone inherits input data
       this._flattenBufferIn();
       clone.options.input.buffer = this.options.input.buffer;
--- a/lib/input.mjs
+++ b/lib/input.mjs
@@ -531,11 +531,6 @@ function _createInputDescriptor (input, inputOptions, containerOptions) {
 function _write (chunk, _encoding, callback) {
   if (Array.isArray(this.options.input.buffer)) {
     if (is.buffer(chunk)) {
-      if (this.options.input.buffer.length === 0) {
-        this.on('finish', () => {
-          this.streamInFinished = true;
-        });
-      }
       this.options.input.buffer.push(chunk);
       callback();
     } else {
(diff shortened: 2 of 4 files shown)


lovell:
This looks like a neat solution, thank you very much.

PR #4590, opened 2026-08-26
Docs: add security guidance for untrusted input

### LINKED ISSUE #4098: Docs: add section with suggestions relating to security (includes relaxing defaults for trusted input)
## Question about an existing feature
Running Sharp on user-generated content
A service which allows arbitrary user input of files, then converted to webps
### When you searched for similar issues, what did you find that might be related?
Tons of people seem to use Sharp for such a purpose, but I'm not sure if I'm missing a step. Running something like this on arbitrary data seems scary at best. I am running it in a kubernetes pod.

nanggo:
I use sharp to process untrusted remote images in an open-source social preview generator. This documents the relevant constructor defaults, download and runtime limits, loader restrictions, and when individual settings can be relaxed for trusted input.

Closes #4098.

### DIFF
--- a/docs/astro.config.mjs
+++ b/docs/astro.config.mjs
@@ -71,6 +71,7 @@ export default defineConfig({
           ]
         },
         { label: 'Performance', slug: 'performance' },
+        { label: 'Security', slug: 'security' },
         {
           label: 'Changelog',
           collapsed: true,
--- a/docs/src/content/docs/security.md
+++ b/docs/src/content/docs/security.md
@@ -1,0 +1,28 @@
+---
+title: Security
+---
+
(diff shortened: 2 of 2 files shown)


lovell:
Thank you very much for taking the time to open a PR. Inspired by this I've revisited and refreshed the existing security policy, expanding it to include some of the topics you've mentioned and discussed at #4098 - see commit https://github.com/lovell/sharp/commit/4180d8b1c11b8cf30b00fcfed7c4cbb1d99c2d81 

This also adds a dedicated security page to the website as well as displaying it on the GitHub repo.

nanggo:
Thanks for incorporating these ideas into the security policy and documentation!

PR #4593, opened 2026-08-29
Prevent TypeError when module load error has no code

### LINKED ISSUE #4592: err.code.endsWith() crashes with unhandled TypeError when wasm32 fallback fails (no .code on CompileError)

- [x] Running `npm install sharp` completes without error.
- [x] Running `node -e "import 'sharp'"` throws (this is the bug being reported).

- [x] I am using the latest version of `sharp` as reported by `npm view sharp dist-tags.latest` (0.35.3).

### What is the output of running `npx envinfo --binaries --system --npmPackages=sharp --npmGlobalPackages=sharp`?

```
System: Alpine Linux 3.23.3 (linux/amd64), inside Docker on a Proxmox VM
Node: v24.13.0
npmPackages: sharp: 0.35.3
```

On a CPU that doesn't support the x86-64-v2 microarchitecture (in my case a Proxmox VM with a generic `kvm64`-class CPU type, missing `ssse3`/`sse4_1`/`sse4_2`/`popcnt`), importing sharp crashes the whole process with an unrelated `TypeError` instead of a clear error message.

1. `require("@img/sharp-linuxmusl-x64/sharp.node")` loads the native binding fine.
2. `sharp._isUsingX64V2()` correctly returns `false` on this CPU.
3. sharp discards the native binding, pushes an error with `code: "Unsupported CPU"`, and falls back to the wasm32 build.
4. `require("@img/sharp-wasm32/sharp.node")` throws:
   ```
   CompileError: WebAssembly.Module(): Wasm SIMD unsupported @+2188
   ```
   This error has **no `.code` property** (`typeof err.code === "undefined"`).
5. That error is pushed into the same `errors` array as the others.
6. This code in `sharp.mjs` then runs:
   ```js
   errors.forEach((err) => {
     if (!err.code.endsWith("MODULE_NOT_FOUND")) {
   ```
   Since the wasm32 `CompileError` has `err.code === undefined`, `.endsWith()` throws:
   ```
   TypeError: Cannot read properties of undefined (reading 'endsWith')
       at file:///.../node_modules/sharp/dist/sharp.mjs:115:19
   ```
   This crashes the whole process instead of surfacing the real, actionable message (that the CPU lacks required instructions and no fallback is available).

I confirmed this isn't a detection bug: `_isUsingX64V2()` is correctly reporting an unsupported CPU here, once I fixed the underlying CPU config the whole thing worked as expected. So the ask isn't about the microarchitecture check itself, just the crash that happens afterward.

### Minimal reproduction of each step

```bash
node -e "require('@img/sharp-linuxmusl-x64/sharp.node')"
# -> loads fine

node -e "const s=require('@img/sharp-linuxmusl-x64/sharp.node'); console.log(s._isUsingX64V2())"
# -> false

node -e "try{require('@img/sharp-wasm32/sharp.node')}catch(e){console.log(e); console.log(typeof e.code)}"
# -> CompileError: WebAssembly.Module(): Wasm SIMD unsupported @+2188
# -> undefined
```

`sharp.mjs` line 115 assumes every error pushed into the `errors` array has a string `.code`, which isn't guaranteed, a `CompileError` from a failed wasm instantiation has none. A defensive check like `String(err.code || '').endsWith(...)` would let the real underlying message reach the user (e.g. "CPU doesn't support x86-64-v2 and wasm32 fallback isn't usable either") instead of an unrelated `TypeError` that just says "endsWith of undefined".

This will affect anyone running sharp 0.35.x on a CPU without x86-64-v2 support and no working wasm/SIMD fallback, which is a fairly common situation in VMs with a generic/capped CPU type (e.g. Proxmox's older default CPU model).

### Environment

- Alpine Linux 3.23.3
- Node.js v24.13.0
- sharp 0.35.3
- linux/amd64, inside Docker, no cross-arch/emulation
- CPU exposed to the VM was missing ssse3/sse4_1/sse4_2/popcnt (fixed on my end by changing the VM's CPU type in Proxmox, unrelated to this bug report but explains why it triggered)

lazerg:
When every attempt to load the native binding fails, the loop that builds the help message assumes each collected error carries a string `code`. A `CompileError` from a failed wasm32 instantiation has none, so `err.code.endsWith` throws a `TypeError` and the real reason for the failure never reaches the user.

The load path is import-time code inside a `node:coverage disable` block, so I couldn't see a way to cover this with a unit test without extracting the help-message builder into its own exported function. Happy to do that if you'd prefer.

Fixes #4592

### DIFF
--- a/lib/sharp.mjs
+++ b/lib/sharp.mjs
@@ -112,8 +112,8 @@ if (!sharp) {
 
   const help = [`Could not load the "sharp" module using the ${runtimePlatform} runtime`];
   errors.forEach((err) => {
-    if (!err.code.endsWith("MODULE_NOT_FOUND")) {
-      help.push(`${err.code}: ${err.message}`);
+    if (!err.code?.endsWith("MODULE_NOT_FOUND")) {
+      help.push(err.code ? `${err.code}: ${err.message}` : err.message);
     }
   });
   const messages = errors.map((err) => err.message).join(" ");


lovell:
Thanks for the PR

PR #4602, opened 2026-09-09
Ensure JxlOptions matches the options jxl() accepts

ramin-010:
I found this while comparing the JSDoc with the type definitions for the different format options. JxlOptions was missing AnimationOptions, even though JXL supports animation at runtime, so TypeScript was rejecting things like jxl({ loop: 3 }) and jxl({ delay: [100, 200] }). I added AnimationOptions to JxlOptions and added two type tests for these cases.

I also noticed the effort comment says 3–9, while the runtime check and JSDoc both allow 1–9, so I fixed that as well. Both changes are in the same interface/type mismatch, so I kept them together, but I can split the effort change into a separate PR if preferred.


lbesecker195 (approved):
Verified against `lib/output.mjs`: `jxl()` validates `effort` as an integer between 1 and 9 and calls `trySetAnimationOptions`, so the type definition was missing `AnimationOptions` and understated the effort range. Type test added. LGTM.

lovell:
@ramin-010 Thank you very much for fixing this.

